PR #63, opened 2023-07-22
chore: Add Solid template

BeanWei:


aklinker1 (approved):
Don't know much about solid, but this looks good to me 👍

PR #121, opened 2023-09-22
feat!: Require a function for `vite` configuration

aklinker1:
When migrating one of my large chrome extensions to WXT, I noticed lots of weird behavior when running it in dev mode. In particular, vue was failing to rebuild content script SFCs. This lead me to believe that Vite can't reuse the Vue plugin for both the dev server HTML pages and the library mode background/content scripts.

I released `0.6.0-alpha1`, which creates new plugins for every vite build, and that seemed to do the trick! There must be some internal state that breaks when the plugin is reused.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "wxt",
   "type": "module",
-  "version": "0.5.6",
+  "version": "0.6.0-alpha1",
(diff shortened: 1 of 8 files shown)


aklinker1:
Not updating templates until this is published. See #122

PR #154, opened 2023-10-05
feat: Allow customizing entrypoint options per browser

### LINKED ISSUE #152: Per-browser entrypoint options
### Feature Request

For example, if I want to set `persistent` to `false` for safari, but leave it as `true` for firefox, that's not possible right now.

So instead of just accepting a string, accept a Record of browser to a string.

```ts
export default defineBackground({
  persistent: {
    safari: false,
  },
});
```

This record should be available for all options, not just this one.

N/A

N/A

### Additional context

N/A

aklinker1:
See new E2E test for example usage. This closes #152 

### Todo

- [x] Implement
- [x] Add E2E test
- [x] Update docs

### DIFF
--- a/docs/guide/build-targets.md
+++ b/docs/guide/build-targets.md
@@ -44,3 +44,67 @@ wxt build --mv2
 ```
 
 When the `-b --browser` flag is not passed, it defaults to `chrome`. So here, we're targetting MV2 for Chrome.
+
+## Customizing Entrypoints
+
+There are several ways to customize entrypoint definitions per browser.
+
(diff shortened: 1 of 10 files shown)


docs/guide/build-targets.md
```diff
@@ -44,3 +44,67 @@ wxt build --mv2
...
+    edge: ['*://*.bing.com/*'],
+  },
+});
+```
+
+:::
+
+:::warning
+Only `defineBackground` and `defineContentScript` support per-browser options right now.
+:::
```
aklinker1, line 110: This PR doesn't implement HTML entrypoint support.

PR #167, opened 2023-10-14
feat!: Use `defineUnlistedScript` to define unlisted scripts

### LINKED ISSUE #150: Add `defineUnlistedScript` for unlisted script entrypoints
### Feature Request

The new `include` and `exclude` options for content scripts and the background let you include/exclude scripts based on the browser, but this doesn't exist for unlisted scripts. In addition, testing  files with no exports is difficult and annoying, providing a default export with a `main` function would simplify this process.

```ts
import { defineUnlistedScript } from 'wxt/client';

export function defineUnlistedScript({
  include?: string[];
  exclude?: string[];
  main(): void | Promise<void>;
});
```

N/A

Could use some kind of structured comment, or a named export, but using a `define*` default export matches the rest of the script styles.

### Additional context

#115

aklinker1:
This closes #150

BREAKING CHANGE: Unlisted scripts must now `export default defineUnlistedScript(...)`
BREAKING CHANGE: `BackgroundScriptDefintition` type renamed to `BackgroundDefinition`

### DIFF
--- a/src/core/utils/importEntrypointFile.ts
+++ b/src/core/utils/importEntrypointFile.ts
@@ -49,7 +49,6 @@ export async function importEntrypointFile<T>(
     cache: false,
     debug: config.debug,
     esmResolve: true,
-    interopDefault: true,
     alias: {
       'webextension-polyfill': resolve(
         config.root,
@@ -69,7 +68,8 @@ export async function importEntrypointFile<T>(
   });
 
   try {
-    return await jiti(path);
+    const res = await jiti(path);
+    return res.default;
   } catch (err) {
     config.logger.error(err);
     throw err;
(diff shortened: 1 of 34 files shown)


src/core/utils/importEntrypointFile.ts
```diff
@@ -49,7 +49,6 @@ export async function importEntrypointFile<T>(
     cache: false,
     debug: config.debug,
     esmResolve: true,
-    interopDefault: true,
```
aklinker1, line 52: Regression after switching to `esbuild` from `babel`: if there was no default export, this function returned an empty object instead of `undefined`.

Removing `interopDefault` fixes that.

PR #182, opened 2023-10-22
feat: Refactor project structure to export `initialize`, `prepare`, and `zip` functions

aklinker1:


aklinker1:
I don't understand why my coverage dropped. I did some more testing, like fully unit testing the `src/core/utils/package.ts` file, and I'm getting different amounts of coverage on the `package.ts` file when running the below commands:

```sh
pnpm test:coverage src/core/utils/__tests__/package.test.ts # 100%
pnpm test:coverage src/core/utils # 100%
pnpm test:coverage src/core # 100%
pnpm test:coverage src # 50%
```

So I think Vitest is not counting coverage correctly. I'm gonna try out instanbul instead of v8 and see if that fixes it.

PR #207, opened 2023-11-03
chore: Fix Svelte and React template READMEs

yyyanghj:
Update React template and Svelte template README


aklinker1 (approved):
Ah thank you, classic copy/paste error.

PR #209, opened 2023-11-04
fix!: Remove `lib` from `.wxt/tsconfig.json`

yyyanghj:
`lib` is mismatched cause we manual set `lib` option so that override the default value which effected by `target` option.

image

### DIFF
--- a/e2e/tests/typescript-project.test.ts
+++ b/e2e/tests/typescript-project.test.ts
@@ -252,7 +252,6 @@ describe('TypeScript Project', () => {
           \\"forceConsistentCasingInFileNames\\": true,
           \\"resolveJsonModule\\": true,
           \\"strict\\": true,
-          \\"lib\\": [\\"DOM\\", \\"WebWorker\\"],
           \\"skipLibCheck\\": true,
           \\"paths\\": {
             \\"@\\": [\\"..\\"],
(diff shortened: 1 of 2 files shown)


aklinker1:
Hmm, I think it makes more sense to remove the `lib` defaults all together.

I added it because by default, `WebWorker` types aren't added. They are *sometimes* useful for MV3 service workers, but rarely. If we exclude the `lib` from `.wxt/tsconfig.json`, like Nuxt does, it will include the `ESNext` libs because the target is `ESNext`.

If people want web worker types, they can add them to their own tsconfig. I can add some docs around this.

yyyanghj:
> Thanks for the PR!
> 
> It makes more sense to remove the `lib` defaults all together. Can you update your PR?
> 
> I added it because `WebWorker` types aren't present by default. They are _sometimes_ useful for MV3 service workers, but rarely. If we exclude the `lib` from `.wxt/tsconfig.json`, like Nuxt does, it will include the `ESNext` lib by default (because the target is `ESNext`).
> 
> If people want web worker types, they can add them to their own tsconfig. I can ad docs around TSConfig customization.

OK. I have updated PR.

aklinker1 (approved):
Thanks

aklinker1:
I'll fix the tests in a few hours, I'm traveling and my internet is bad. Just need to update some snapshots.

aklinker1:
@yyyanghj Released in [`v0.9.0`](https://github.com/wxt-dev/wxt/releases/tag/v0.9.0)

PR #211, opened 2023-11-04
chore: Remove whitespace from genearted `.wxt` files

aklinker1:
Pulled out of https://github.com/wxt-dev/wxt/pull/209#pullrequestreview-1713793601 to investigate why the snapshots are so different.


aklinker1:
Vitest's visualization is just bad in this case. The diff is actually as small as you would expect.

PR #223, opened 2023-11-08
feat: Download private packages when zipping sources

### LINKED ISSUE #185: Download private packages for Firefox code review
### Feature Request

If your extension has private npm packages that require auth to upload, you don't want to share you token when generating sources.

We should download tarballs for all private packages, and update the zipped package.json to resolve to the local versions instead of the ones on the remote.

Private packages are any dependencies or sub-dependencies that start with a scope with a custom token in your `.npmrc` file.

N/A

Alternatively, we need to allow the `.npmrc` file to be zipped.

### Additional context

N/A

aklinker1:
This closes #185. It only adds support for `npm`, `pnpm`, and `bun`. Yarn doesn't seem to expose the URLs of package tarballs, so it can't be accomplished with the same `list` command as the others. There are other approaches (like parsing the `yarn.lock` file), but I won't try them out now.

- [x] Find private packages
- [x] Download private packages
- [x] Add version resolutions to `package.json` inside ZIP
- [ ] E2E tests


src/core/utils/package-manager/pnpm.ts
```diff
@@ -0,0 +1,53 @@
...
+        project[0]?.dependencies
+          ? flattenNpmDependencyMap(project[0].dependencies)
+          : [],
+        project[0]?.devDependencies
+          ? flattenNpmDependencyMap(project[0].devDependencies)
+          : [],
+      ].flat();
+    },
+
+    addResolutions(packageJson, entries) {
```
aklinker1, line 41: Add unit tests for all `addResolutions`

aklinker1:
Looks like [`npm pack <module-name>`](https://docs.npmjs.com/cli/v10/commands/npm-pack) can be used to download a module, so that simplifies things quite a bit

aklinker1:
Closing in favor of #507

Issue #224, opened 2023-11-09
webextension-polyfill incompatible with Manifest v3


dvlden:
I believe that polyfill is injected by wxt at all times, even when manifest v3 is used for build?
Correct me if I am wrong, but polyfill should not exist for manifest v3.

Perhaps this is all that's needed:
[https://github.com/mozilla/webextension-polyfill/issues/329#issuecomment-1594562103](https://github.com/mozilla/webextension-polyfill/issues/329#issuecomment-1594562103)

Aside, I'm having issues using `browser.browserAction.setIcon()` and `browser.browserAction.setBadgeText()` because on manifest v3, that's no longer correct API. It's now `browser.action`. If I have to create some of these fallbacks, then I think `webextension-polyfill` is useless for everything, except types. But then, we have chrome types here as an alternative: [https://www.npmjs.com/package/@types/chrome](https://www.npmjs.com/package/@types/chrome)

So perhaps you can make interface to match `browser` to be exact as `chrome`.

I'd love the option to opt-out of webextension-polyfill, so that I can create my own tiny wrapper, unless you see my point here, or unless I'm wrong and there's something that has to be enabled/disabled to make this work as expected?

Aside, how do I change manifest v2 to v3 for Firefox, because I want to test what may or may not work?

dvlden:
To answer some of these questions for myself and for the others, to create a different manifest build, just pass `--mv3` to the build or dev commands.

The `--mv3` will not change `service_worker` to `scripts` for Firefox browser. So it can't work as expected.

Still not sure how to get around all of it.

aklinker1:
> Correct me if I am wrong, but polyfill should not exist for manifest v3.

For the most part, you're correct. The polyfill is basically a noop for Chrome MV3 extensions, just like any other polyfill that is injected into a environment where it's not needed.

However, as a framework that supports more than just Chrome MV3 (Firefox MV2 and legacy Chrome MV2), WXT will not be dropping the polyfill until:

1. Firefox releases MV3 support, which currently expiremental, and
2. Google officially kills MV2

Even after Firefox supports MV3, [they've pledged to keep some of the MV2 APIs ad-blockers use](https://blog.mozilla.org/addons/2022/05/18/manifest-v3-in-firefox-recap-next-steps/), among other changes. This means that Chrome and Firefox will never be the same, and as developers, we will always have to do some form of feature detection.

> `webextension-polyfill` incompatible with Manifest v3

[...]

PR #231, opened 2023-11-11
feat: Experimental option to exclude `webextension-polyfill`

aklinker1:
Add an experimental option to remove `webextension-polyfill` from all builds from dev mode.

```ts
export default defineConfig({
  experimental: {
    includeBrowserPolyfill: false,
  },
});
```

Nothing else changes. Continue importing `browser` from `wxt/browser`. HMR and dev mode continues to work with the extension in `demo/`.

| `includeBrowserPolyfill: true` (default) | `includeBrowserPolyfill: false` |
|:-:|:-:|
| 470 kB | 246 kB |
| Screenshot 2023-11-11 at 1 09 11 PM | Screenshot 2023-11-11 at 1 08 56 PM |

### DIFF
--- a/demo/wxt.config.ts
+++ b/demo/wxt.config.ts
@@ -14,7 +14,4 @@ export default defineConfig({
   alias: {
     public: 'src/public',
   },
-  experimental: {
-    webextensionPolyfill: false,
-  },
 });
--- a/demo/wxt.config.ts
+++ b/demo/wxt.config.ts
@@ -14,4 +14,7 @@ export default defineConfig({
   alias: {
     public: 'src/public',
   },
+  experimental: {
+    webextensionPolyfill: false,
+  },
 });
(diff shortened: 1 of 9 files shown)


demo/wxt.config.ts
```diff
@@ -14,4 +14,7 @@ export default defineConfig({
   alias: {
     public: 'src/public',
   },
+  experimental: {
+    webextensionPolyfill: false,
+  },
```
aklinker1, line 19: Revert, add E2E test instead.

src/core/vite-plugins/excludeBrowserPolyfill.ts
aklinker1, line 1: @dvlden Here's a working plugin for removing the polyfill. I guess you don't need the regex characters like I [suggested here](https://github.com/wxt-dev/wxt/issues/224#issuecomment-1804922080).

```diff
        resolve: {
          alias: {
-           '^webextension-polyfill$': 'chrome.js',
+           'webextension-polyfill': 'chrome.js',
          },
        },
```

PR #260, opened 2023-12-05
docs: Adds tl;dv to homepage

italodeverdade:
Our next web-ext release is using wxt. We recently migrated our codebase from webextension-toolbox to wxt, and it was a game changer!


aklinker1:
Glad to hear!

aklinker1 (approved):

PR #261, opened 2023-12-06
feat: Vite 5 support

aklinker1:


aklinker1:
Can't get types working while supporting vite 4 and vite 5, so I'm making this a breaking change and only supporting vite 5 for WXT 0.11+

aklinker1:
Alright, everything is working, and the templates all work except Solid's. I've disabled it's validation for now since it's known to not work against the latest `main` branch, and will merge.

PR #274, opened 2023-12-12
fix: Discover `.js`, `.jsx`, and `.tsx` unlisted scripts correctly

### LINKED ISSUE #271: defineUnlistedScript works with index.ts but not index.tsx
### Describe the bug

- Given an unlisted script directory, files named `index.ts` work as intended, but `index.tsx` does not. The guide says `entrypoints/<name>/index.[jt]sx?`

### To Reproduce

[test.zip](https://github.com/wxt-dev/wxt/files/13647280/test.zip)

1. Create a new project with `pnpx wxt@latest init test`
2. Select React, pnpm
3. Create a new directory under `entrypoints` called `test`
4. In `test` create an `index.tsx` file with the code below

```
export default defineUnlistedScript({
  main() {
    console.log('test');
  }
});
```

### Expected behavior

- `.output` contains a file named `test.js`, this doesn't happen
- It will work if you rename `index.tsx` to `index.ts`

aklinker1:
This closes #271


templates/solid/entrypoints/popup/main.tsx
aklinker1, line 1: Moved this file to fix #275

PR #284, opened 2023-12-13
feat!: Add support for "main world" content scripts

### LINKED ISSUE #282: `defineContentScript` doesn't work with `world: "MAIN"`
### Describe the bug

Main world doesn't support browser APIs, but `defineContentScript` imports the polyfill because of `ContentScriptContext`.

Because these content scripts need sandboxed and can't have a `ctx`, WXT should export a separate function, `defineMainWorldContentScript`, from `wxt/sandbox` module, which doesn't include the `webextension-polyfill`, and thus avoids the error.

Alternatively, we could try and define globals inside [`content-script-entrypoint.ts`](https://github.com/wxt-dev/wxt/blob/0873c24ad828504c106329c869d5fd2b2ce8263d/src/virtual/content-script-entrypoint.ts) and tree-shake the polyfill out. Not sure if that's possible though.

Or a third option, if the content script runs in the main world, alias `webextension-polyfill` to a noop module, similar to how we alias it for [`experimental.includeWebextensionPolyfill`](https://wxt.dev/api/wxt/interfaces/InlineConfig.html#includebrowserpolyfill).

### To Reproduce

[Archive.zip](https://github.com/wxt-dev/wxt/files/13663832/Archive.zip)

```sh
pnpm dev
```

Then go to https://google.com and open the console.

Screenshot 2023-12-13 at 11 36 00 AM

### Expected behavior

Main world content scripts shouldn't import the polyfill and the script shouldn't crash.

### Environment

```
  System:
    OS: macOS 13.4.1
    CPU: (8) arm64 Apple M1 Pro
    Memory: 3.02 GB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 18.16.1 - ~/.asdf/installs/nodejs/18.16.1/bin/node
    Yarn: 1.22.19 - ~/.asdf/installs/nodejs/18.16.1/bin/yarn
    npm: 9.5.1 - ~/.asdf/plugins/nodejs/shims/npm
    pnpm: 8.9.0 - ~/.asdf/installs/nodejs/18.16.1/bin/pnpm
  Browsers:
    Chrome: 120.0.6099.71
    Safari: 16.5.2
  npmPackages:
    wxt: ^0.11.0 => 0.11.2 
```

### Additional context

CC @AndrewWalsh

aklinker1:
This closes #282.

Now, when a content script is built, it will use different virtual entrypoints depending on the `world` option. If it's `undefined` or `"ISOLATED"`, it will continue working as-is. If it's `"MAIN"`, then a separate entrypoint is imported that doesn't load the `webextension-polyfill`, since main world content scripts run in a sandboxed environment without access to the extension APIs.

BREAKING CHANGE: In order to isolate the `webextension-polyfill` from being imported automatically from `wxt/client` when importing `defineContentScript`, `defineContentScript` and `defineBackgroundScript` have been moved from `wxt/client` to `wxt/sandbox`.


mefengl:
@aklinker1

It seems this line might not reflect the recent breaking changes. Should it be updated to `wxt/sandbox`?

https://github.com/wxt-dev/wxt/blob/5f241cf407549d969b546eeef2a12a98406d901d/src/core/builders/vite/plugins/noopBackground.ts#L19

aklinker1:
@mefengl good catch! I'll fix it today

aklinker1:
@mefengl Fixed in #298, released in [`v0.12.5`](https://github.com/wxt-dev/wxt/releases/tag/v0.12.5)

PR #333, opened 2024-01-09
feat!: Refactor content script UI functions and add helper for "integrated" UIs

### LINKED ISSUE #273: Merge content script UI functions
I want to add a helper function for integrated UIs, similar to `createContentScriptUi` and `createContentScriptIframe`, but I'm recognizing that adding a third function would be confusing and that the API reference is hard to follow because of how I've constructed the [options type](https://wxt.dev/api/wxt/client/type-aliases/ContentScriptUiOptions.html).

So I think I'm going to simplify all that into a single function with a basic type for the options. Something like this:

```ts
export function createContentScriptUi(
  options: {
    mode?: "integrated" | "shadow-root" | "iframe",
    position?: "inline" | "overlay" | "modal",
    zIndex?: number;
    anchor?: string | Element;
    alignment?: "top-left" | "top-right" | "bottom-right" | "bottom-left";
    page?: string;
    css?: string;
    name?: string;
    mount?(container: Element): TApp;
    onRemove?(app: TApp): void;
  },
): ContentScriptUi;
```

_Originally posted by @aklinker1 in https://github.com/wxt-dev/wxt/issues/267#issuecomment-1852492032_

aklinker1:
This closes #273 

Breaking changes:

- `type: "inline" | "overlay" | "modal"` has been changed to `position: "inline" | "overlay" | "modal"`
- `createContentScriptUi({ ... })` &rarr; `createShadowRootUi({ ... })`
- `createContentScriptIframe({ ... })` &rarr; `createIframeUi({ ... })`
- `onRemove` is now called ***before*** the UI is removed from the DOM, previously it was called after the UI was removed


aklinker1:
Main problem with this override approach, the type-hints and type error messages suck. 

https://github.com/wxt-dev/wxt/assets/10101283/b8de66f2-920e-4eba-b75e-73fa2fb0c97d

Gonna try something else...

aklinker1:
I tried merging the options into one object, and that worked well for input options, but I also had to combine the output types, and now the return type isn't inferred correctly...

I also tried adding the type as an additional parameter before the options in the overrides, but it still sucks at providing type hints. Instead, I'm gonna keep the fucntions as separate functions, but remove the "ContentScript" from them.

Issue #330, opened 2024-01-08
Support control the accessable of entry points


yunsii:
### Feature Request

It seems that all existed entry points must be build and released, I want to work with some of all entry points or make some entry points only accessable in specific env, how can I acheive this?

For instance, if an extention is too big, it is very helpful for controling the accessable of entry points in dev env.

#### Is your feature request related to a bug?

N/A

### What are the alternatives?

I have to rename the entry name.

aklinker1:
Interesting idea, have you thought about what you'd want this to look like?

There are `include` and `exclude` options for entrypoints, but that only works for including and excluding entrypoints for specific browsers.

Right now, I'd recommend just renaming any entrypoints you don't want to run, putting a `.` in front of the name (`overlay.content/index.ts` &rarr; `.overlay.content/index.ts`). That should exclude them from being built.

yunsii:
`include` and `exclude` type infer from `type TargetBrowser = string`, I think I do not known the specific acceptable value. And the rename approach is so inconvenience to cooperation obviously, Maybe add extra field `accessable` (boolean) like `include` and `exclude` to control the accessable?

aklinker1:
It wouldn't do what you're asking, I just provided it as an example: https://wxt.dev/guide/multiple-browsers.html#customizing-entrypoints

If renaming is inconvinent, wouldn't a boolean also be inconvinent in a similar way? Both would involve a change to a version controlled file, or multiple version controlled files.

Would a CLI flag like `wxt -e popup` or `wxt --allow-entrypoint popup` be better? That way the change wouldn't be tracked in version control. CLI flags are also configurable through the `wxt.config.ts` file, so there would be a way to configure it all in one place as well, instead it spread accross multiple files.

```ts
import { defineConfig } from 'wxt';

export default defineConfig({
  allowedEntrypoints: ["popup", "options", ...],
});
```

What do you think of that?

aklinker1:
Hmm, we could probably add types for it.

yunsii:
Is the feature in the recent plan? We have to wait not short of time to develop only one entrypoint for now, because there are 10+ existing entrypoints 😂

aklinker1:
I don't plan on working on this right now. My priority is everything required for v1.0 to be released.

It could be a good first issue though, if anyone is looking to contribute. To get started:

- You'll need to figure out how to add an array flag with [`cac`](https://www.npmjs.com/package/cac). Will likely go around here: 
[...]

PR #344, opened 2024-01-14
feat: Add `filterEntrypoints` option to speed up development

yunsii:
ref: https://github.com/wxt-dev/wxt/issues/330

Add a simple test case and passed, but I found that test cases of `src/core/utils/building/__tests__/find-entrypoints.test.ts` seems outdated, not only caused by my changes:

image

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -20,6 +20,9 @@ cli
   .option('-c, --config <file>', 'use specified config file')
   .option('-m, --mode <mode>', 'set env mode')
   .option('-b, --browser <browser>', 'specify a browser')
+  .option('-e, --allow-entrypoint <entrypoint>', 'specify allow entrypoints', {
+    type: [],
+  })
   .option('--mv3', 'target manifest v3')
   .option('--mv2', 'target manifest v2')
   .action(
@@ -31,6 +34,7 @@ cli
         manifestVersion: flags.mv3 ? 3 : flags.mv2 ? 2 : undefined,
         configFile: flags.config,
         debug: flags.debug,
+        allowEntrypoints: flags.allowEntrypoints,
       });
       await server.start();
       return { isOngoing: true };
@@ -43,6 +47,9 @@ cli
   .option('-c, --config <file>', 'use specified config file')
   .option('-m, --mode <mode>', 'set env mode')
   .option('-b, --browser <browser>', 'specify a browser')
+  .option('-e, --allow-entrypoint <entrypoint>', 'specify allow entrypoints', {
+    type: [],
+  })
   .option('--mv3', 'target manifest v3')
   .option('--mv2', 'target manifest v2')
   .option('--analyze', 'visualize extension bundle')
@@ -58,6 +65,7 @@ cli
         analysis: {
           enabled: flags.analyze,
         },
+        allowEntrypoints: flags.allowEntrypoints,
       });
     }),
   );
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -38,7 +38,7 @@ cli
         manifestVersion: flags.mv3 ? 3 : flags.mv2 ? 2 : undefined,
         configFile: flags.config,
         debug: flags.debug,
-        filterEntrypoints: flags.filterEntrypoints,
+        filterEntrypoints: flags.filterEntrypoint,
       });
       await server.start();
       return { isOngoing: true };
@@ -73,7 +73,7 @@ cli
         analysis: {
           enabled: flags.analyze,
         },
-        filterEntrypoints: flags.filterEntrypoints,
+        filterEntrypoints: flags.filterEntrypoint,
       });
     }),
   );
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -38,7 +38,7 @@ cli
         manifestVersion: flags.mv3 ? 3 : flags.mv2 ? 2 : undefined,
         configFile: flags.config,
         debug: flags.debug,
-        filterEntrypoints: flags.filterEntrypoint,
+        filterEntrypoints: getArrayFromFlags(flags, 'filterEntrypoint'),
       });
       await server.start();
       return { isOngoing: true };
@@ -73,7 +73,7 @@ cli
         analysis: {
           enabled: flags.analyze,
         },
-        filterEntrypoints: flags.filterEntrypoint,
+        filterEntrypoints: getArrayFromFlags(flags, 'filterEntrypoint'),
       });
     }),
   );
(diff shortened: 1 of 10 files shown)


src/types/external.ts
```diff
@@ -31,6 +31,12 @@ export interface InlineConfig {
    * @default "${config.srcDir}/entrypoints"
    */
   entrypointsDir?: string;
+  /**
+   * Specify allow entrypoints, default allow all entrypoints
+   *
+   * @default true
+   */
+  allowEntrypoints?: true | string[];
```
aklinker1, line 39: I'd prefer just `string[]` here. True is the same as undefined, I don't want more than one type to have the same meaning.

```suggestion
  allowEntrypoints?: string[];
```

src/types/internal.ts
```diff
@@ -21,6 +21,7 @@ export interface InternalConfig {
   wxtDir: string;
   typesDir: string;
   entrypointsDir: string;
+  allowEntrypoints: true | string[];
```
aklinker1, line 24:
```suggestion
  allowEntrypoints: string[] | undefined;
```

src/core/utils/building/find-entrypoints.ts
```diff
@@ -125,6 +144,9 @@ export async function findEntrypoints(
     if (include?.length && !exclude?.length) {
       return include.includes(config.browser);
     }
+    if (disallowedEntrypointNames.includes(entry.name)) {
+      return false;
+    }
```
aklinker1, line 149: Now that I think about it, putting this inside `findEntrypoints` is wrong. When the `.wxt` directory is updated, the `.wxt/types/paths.d.ts` will also be filtered. We should probably move the filter to the `rebuild` function to prevent that from happening...

But we also need to move the `include` and `exclude` checks to `rebuild` for the same reason.

Let's keep this as-is for the this PR, no changes required. Let's move the filtering to `rebuild` in a separate, future PR.

aklinker1 (commented):
Looks good! I think I prefer `filterEntrypoints` now that I see it implemented. So rename:

- `allowEntrypoints` &rarr; `filterEntrypoints`
- `-e, --allow-entrypoint <entrypoint>` &rarr; `-e, --filter-entrypoint <entrypoint>`
- `disallowed` &rarr; `skipped`

Once the PR is opened for review, I'll go in and update the wording of some of the comments.

aklinker1:
> Add a simple test case and passed, but I found that test cases of src/core/utils/building/__tests__/find-entrypoints.test.ts seems outdated, not only caused by my changes:

They are not out of date, must be something with your setup. CI ensures those tests are passing on both Windows and Linux. I use mac for development, and can confirm they work on Mac as well.

https://github.com/wxt-dev/wxt/actions/workflows/validate.yml

src/core/utils/building/find-entrypoints.ts
yunsii: > But we also need to move the `include` and `exclude` checks to `rebuild` for the same reason.

Yes, I also found this problem.

yunsii:
I use Ubuntu in WSL, and test cases failed like following:

```diff
 FAIL  |unit| src/core/utils/building/__tests__/find-entrypoints.test.ts > findEntrypoints > should find entrypoint for content.css
 FAIL  |unit| src/core/utils/building/__tests__/find-entrypoints.test.ts > findEntrypoints > should find entrypoint for content/index.css
AssertionError: expected { name: 'content', …(5) } to deeply equal { type: 'content-script-style', …(4) }

- Expected
+ Received

  Object {
    "inputPath": "/src/entrypoints/content.css",
    "name": "content",
-   "options": Object {},
+   "options": Object {
+     "exclude": undefined,
+     "include": undefined,
+   },
    "outputDir": "/home/yuns/github/wxt/.output/content-scripts",
+   "skipped": false,
    "type": "content-script-style",
  }

 ❯ src/core/utils/building/__tests__/find-entrypoints.test.ts:561:28
    559| 
    560|     expect(entrypoints).toHaveLength(1);
    561|     expect(entrypoints[0]).toEqual(expected);
       |                            ^
    562|   });
    563| 
```

I only add `skipped` field, the `options` field change is strange.

yunsii:
> I use Ubuntu in WSL, and test cases failed like following:
> 
> ```diff
>  FAIL  |unit| src/core/utils/building/__tests__/find-entrypoints.test.ts > findEntrypoints > should find entrypoint for content.css
>  FAIL  |unit| src/core/utils/building/__tests__/find-entrypoints.test.ts > findEntrypoints > should find entrypoint for content/index.css
> AssertionError: expected { name: 'content', …(5) } to deeply equal { type: 'content-script-style', …(4) }
> 
> - Expected
> + Received
> 
>   Object {
>     "inputPath": "/src/entrypoints/content.css",
>     "name": "content",
> -   "options": Object {},
> +   "options": Object {
> +     "exclude": undefined,
> +     "include": undefined,
> +   },
>     "outputDir": "/home/yuns/github/wxt/.output/content-scripts",
> +   "skipped": false,
>     "type": "content-script-style",
>   }
> 
>  ❯ src/core/utils/building/__tests__/find-entrypoints.test.ts:561:28
>     559| 
>     560|     expect(entrypoints).toHaveLength(1);
>     561|     expect(entrypoints[0]).toEqual(expected);
>        |                            ^
>     562|   });
>     563| 
> ```
> 
> I add `skipped` field only, the `options` field change is strange.

Got it, the root cause is that I introduced the `skipped` field, if I fixed it, the diff of `options` also disappeared 😂

yunsii:
Any progress? we really need this feature.

aklinker1 (approved):
CLI wasn't working, I fixed that. Also added colors to the list of filtered entrypoints:

Screenshot 2024-01-16 at 1 50 21 PM

yunsii:
> CLI wasn't working, I fixed that. Also added colors to the list of filtered entrypoints:
> 
> Screenshot 2024-01-16 at 1 50 21 PM
> 

Is it possible to add test cases for CLI?

aklinker1:
Probably... I've never done it before though. Something like this might work.

```ts
vi.mock("../build");
const buildMock = vi.mocked(build)

describe("CLI", () => {
  beforeEach(() => {
    vi.resetModules()
  });

  it("should ...", () => {
    process.argv = ["wxt", "build", ...];
    await import("../CLI");

    expect(buildMock).toBeCalledWith(...)
  });
})
```

Shouldn't need E2E tests just to check if the flags are parsed correctly, unit tests should be enough.

src/cli.ts
```diff
@@ -165,3 +181,12 @@ function wrapAction(
...
 }
+
+/**
+ * Array flags, when not passed, are either `undefined` or `[undefined]`. This function filters out
+ * the
+ */
+function getArrayFromFlags<T>(flags: any, name: string): T[] | undefined {
+  const array = [flags[name]].flat() as Array<T | undefined>;
+  return array.filter((item) => item != null) as T[];
+}
```
yunsii, line 192: @aklinker1 The result will be empty array at least, so it caused `userConfig` of `mergeInlineConfig(inlineConfig, userConfig)` has no effect.

After released v1.4.2, I have to patch wxt temporarily.

PR #350, opened 2024-01-16
fix: Allow `browser.runtime.getURL` to include hashes and query params for HTML paths

### LINKED ISSUE #349: browser.runtime.getURL can not pass options.html with hash part
### browser.runtime.getURL can not pass options.html with hash part

With options.html developed by Vue (SPA)，it simply use hash as route path param. However, **wxt** generate `.wxt/types/path.d.ts` specify `browser.runtime.getURL` param's type to some strings, among them are `options.html`. So, when there pass hash part, it will prompt ts error, for example
```ts
Argument of type '`/options.html#/search?keyword=${string}&city=${string}`' is not assignable to parameter of type 'PublicPath'.ts(2345)
```

### To Reproduce
with content script,
```
  main() {
      const url = browser.runtime.getURL(
        `/options.html#/search?keyword=1`
     )
     window.open(url)
  },
})
```
with options.html, use vue & vue-router

aklinker1:
This closes #349

### DIFF
--- a/demo/src/entrypoints/background.ts
+++ b/demo/src/entrypoints/background.ts
@@ -15,6 +15,10 @@ export default defineBackground(() => {
   browser.runtime.getURL('/');
   browser.runtime.getURL('/background.js');
   browser.runtime.getURL('/icon/128.png');
+  browser.runtime.getURL('/example.html#hash');
+  browser.runtime.getURL('/example.html?query=param');
+  // @ts-expect-error: should only allow hashes/query params on HTML files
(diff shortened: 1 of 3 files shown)


aklinker1:
Fix is merged, will try and release a new version by tomorrow. There's a few other PRs I'd like to merge before the next release. Will comment here when it's live.

aklinker1:
Released in https://github.com/wxt-dev/wxt/releases/tag/v0.14.2

PR #370, opened 2024-01-25
feat: Add `submit` command

### LINKED ISSUE #57: `wxt submit`
Use [`publish-browser-extension`](https://www.npmjs.com/package/publish-browser-extension) to create the `wxt submit` command.

aklinker1:
This closes #57!

Adds the `wxt submit` and `wxt submit init` commands. Not using the word "publish" since `npm publish` is a built-in command. Naming it submit will hopefully guide people to calling their package.json scripts "submit" as well.

### DIFF
--- a/docs/guide/publishing.md
+++ b/docs/guide/publishing.md
@@ -1,94 +1,66 @@
+---
+outline: deep
+---
+
 # Publishing
 
-WXT offers several utilities that simplify the publishing process.
+WXT will help you ZIP your extensions and submit them to the stores for review.
 
 ## First Time Publishing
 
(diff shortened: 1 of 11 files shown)


aklinker1:
Right now, I have it setup as an alias for `publish-extension`, and you have to pass the zips you want to upload.

I'm considering including building, zipping, and discovering all the zips in a single command. Like for example, the Firefox sources zip should never need to be passed for Firefox, WXT should know that it exists and is required.

aklinker1:
Nope, for now, it'll just be an alias.

PR #384, opened 2024-01-29
feat: Allow passing preferences to chrome, enable dev mode and conten…

aklinker1:
See the JS docs for example config:


pftom:
@aklinker1 why chromiumPref is just `string` types with `object` assign values?

Issue #386, opened 2024-01-29
Tailwind changes no longer detected


dvlden:
### Describe the bug

It's like it suddenly stopped working and it became a really painful DX. I am not really certain when did this happen, but I've been developing some new features on another branch, with so much time waste and pain. Every time I make some style changes, I must restart WXT in order to see them. Reloading the extension does not work either. Previously, this behaviour was only when I made changes in entry file `popup/index.html` for example, but now it is happening in all files.

### To Reproduce

Will update issue once I figure out what the problem is.

1. Pending...

### Expected behavior

Should take effect once file is saved.

### Video recording

[Video recording](https://github.com/wxt-dev/wxt/assets/3460368/8eb6b564-95a0-4110-8c7f-ba6100ca3a4a)

### Environment

```

  System:
    OS: macOS 14.2.1
    CPU: (12) arm64 Apple M2 Max
    Memory: 55.40 GB / 96.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 21.2.0 - ~/.asdf/installs/nodejs/21.2.0/bin/node
    npm: 10.2.3 - ~/.asdf/plugins/nodejs/shims/npm
    pnpm: 8.14.3 - /opt/homebrew/bin/pnpm
    bun: 1.0.21 - /opt/homebrew/bin/bun
  Browsers:
    Brave Browser: 121.1.62.153
    Chrome: 121.0.6167.85
    Edge: 121.0.2277.83
    Firefox Nightly: 124.0a1
    Safari: 17.2.1
    Safari Technology Preview: 17.4
  npmPackages:
    wxt: ^0.13.0 => 0.13.5

```

### Additional context

I tested the TW example, saw it's broken, but HMR works. Going to fix the template and then hunt down why it isn't working for me anymore.

aklinker1:
I see you're working on the reproduction, but I just wanted to say, I like the gradient border and logo you've got, it looks very crisp 😎

dvlden:
> I see you're working on the reproduction, but I just wanted to say, I like the gradient border and logo you've got, it looks very crisp 😎

Yeah, I am trying to reproduce on my own project. Go back a handful of commits to the point where it worked and then see how/where could I fu*k it up. 

Thanks, we could probably use something similar for WXT examples too. 💯

aklinker1:
[...]

PR #387, opened 2024-01-29
fix(dev): Resolve `script` and `link` aliases

dvlden:
Proof of concept. I am sure we can do better, but I just tried the simplest approach and it works fine. 

Resolves: #386

### DIFF
--- a/src/core/builders/vite/plugins/devHtmlPrerender.ts
+++ b/src/core/builders/vite/plugins/devHtmlPrerender.ts
@@ -173,7 +173,7 @@ export function pointToDevServer(
         // Outside the config.root directory, serve the absolute path
         element.setAttribute(
           attr,
-          `${server.origin}/@fs${resolvedAbsolutePath}`,
+          `${server.origin}/@fs${normalizePath(resolvedAbsolutePath)}`,
         );
       } else {
         // Inside the project, use relative path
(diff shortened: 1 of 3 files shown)


src/core/builders/vite/plugins/devHtmlPrerender.ts
```diff
@@ -61,6 +61,17 @@ export function devHtmlPrerender(
               const abs = resolve(dirname(id), src);
               const pathname = relative(config.root, abs);
               element.setAttribute(attr, `${server.origin}/${pathname}`);
+            } else {
+              const [alias, path] = Object.entries(config.alias).find(([key]) =>
+                src.startsWith(key),
+              )!;
+
+              const baseUrl = `${server.origin}/${relative(config.root, path)}`;
```
aklinker1, line 69: Looks really good, about what I was thinking. You'll need to call `normailzePath` on the relative path to make it work on Windows, otherwise the URL would look like `http://localhost:5173\assets\main.css`.

aklinker1:
Sounds good, thanks for opening a PR. I'll try and help out tonight if I can

src/core/builders/vite/plugins/devHtmlPrerender.ts
dvlden: Interesting. I guess I should add `normalize` from `node:path` on all the paths. No matter if they are absolute/relative or aliased?

dvlden:
What I dislike here is:

```ts
`${server.origin}/${pathname}${src.replace(`${alias}`, '')}`
//                          ^ missing `/` because it is included by alias.
```

aklinker1:
@dvlden I extracted the function and added tests. Let me know what you think of my changes.

dvlden:
@aklinker1 lovely, clean!

aklinker1:
And windows tests failed lol

Issue #357, opened 2024-01-18
Content Script ESM Support


aklinker1:
### Feature Request

As discussed in #335, it is possible to load ESM content scripts using a dynamic import. The downside is that since it's async, the standard `run_at` option has basically no effect.

I propose adding a new option to `defineContentScript`: `type: "module"`. Similar to the background's `type: "module"` option.

When WXT sees an async content script, it will load the script asynchronously using a dynamic import.

```ts
await import(
  /* @vite-ignore */
  browser.runtime.getURL("path/to/chunk/entrypoint.js")
);
```

Questions:
- [x] ~~_Does the `run_at` in the manifest make a difference in loading speed when using a dynamic import_~~
   > Jan 20, 2024: Slightly, we should respect it, and allow configuring it, with the knowledge that `document_start` doesn't work. See https://github.com/wxt-dev/wxt/issues/357#issuecomment-1901769303
- [x] ~~_Can async content scripts be bundled in one step alongside HTML entrypoints, or do they have to be separated into their own step? My concern here is mixing chunks with side-effects that only work in HTML pages_~~
   > Jan 20, 2024: I don't think this is an issue, it's only an issue for the background service worker where there's no `window` global
- [ ] How will dev mode function and reload content scripts. What about main world content scripts, should those be grouped with the regular HTML file builds, or be in their own build step?
   > Feb 2, 2024: Have not tested dev mode yet, but I am building the ESM background script with the other extension pages, so I'll do the same with isolated world content scripts. Main world content scripts will be combined with the sandbox pages since those don't have access to the browser APIs, like sandbox pages.
- [x] ~~_Should `type: "module"` be the default value? It works well with the default `run_at: "document_idle"`, and will likely provide a much better dev experience._~~
   > Jan 21, 2024: If dev mode works well, yes, it will be the default.
   >
   > Feb 2, 2024: Initially, I'm going to make this an opt-in just like the background's `type: "module"`. If it's working well for people, I'll try and make the the default before v1.0.

#### Is your feature request related to a bug?

#335

### What are the alternatives?

No real alternatives to the feature as a whole. Instead of adding a new field, we could use `runAt: "async"`, but that wouldn't provide a way to set the actual `run_at` in the manifest. That said, the `run_at` doesn't really matter, it can cause the browser to import the script earlier, but the code will never run before the `DOMContentLoaded` event.

### Additional context

CC @yunsii 

This will fix: https://github.com/wxt-dev/wxt/issues/270

[...]

Issue #53, opened 2023-07-20
Reload background service worker without reloading entire extension


aklinker1:
From Chrome's debugging article: https://developer.chrome.com/docs/extensions/mv3/tut_debugging/

> Also, if you have made changes to the service worker code, you can use the Update button and skipWaiting to apply the changes immediately.
>
> ![screesnhot](https://wd.imgix.net/image/BhuKGJaIeLNPW9ehns59NfwqKxF2/mJISZTRN34bmSbENQpVq.png?auto=format&w=845)

If there's a way to do this programatically, that would be great. It means extension pages wouldn't have to be closed when changing the background.

Maybe we have access to these ServiceWorker APIs inside the background?

- https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerRegistration/update
- https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerGlobalScope/skipWaiting

aklinker1:
By running the following code in the top-level scope of the background service worker, I was able to access the two functions necessary to update the service worker without reloading the entire extension:

```ts
addEventListener('install', (event) => {
  const sw = event.target as ServiceWorkerGlobalScope;
  const { registration, skipWaiting } = sw;
  const update = registration.update;
  console.log({ skipWaiting, update });
  globalThis.reloadBackground = async () => {
    await update();
    await skipWaiting();
  };
});
```

<img width="229" alt="Screenshot 2023-07-20 at 4 01 40 PM" src="https://github.com/aklinker1/wxt/assets/10101283/c480e79f-68fb-48e3-a036-699a2e3fc474">

Extension service workers can listen to the standard web worker [`install` event](https://developer.chrome.com/docs/extensions/mv3/service_workers/service-worker-lifecycle/), from which we're able to access the `ServiceWorkerGlobalScope` and `ServiceWorkerRegistration`.

Unfortunately, calling the `reloadBackground` function from the dev console results in an error:

<img width="640" alt="Screenshot 2023-07-20 at 4 15 11 PM" src="https://github.com/aklinker1/wxt/assets/10101283/f9385d3d-7180-4d67-98a2-248f89612893">

aklinker1:
Here's a minimal extension demoing this behavior: [minimal-example.zip](https://github.com/aklinker1/wxt/files/12114283/minimal-example.zip)

aklinker1:
[...]

PR #398, opened 2024-02-02
feat!: ESM background support

### LINKED ISSUE #392: Background service worker : `document` is not defined
### Describe the bug

In background context : If imported code from third party librariries are using `document`, the background script may fail to start or load with `Uncaught ReferenceError: document is not defined` error.

### To Reproduce

https://github.com/lionelhorn/wxt-wasm-vlcn-repro

Steps to reproduce the bug using the reproduction:

1. Install dependencies: `pnpm i`
2. Start dev mode: `pnpm dev`
3. Open `chrome://extensions/`
4. Reload `wxt-wasm-vlcn-repro` extension
5. Open errors

### Expected behavior
Background script / service worker to start without errors.

### Screenshots
image

### Environment

```
  System:
    OS: Windows 11 10.0.22631
    CPU: (32) x64 AMD Ryzen 9 7950X 16-Core Processor
    Memory: 33.11 GB / 63.15 GB
  Binaries:
    Node: 18.18.2 - C:\Program Files\nodejs\node.EXE
    npm: 10.2.1 - C:\Program Files\nodejs\npm.CMD
    pnpm: 8.14.0 - C:\Program Files\nodejs\pnpm.CMD
  Browsers:
    Edge: Chromium (121.0.2277.83)
```

aklinker1:
> [!NOTE]
> This PR is being merged into a feature branch, and will not be immediately available. I plan on adding #357 to the same feature branch before doing a release.

Adding `type: "module"` to the background script definition will cause it to be included in the ESM build along side any UIs. This speeds up development, reduces bundle size, and adds support things like WASM, which requires ES modules to load (this will address #392).

> Note that even if we're using ESM for the background, we need to restart the service worker by reloading the entire extension (see #53 for more details). Since HMR doesn't really do anything for the background, so I'm not adding support for that.

### Todo

- [x] Include in UI multi-page build when `type=module`
- [x] Double check that `type=module` is ignored for MV2
- [x] Fix dev mode reload messages
- [x] Add tests

### DIFF
--- a/src/core/builders/vite/index.ts
+++ b/src/core/builders/vite/index.ts
@@ -75,7 +75,7 @@ export async function createViteBuilder(
    * Return the basic config for building an entrypoint in [lib mode](https://vitejs.dev/guide/build.html#library-mode).
    */
   const getLibModeConfig = (entrypoint: Entrypoint): vite.InlineConfig => {
-    const entry = getEntrypointEntry(entrypoint);
+    const entry = getRollupEntry(entrypoint);
     const plugins: NonNullable<vite.UserConfig['plugins']> = [
       wxtPlugins.entrypointGroupGlobals(entrypoint),
     ];
@@ -141,7 +141,7 @@ export async function createViteBuilder(
       build: {
         rollupOptions: {
           input: entrypoints.reduce<Record<string, string>>((input, entry) => {
-            input[entry.name] = getEntrypointEntry(entry);
+            input[entry.name] = getRollupEntry(entry);
             return input;
           }, {}),
           output: {
@@ -254,7 +254,7 @@ function getBuildOutputChunks(
  * Returns the input module ID (virtual or real file) for an entrypoint. The returned string should
  * be passed as an input to rollup.
  */
-function getEntrypointEntry(entrypoint: Entrypoint): string {
+function getRollupEntry(entrypoint: Entrypoint): string {
   let virtualEntrypointType: VirtualEntrypointType | undefined;
   switch (entrypoint.type) {
     case 'background':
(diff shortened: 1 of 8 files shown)


src/core/builders/vite/index.ts
```diff
@@ -157,14 +141,14 @@ export async function createViteBuilder(
...
+            input[entry.name] = getEntrypointEntry(entry);
             return input;
           }, {}),
           output: {
             // Include a hash to prevent conflicts
             chunkFileNames: 'chunks/[name]-[hash].js',
-            // Include a hash to prevent conflicts
-            entryFileNames: 'chunks/[name]-[hash].js',
+            // Place JS entrypoints in main directory without a hash. (popup.html & popup.js are next to each other)
+            entryFileNames: '[name].js',
```
aklinker1, line 151: Breaking change to the build output format.

aklinker1: This change ensures the ESM background entrypoint (a JS file, not HTML), is output to the root of the build directory as well, just like the non-ESM background.

lionelhorn: @aklinker1 Is it possible to control the output path using wxt.config.ts?

To have one build for dev and prod.
So when I build for dev it doesn't overwrite previously built one for prod.
```
.output/
  dev-chrome-mv3/
    chunks/
      popup-<hash>.js
    popup.html
```
And also

```
.output/
  prod-chrome-mv3/
    chunks/
      popup-<hash>.js
    popup.html
```

aklinker1: @lionelhorn In a hacky way, yes. You could set an environment variable, and reference it in the `wxt.config.ts` file while setting the `outDir`. WXT doesn't provide any options for changing the target directories inside the output directory.

```sh
DEV=true wxt
wxt build
```

```ts
import { defineConfig } from 'wxt';

export default defineConfig({
  outDir: process.env.DEV === "true" ? ".output/dev" : ".output"
});
```

This would give you something like:

```
.output/
   chrome-mv3/
   dev/
     chrome-mv3/
     firefox-mv2/
   firefox-mv2/
```

PR #404, opened 2024-02-03
chore: Add dependabot for github actions

Timeraa:
Dependabot should also watch GitHub Actions for updates.

Ideally I would like to migrate it to [renovate](https://github.com/renovatebot/renovate) as I do find renovate a lot more extensible / useful as you can have one config for all your packages / projects in one repository and it has way more dependency checks it can do like actions, npm modules, your package.json's `packageManager` field and many more. Plus it can automerge if checks pass like tests (if enabled)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,3 +9,7 @@ updates:
     directory: '/' # Location of package manifests
     schedule:
       interval: 'monthly'
+  - package-ecosystem: 'github-actions'
+    directory: '/'
+    schedule:
+      interval: 'monthly'


aklinker1 (approved):
Thanks, I'll checkout renovate. I've never used it before, I've just stuck with dependabot because it's easy to setup.

I've had problems working with dependabot in monorepos, so we might have to switch to it soon. I'm planning on refactoring the repo to a monorepo and publishing the storage and localization APIs as seperate packages.

Timeraa:
> Thanks, I'll checkout renovate. I've never used it before, I've just stuck with dependabot because it's easy to setup.
> 
> I've had problems working with dependabot in monorepos, so we might have to switch to it soon. I'm planning on refactoring the repo to a monorepo and publishing the storage and localization APIs as seperate packages.

I've been stuck with Dependabot back then too but it always was a bit of a hassle to copy over the same config and it also sometimes didn't have options / strictness policies that I wanted so eventually I moved to renovate

PR #406, opened 2024-02-03
feat: ESM content script support

### LINKED ISSUE #357: Content Script ESM Support

aklinker1:
This PR likely won't be merged, this is a complex change and I just want to see what it will look like to get dev mode working.

Add `type: "module"` to your content script options:

```ts
  type: "module",
  // ...
});
```

This closes #357.

### Todo

- [x] Build mode works
- [ ] Try out dev mode... Not sure what will be required there yet


demo/src/entrypoints/background.ts
```diff
@@ -1,7 +1,7 @@
 import messages from 'public/_locales/en/messages.json';
 
 export default defineBackground({
-  // type: 'module',
+  type: 'module',
```
aklinker1, line 4: Revert

aklinker1:
Closing in favor of #584

Issue #420, opened 2024-02-05
Saving content script in dev mode for Manifest V2 throws a TypeError


hikiko4ern:
### Describe the bug

Saving the content script in dev mode for Manifest V2 (e.g. Firefox) throws

```
file://<path>/node_modules/wxt/dist/cli.js:2177
        return [key, value.sort()];
                           ^

TypeError: Cannot assign to read only property '0' of object '[object Array]'
    at Array.sort (<anonymous>)
    at file://<path>/node_modules/wxt/dist/cli.js:2177:28
    at Array.map (<anonymous>)
    at hashContentScriptOptions (file://<path>/node_modules/wxt/dist/cli.js:2175:34)
    at file://<path>/node_modules/wxt/dist/cli.js:2486:22
    at Array.reduce (<anonymous>)
    at addEntrypoints (file://<path>/node_modules/wxt/dist/cli.js:2485:51)
    at generateManifest (file://<path>/node_modules/wxt/dist/cli.js:2282:3)
    at async rebuild (file://<path>/node_modules/wxt/dist/cli.js:2665:65)
    at async file://<path>/node_modules/wxt/dist/cli.js:3144:37
```

at this line

https://github.com/wxt-dev/wxt/blob/bb022646e320114c685acea0eb2d8e74d0f84f6c/src/core/utils/content-scripts.ts#L38

This error is not thrown when using MV3.

### To Reproduce

[wxt-sort-readonly.zip](https://github.com/wxt-dev/wxt/files/14166489/wxt-sort-readonly.zip)

1. Install dependencies: `pnpm i`
2. Start dev mode for MV2: `pnpm dev:firefox` or `pnpm dev --mv2`
3. Save file `entrypoints/content.ts`

### Environment

```
  System:
    OS: Linux 5.15 Debian GNU/Linux trixie/sid
    CPU: (32) x64 AMD Ryzen 9 5950X 16-Core Processor
    Memory: 59.91 GB / 70.67 GB
    Container: Yes
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.11.0 - /mnt/wslg/runtime-dir/fnm_multishells/29666_1707135069144/bin/node
    npm: 10.2.4 - /mnt/wslg/runtime-dir/fnm_multishells/29666_1707135069144/bin/npm
    pnpm: 8.15.1 - /mnt/wslg/runtime-dir/fnm_multishells/29666_1707135069144/bin/pnpm
[...]

Issue #340, opened 2024-01-12
web_accessible_resources not MV2 compliant for Firefox


dboyd13:
### web_accessible_resources not outputted MV2 compliant

When specifying the `web_accessible_resources` within `wxt.config.ts` the MV2 output is not compliant with the schema [Expected by Firefox](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/web_accessible_resources) (tested on FireFox `115.5.0` and WXT `0.14.0`)

```
Extension is invalid

Reading manifest: Error processing web_accessible_resources.0: Expected string instead of {"resources":["script.js"],"matches":["https://example.com/*"]}
```

### To Reproduce

1. Specify `web_accessible_resources` within `wxt.config.ts`

```
manifest: {
    name: 'Extension name',
    description: "Extension description",
    version_name: "0.0.1",
    content_scripts: [
      {
        matches: ["*://*.github.com/*"],
        js: ['content-script.js'],
        run_at: "document_end"
      }
    ],
    web_accessible_resources: [
      {
        "resources": ["script.js"],
        "matches": ["https://example.com/*"]
      }
    ],
    permissions: ["storage", "tabs"],
  }
});
```

1. Build extension (`wxt build -b firefox`)

1. Review the produced `manifest.json` file within `.output/firefox-mv2`, which has `web_accessible_resources` not aligned with Firefox [MV2 expectations](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/web_accessible_resources)

1. Verify that it's not accepted by attempting to load into Firefox, and receive the following error:

```
Extension is invalid

Reading manifest: Error processing web_accessible_resources.0: Expected string instead of {"resources":["script.js"],"matches":["https://example.com/*"]}
```

### Expected behavior

Produces a MV2 manifest where `web_accessible_resources` value is an array containing string(s).

### Environment

```
  System:
    OS: macOS 13.6.3
    CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
    Memory: 438.41 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 18.17.1 - ~/.nvm/versions/node/v18.17.1/bin/node
    Yarn: 1.22.19 - ~/node_modules/.bin/yarn
    npm: 9.6.7 - ~/.nvm/versions/node/v18.17.1/bin/npm
[...]

PR #423, opened 2024-02-06
feat: Automatically convert MV3 `web_accessible_resources` to MV2

aklinker1:
Now, instead of having to use a function and an if statement to define `web_accessible_resources` when supporting both MV2 and MV3, you can just define the resources with the MV3 object style, and WXT will downgrade them automatically when targetting MV2.

The example I gave in #340:

```ts
const baseManifest = {
  // ...
}
const webAccessibleResources = [
  {
    "resources": ["script.js"],
    "matches": ["https://example.com/*"]
  }
];

export default defineConfig({
  manifest: ({ manifestVersion }) => ({
    ...baseManifest,
    web_accessible_resources: manfiestVersion === 3
      ? webAccessibleResources
      : webAccessibleResources.flatMap(entry => entry.resources),
  })
});
```

Can now be simplified to:

```ts
export default defineConfig({
  manifest: {
    // ...
    web_accessible_resources: [
      {
        "resources": ["script.js"],
        "matches": ["https://example.com/*"]
      }
    ],
  }
});
```

If you inproperly define MV2 `web_accessbile_resources` and target MV3, WXT will now throw an error:

Screenshot 2024-02-03 at 3 07 03 PM


aklinker1:
I worked on this before looking into #420 (nice) because the demo app wasn't setup to run as MV2. While working on it, I was able to create a test case that triggered the same error, so this PR fixes #420 as well as adding the new feature.

Issue #422, opened 2024-02-05
`wxt build --analyze` adds file size of stats*.json


lionelhorn:
### Describe the bug

Building with --analyze cli param wrongly adds the size of each stats*.json files
What are the stats*.json files for? Is it a leftover from the building of stats.html?

Also, would it possible to control where the stats.html is written and it's filename format?
Would be nice to get at least a timestamp in it.

### To Reproduce
Any project using 0.16.2

### Expected behavior

For the wxt build --analyze to report the same output size as `wxt build`

### Screenshots
```shell
>wxt build           

WXT 0.16.2                                                                                                                                                                                                                                                          21:36:19  
ℹ Building chrome-mv3 for production with Vite 5.0.12                                                                                                                                                                                                              21:36:19
✔ Built extension in 533 ms                                                                                                                                                                                                                                        21:36:20
  ├─ .output\chrome-mv3\manifest.json  449 B
  └─ .output\chrome-mv3\background.js  93.6 kB
Σ Total size: 94.05 kB
✔ Finished in 842 ms                
```

```shell
>wxt build --analyze
WXT 0.16.2                                                                                                                                                                                                                                                          21:35:51  
ℹ Building chrome-mv3 for production with Vite 5.0.12                                                                                                                                                                                                              21:35:52
✔ Built extension in 536 ms                                                                                                                                                                                                                                        21:35:52   
  ├─ .output\chrome-mv3\manifest.json  449 B
  ├─ .output\chrome-mv3\background.js  93.6 kB
  └─ .output\chrome-mv3\stats-0.json   18.86 kB
Σ Total size: 112.92 kB
ℹ Analysis complete:                                                                                                                                                                                                                                               21:35:52
  └─ stats.html
✔ Finished in 966 ms      
```

aklinker1:
> What are the stats*.json files for? Is it a leftover from the building of stats.html?

Yes, they get combined into the final stats.jtml file. I can probably remove then, I need to check the order of operations.

> Also, would it possible to control where the stats.html is written and it's filename format?
Would be nice to get at least a timestamp in it.

Yeah, I can add config for that!

aklinker1:
[`v0.16.4`](https://github.com/wxt-dev/wxt/releases/tag/v0.16.4) includes both the fix to remove the stats JSON files from the build summary and a option, `analysis.outputFile`, to customize where the output location. See #426 for an example of how to add a timestamp.

lionelhorn:
[...]

PR #431, opened 2024-02-06
feat: Add option to customize the analysis artifacts output

lionelhorn:
As discussed in https://github.com/wxt-dev/wxt/issues/422#issuecomment-1930095857

Added an option to be able to keep and move the stats-*.json artifacts.

Will use the outputFile for filename
```js
  analysis: {
    keepArtifacts: true,
    outputFile: `stats/bundle.html`
  }
```

Would result in 
```
stats/bundle-0.json
stats/bundle-1.json
```
etc..


src/core/utils/building/internal-build.ts
```diff
@@ -104,7 +104,28 @@ async function combineAnalysisStats(): Promise<void> {
...
+
+        // Rename stats*.json artifacts based on outputFile filename
+        const cleanedStatsFilename = path
+          .parse(statsFile)
+          .name.replace('stats-', '');
+        const artifactDestinationPath = `${outputDir}/${
+          path.parse(wxt.config?.analysis?.outputFile).name
+        }_${cleanedStatsFilename}.json`;
+
+        fs.move(statsFile, artifactDestinationPath, { overwrite: true });
```
aklinker1, line 123: Instead of moving files, can we change the filename we write them to in the first place?

https://github.com/wxt-dev/wxt/blob/2b374b98d281521707f1b79e87daf8b931dedd46/src/core/builders/vite/plugins/bundleAnalysis.ts#L13

aklinker1 (changes requested):

aklinker1:
I've added E2E tests with the expected output. I changed the expected logic - `stats-*.json` files should always be output next to the HTML file. Should simplify things.

aklinker1:
@lionelhorn Let me know if you're OK with these changes.

aklinker1 (approved):

lionelhorn:
@aklinker1 LGTM. Better indeed :)

PR #436, opened 2024-02-10
fix: Use `insertBefore` on mounting content script UI

ookkoouu:
`insertBefore` recreates the anchor element.
When using dynamic injection triggered by adding the anchor element, the browser will run into infinite loop.
and mini refactor included.

### DIFF
--- a/src/client/content-scripts/ui/__tests__/index.test.ts
+++ b/src/client/content-scripts/ui/__tests__/index.test.ts
@@ -367,7 +367,7 @@ describe('Content Script UIs', () => {
       });
     });
 
-    describe.todo('after', () => {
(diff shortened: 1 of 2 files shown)


aklinker1:
Thanks for the PR! I don't know how I didn't know about the `prepend` method 🤦

aklinker1 (approved):
Thanks for the PR! I don't know how I didn't know about the `prepend` method 🤦 

I knew the replaceWith usage was wrong, but it was simple and it worked... but yeah, recreating the element is bad.

aklinker1:
Released in https://github.com/wxt-dev/wxt/releases/tag/v0.16.6

PR #454, opened 2024-02-15
fix: Watch files outside project root during development

### LINKED ISSUE #428: Dev server doesn't watch changes in other packages
### Describe the bug
Dev server doesn't watch changes in code from other packages (but still internal to the repo)

### Expected behavior
Changes to be picked up as if they were in the same package.
React component changes from the same package are detected. 
Changes outside the extension package are not.

UI added as a content script

### Screenshots
https://github.com/wxt-dev/wxt/assets/50528276/8fecbe29-b462-4cd8-94c6-d01eefa8675e

aklinker1:
This closes #428.

https://github.com/wxt-dev/wxt/assets/10101283/ea1f1381-edbd-4003-b57a-a279c0a51038

### DIFF
--- a/src/core/create-server.ts
+++ b/src/core/create-server.ts
@@ -51,24 +51,13 @@ export async function createServer(
     // Build after starting the dev server so it can be used to transform HTML files
     server.currentOutput = await internalBuild();
 
-    // Add file watchers for files not loaded by the dev server
-    const additionalFiles = server.currentOutput.steps
-      .flatMap((step, i) => {
-        if (Array.isArray(step.entrypoints) && i === 0) {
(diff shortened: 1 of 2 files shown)


src/core/create-server.ts
```diff
@@ -278,3 +286,23 @@ function getFilenameList(names: string[]): string {
...
+        // Dev server is already watching all HTML/esm files
+        return [];
+      }
+
+      return step.chunks.flatMap((chunk) => {
+        if (chunk.type === 'asset') return [];
+        return chunk.moduleIds;
+      });
+    })
+    .filter(
```
aklinker1, line 303: Also filter out files already in `wxt.config.root`

PR #458, opened 2024-02-17
fix: Fix order of calling `onMount`

ookkoouu:
`createShadowRootUi` was different from the others


ookkoouu:
i found another issue, i'll close it for now

PR #459, opened 2024-02-17
fix: Fix order of ShadowRootUI hooks calling

ookkoouu:
According to #333, it seems `onMount` should call after the UI is mounted, and `onRemove` should call before the UI is removed, but ShadowRootUI was not.

It's inconvenient that `shadowHost` is not connected when mounting UI.

### DIFF
--- a/src/client/content-scripts/ui/index.ts
+++ b/src/client/content-scripts/ui/index.ts
@@ -119,19 +119,18 @@ export async function createShadowRootUi<TMounted>(
   let mounted: TMounted;
 
   const mount = () => {
-    // Mount UI inside shadow root
-    mounted = options.onMount(uiContainer, shadow, shadowHost);
-
     // Add shadow root element to DOM
     mountUi(shadowHost, options);
     applyPosition(shadowHost, shadow.querySelector('html'), options);
+    // Mount UI inside shadow root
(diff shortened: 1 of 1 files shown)


aklinker1:
> It's inconvenient that `shadowHost` is not connected when mounting UI.

What's your use case? This was intentional btw. For the shadow root, having the order flipped for `onMount` and `onRemoved` (like it is today) makes more sense to me. Don't you want your UI mounted and ready to go before showing it on the UI? Same with `onRemove`, don't you want to remove the shadow root from the UI before unmounting and destroying your UI?

aklinker1:
> onRemove is now called before the UI is removed from the DOM, previously it was called after the UI was removed

Oh I see I said that... Hmm

ookkoouu:
I wanted to mount the UI on multiple elements like a list. To know from `onMount` where each UI will be inserted, the `shadowHost` needs to be connected to the DOM already. (ex: mount the UI on a Youtube live chat icon, want to get its username)
Rendering artifacts is not so important IMO.

aklinker1 (approved):
Ok, this should be fine.

aklinker1:
Released in [`v0.16.9`](https://github.com/wxt-dev/wxt/releases/tag/v0.16.9)

Issue #469, opened 2024-02-20
Injecting Content Script conditionally


mbersier:
So I have this project where I need to inject my content script from my background.js only if a given cookie is set.
What I did is converting my content script js to a UnlistedScript, but UnlistedScript doesn't pass a ctx to it's main method. That means I cannot use the createShadowRootUi function and init my app in it.

Is there a way to still use content script, but inject it manually?

aklinker1:
I'll add a `registration:"runtime"` option to disable adding it to the manifest (and adding `host_permissions` instead).

aklinker1:
Released in [`v0.16.10`](https://github.com/wxt-dev/wxt/releases/tag/v0.16.10)

mbersier:
Thx @aklinker1 for this!
Since this new update, I noticed an issue with the generated background.js when registration is set to runtime.
It seems like reloadContentScript always fetches the content script for the hot reload, so it's not working well in dev, but no issues when building the app.

Sorry if I'm not creating a new issue, not sure if I should or not :/

aklinker1:
> reloadContentScript always fetches the content script for the hot reload

@mbersier I don't understand whats wrong this... For me, `reloadContentScript` is getting called correctly. When a script changes, it is reloaded properly.

aklinker1:
@mbersier Could you provide a simple reproduction so I don't have to create one myself when testing this out?

aklinker1:
@mbersier or could you just try `wxt@0.16.12-alpha1`? I just published test version with the dev mode changes

mbersier:
@aklinker1 Good news! wxt@0.16.12-alpha1 seems to be working fine in dev now!
Thanks!

aklinker1:
Awesome, I need to cleanup the code that finds the correct content script to reload, I'll release an update after that. Hopefully today or tomorrow

aklinker1:
Released in [`v0.17.0`](https://github.com/wxt-dev/wxt/releases/tag/v0.17.0)

PR #474, opened 2024-02-21
fix: Improve dev mode for content scripts registered at runtime

aklinker1:
Additional dev mode improvements for #469

### DIFF
--- a/src/virtual/utils/reload-content-scripts.ts
+++ b/src/virtual/utils/reload-content-scripts.ts
@@ -60,15 +60,13 @@ export async function reloadRuntimeContentScriptMv3(
   const registered = await browser.scripting.getRegisteredContentScripts();
   logger.debug('Existing scripts:', registered);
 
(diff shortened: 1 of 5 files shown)


src/virtual/utils/reload-content-scripts.ts
```diff
@@ -33,6 +50,37 @@ export async function reloadContentScriptMv3(contentScript: ContentScript) {
...
+) {
+  logger.log('Reloading content script:', contentScript);
+  const registered = await browser.scripting.getRegisteredContentScripts();
+  logger.debug('Existing scripts:', registered);
+
+  const existing = registered.find((cs) => {
+    const hasEveryJs = contentScript.js?.every((js) => cs.js?.includes(js));
+    const hasEveryCss = contentScript.css?.every(
+      (css) => cs.css?.includes(css),
+    );
```
aklinker1, line 67: This is a really bad way of checking this... extract the function and add unit tests

PR #477, opened 2024-02-22
feat(storage): Improved support for default values on storage items

aklinker1:
- Exposed `defaultValue` property on defined storage items
- Improved typing for `watch` function when a default value is present


aklinker1:
Reworded commit to add breaking change, see 9c80ad832bc1caffb0f22b96093ca93171ef0991

PR #478, opened 2024-02-22
fix(storage): Only run migrations when the extension is updated

aklinker1:
Once this is merged, see https://wxt.dev/guide/storage.html#running-migrations for ensuring migrations are ran.


aklinker1:
Reworded commit to add breaking change, see 164978582626d380b7b84f0d801053e08b160164

Issue #379, opened 2024-01-28
permissions and host_permissions auto-collecting


honwhy:
### permissions and host_permissions auto-collecting

In defineContentScript or defineBackgound, excel from matches which is auto collecting into manifest.content_scripts[i].matches, how about define another keys permission, host_permissions, and auto collect them and deduplicate so as to generate manifest.permissions or host_permissions in mv3 style.
#### Is your feature request related to a bug?

"N/A"

### What are the alternatives?

alternative solutions, is defined permissions, host_persmissions and web_accessible_resources according to mv2 or mv3 request.

### Additional context

ex,
```
export default defineContentScript({
  matches: ['https://*.blog.csdn.net/**'],
  permission: [],
  host_permissions: [],
  main(ctx) {
   ...
  },
})
```

aklinker1:
Closing in favor of #544, which includes these fields and others.

PR #489, opened 2024-02-27
feat(manifest): Automatically add `sidePanel` permission

eastack:
Automatically add sidePanel permission when a sidepanel entrypoint exists

### DIFF
--- a/src/core/utils/manifest.ts
+++ b/src/core/utils/manifest.ts
@@ -338,6 +338,7 @@ function addEntrypoints(
         open_at_install: defaultSidepanel.options.openAtInstall,
       };
     } else if (wxt.config.manifestVersion === 3) {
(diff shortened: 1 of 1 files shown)


aklinker1 (commented):
~~You have to define that permission separately? I thought side panels were like the popup, just including an `action` field in the manifest gives you access to the `browser.action` APIs. Is it not the same for side panels?~~

Edit: Huh, interesting, you do have to define it separately... WXT doesn't automatically add any other API permissions (ignoring dev mode), so I'm a little hesitant to start doing this right now. Let me sleep on this to decide if it's something I want to support right now.

aklinker1:
I'm gonna say no on this one for now. If you have a side panel and need to use the API, manually add it as a permission for now. This is kinda related to #379, so I'll throw it in that issue and implement it then.

Issue #336, opened 2024-01-09
Allow using imported variables in entrypoint options


yunsii:
### Describe the bug

In a content script entrypoint, if you import a variable, lets call it `MATCHES`, and try to use that variable to for the `matches` option, the build fails with the following error message:

```ts
import { MATCHES } from "@/data/matches";
import { sayHello } from "@/helpers/dev";

export default defineContentScript({
  matches: MATCHES.allUrl,
  async main() {
    sayHello("_document-idle.content");
  },
});
```

```
ERROR: MATCHES is not defined
```

![image](https://github.com/wxt-dev/wxt/assets/18096089/4467d693-d694-437e-99ae-015280106d74)

But `import { sayHello } from "@/helpers/dev"` is okey.

### To Reproduce

https://github.com/yunsii/wxt-reproduction

Steps to reproduce the bug using the reproduction:

1. `pnpm i --ignore-scripts`
2. `pnpm build --debug`

Here's the output:

```
> wxt-starter@0.0.0 build /Users/aklinker1/Development/local/wxt-entrypoint-imports-repro
> wxt build "--debug"


WXT 0.13.5                                                                                                          11:12:19 AM
ℹ Building chrome-mv3 for production with Vite 5.0.11                                                              11:12:20 AM
[11:12:20 AM] ⚙ Loading file metadata: /Users/aklinker1/Development/local/wxt-entrypoint-imports-repro/src/entrypoints/_document-idle.content/index.ts
[11:12:20 AM] ⚙ Loading file metadata: /Users/aklinker1/Development/local/wxt-entrypoint-imports-repro/src/entrypoints/background.ts
⚙ Text:                                                                                                            11:12:20 AM
export default defineBackground(() => {
  console.log('Hello background!', { id: browser.runtime.id });
});

No imports:
import { defineUnlistedScript, defineContentScript, defineBackground } from 'wxt/sandbox';

export default defineBackground(() => {
  console.log('Hello background!', { id: browser.runtime.id });
});

Code:
import { browser } from 'wxt/browser';
import { defineUnlistedScript, defineContentScript, defineBackground } from 'wxt/sandbox';

export default defineBackground(() => {
  console.log('Hello background!', { id: browser.runtime.id });
});

[jiti] [transpile] /Users/aklinker1/Development/local/wxt-entrypoint-imports-repro/src/entrypoints/background.ts (40.18ms)
[...]

PR #492, opened 2024-02-28
docs: Added DocVersionRedirector to "Using WXT" section

leos:
Adding [DocVersionRedirector](https://github.com/leos/DocVersionRedirector/). If you have a few moments, would love a quick look at the code and any feedback on how I'm using WXT.

### DIFF
--- a/docs/.vitepress/components/UsingWxtSection.vue
+++ b/docs/.vitepress/components/UsingWxtSection.vue
@@ -19,6 +19,7 @@ const chromeExtensionIds = [
   'youtube中文配音/oglffgiaiekgeicdgkdlnlkhliajdlja', // Youtube中文配音
   'agjnjboanicjcpenljmaaigopkgdnihi', // PreMiD
   'aiakblgmlabokilgljkglggnpflljdgp', // Markdown Sticky Notes
+  'nomnkbngkijpffepcgbbofhcnafpkiep', // DocVersionRedirector
 ];
 
 const { data, err, isLoading } = useListExtensionDetails(chromeExtensionIds);


aklinker1:
Cool project! I checked it out, everything looks good except for the background entrypoint, the `browser.runtime.onInstalled` should go inside the main function of `defineBackground`. 

https://github.com/leos/DocVersionRedirector/blob/147481ab1b206cae3c36aad06dd04ec96e19b3fb/src/entrypoints/background.ts#L3-L6

I would actually expect your build to crash, including a side-effect like that in an entrypoint - if you keep adding code outside the main function, you'll eventually run into #336. For example, notice how all the code is inside the main function in these background scripts:

- https://github.com/aklinker1/github-better-line-counts/blob/2454214c2026628639f2d7ac9063ba42f4b8a18e/src/entrypoints/background.ts
- https://github.com/wxt-dev/wxt/blob/50e8c86ce993347425b953bfd472f36e1f15e90b/demo/src/entrypoints/background.ts
- https://github.com/anime-skip/player/blob/f07125a19341fac635967a793d16d1611e7bc44a/apps/extension/src/entrypoints/background.ts

Otherwise it's looking good! I don't use react much, hows the developer experience developing a react-based extension with WXT? HMR working correctly?

aklinker1:
> Connected to this discussion, is there a discord/slack/gitter for wxt? It may make it a bit easier to communicate about the extension (although, you may not want to do it in chat :-) ).

I've hesitated creating a discord because I don't like being overrun with notifications lol. But I need to do it now that WXT has almost hit 2k stars.

> Do I need to also do the same for unlisted scripts?

Neither of those files are unlisted scripts. Unlisted scripts are JS/TS files in the entrypoints directory that aren't content scripts or background scripts.

https://wxt.dev/entrypoints/unlisted-scripts.html

Unlisted scripts are scripts you inject dynamically, usually by adding a `script` block to the page from a content script. Those scripts run in the page's context, and have access to more information about the website. In your case, helper files or utils aren't considered unlisted scripts. So you don't need `defineUnlistedScript` for those.

leos:
> I've hesitated creating a discord because I don't like being overrun with notifications lol. But I need to do it now that WXT has almost hit 2k stars.

No worries - do what works best for you!

> Neither of those files are unlisted scripts. Unlisted scripts are JS/TS files in the entrypoints directory that aren't content scripts or background scripts.

I see. I think I had to add that because those files weren't being packaged into the build w/o `defineUnlistedScript`. Let me try removing it and see what happens.

leos:
Everything works great without `defineUnlistedScript`, thanks again!

aklinker1:
As long as you import them, they'll be bundled into the final output. They may not showup as separate files or have the same file name, but if you're importing them, they'll be there.

PR #501, opened 2024-02-28
chore: Use JSZip for `wxt zip`, enabling future features

aklinker1:
Pulling out of #223.

### DIFF
--- a/src/core/zip.ts
+++ b/src/core/zip.ts
@@ -9,7 +9,7 @@ import { printFileList } from '~/core/utils/log/printFileList';
 import { internalBuild } from '~/core/utils/building';
 import { registerWxt, wxt } from './wxt';
 import JSZip from 'jszip';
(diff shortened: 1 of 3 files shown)


src/core/zip.ts
```diff
@@ -81,3 +72,49 @@ export async function zip(config?: InlineConfig): Promise<string[]> {
...
+  directory: string,
+  outputPath: string,
+  options?: {
+    include?: string[];
+    exclude?: string[];
+    transform?: (
+      file: string,
+      content: string,
+    ) => Promise<string | undefined | void> | string | undefined | void;
+    additionalWork?: (archive: JSZip) => Promise<void> | void;
```
aklinker1, line 85: Specifically I've added the `transform` option to transform the file before zipping, and the `additionalWork` option to add generated files to the zip.

PR #507, opened 2024-02-29
feat: Download and override private packages for Firefox code review

### LINKED ISSUE #185: Download private packages for Firefox code review

aklinker1:
List package names in your `wxt.config.ts`, and WXT will download them and include them in the sources ZIP when running `wxt zip -b firefox`.

https://github.com/wxt-dev/wxt/blob/d4867c15a7a786eba345742a8a8a7baafaf9b5ed/demo/wxt.config.ts#L18-L20

For this to work, regardless of the project's package manager in use, NPM must be installed and .npmrc must include registry auth tokens. `npm pack <package-name>` is always used to download the tarball, even if you use bun/pnpm/yarn.

Will also add overrides/resolutions to your `package.json` (in the ZIP, not the one on disk) so the downloaded modules are used instead of the local ones. Also supports if multiple versions of a dependency need downloaded.

This replaces #223, and closes #185.

## Todo

- [x] Download and update resolutions
- [x] Support multiple versions of a downloaded package
- [x] Add E2E test
- [x] Try and move `local_modules` to `.wxt/local_modules`


aklinker1:
Add docs about .npmrc requirements

PR #520, opened 2024-03-03
fix(zip): List all private packages correctly in a PNPM workspace

aklinker1:
If a private package is in the root `package.json`, it was not discovered. Now it is, using `pnpm ls -r` instead of just `pnpm ls`. This is valid for non-workspace repos as well.


aklinker1:
Released in [`v0.17.6`](https://github.com/wxt-dev/wxt/releases/tag/v0.17.6)

PR #540, opened 2024-03-10
docs: Add examples for GitHub Actions

GuiEpi:
Here is my proposal for the github action in the publishing section of the documentation.

### Jobs
1. `Version`
Checks for changes in the `package.json` file. If there is a change, it retrieves the version of your extension.
3. `Submit`
 Submits your extension to the store using the `wxt submit` command.

`pnpm` and `npm` code-group.

Here's a[ link to my CI Deploy](https://github.com/GuiEpi/plex-skipper/actions/runs/8222114482/job/22483242808) to show you how it works.
I follow the [pnpm continuous integration guide ](https://pnpm.io/continuous-integration#github-actions) and your [publishing](https://wxt.dev/guide/publishing.html) section.

### DIFF
--- a/docs/guide/publishing.md
+++ b/docs/guide/publishing.md
@@ -114,12 +114,14 @@ jobs:
 ```
 
 :::tip Advanced Automation
+
 The action above lays the foundation for basic automation processes, including `zip` and `submit` actions. To further enhance your GitHub Action capabilities and delve into more complex scenarios, consider exploring the following examples. They introduce advanced features such as version management and GitHub releases, tailored for different needs:
 
 - [Manual submission](https://github.com/aklinker1/github-better-line-counts/blob/main/.github/workflows/submit.yml): For scenarios requiring control and manual intervention.
 - [Automatic submission](https://github.com/GuiEpi/plex-skipper/blob/main/.github/workflows/deploy.yml): For fully automated update and deployment cycles.
-  > These examples are designed to provide clear insights and are a good starting point for customizing your own workflows. Feel free to explore and adapt them to your project needs.
-  > :::
+
+> These examples are designed to provide clear insights and are a good starting point for customizing your own workflows. Feel free to explore and adapt them to your project needs.
+> :::
 
 ## Stores
 
--- a/docs/guide/publishing.md
+++ b/docs/guide/publishing.md
@@ -66,13 +66,7 @@ See the [Firefox Addon Store](#firefox-addon-store) section for more details abo
 
 Here's an example of a GitHub Action to automate submitting new versions of your extension for review. Ensure that you've added all required secrets used in the workflow to the repo's settings.
 
-:::code-group
-
-```yml [pnpm]
-env:
-  DIRECTORY: .output
-  PROJECT_NAME: myextension
-
+```yml
 name: Release
 
 on:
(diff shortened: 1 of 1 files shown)


docs/guide/publishing.md
```diff
@@ -66,10 +66,165 @@ See the [Firefox Addon Store](#firefox-addon-store) section for more details abo
...
+        uses: actions/setup-node@v4
+        with:
+          node-version: 20
+          cache: 'pnpm'
+      - name: Install dependencies
+        run: pnpm install
+      - name: Build extensions
+        run: |
+          pnpm run build && 
+          pnpm run build:firefox
```
aklinker1, line 129: This step isn't necessary. Extension is built during ZIP.

```suggestion
```

docs/guide/publishing.md
```diff
@@ -66,10 +66,165 @@ See the [Firefox Addon Store](#firefox-addon-store) section for more details abo
...
+          FIREFOX_EXTENSION_ID: ${{ secrets.FIREFOX_EXTENSION_ID }}
+          FIREFOX_JWT_ISSUER: ${{ secrets.FIREFOX_JWT_ISSUER }}
+          FIREFOX_JWT_SECRET: ${{ secrets.FIREFOX_JWT_SECRET }}
+```
+
+```yml [npm]
+env:
+  DIRECTORY: .output
+  PROJECT_NAME: myextension
+
```
aklinker1, line 154: Do you think we need multiple examples for package managers? Or should we just list one?

docs/guide/publishing.md
```diff
@@ -66,10 +66,165 @@ See the [Firefox Addon Store](#firefox-addon-store) section for more details abo
...
+        uses: actions/setup-node@v4
+        with:
+          node-version: 20
+          cache: npm
+      - name: Install dependencies
+        run: npm install
+      - name: Build extensions
+        run: |
+          npm run build && 
+          npm run build:firefox
```
aklinker1, line 205:
```suggestion
```

aklinker1 (changes requested):
Looks good. Here's the workflow I was gonna add, for reference: https://github.com/aklinker1/github-better-line-counts/blob/main/.github/workflows/submit.yml

I guess I just forgot to add it? I wrote that a while ago.

It generates a changelog and github release based on conventional commits. I wonder what the best way to document different versioning strategies is?

docs/guide/publishing.md
GuiEpi: Could there be a section called "dispatch" and another called "by commit"?

_Sorry, I made a mistake above when I thought I was commenting on your answer...._

docs/guide/publishing.md
GuiEpi: I did it, but I think it would be better to keep only pnpm.

GuiEpi:
Could there be a section called "dispatch" and another called "by commit"?

aklinker1:
Maybe we just list the two steps that are important (zip and submit), then link to each of our projects for a full workflow?

GuiEpi:
I think it's a good idea, it could look like what I've pushed.

aklinker1 (approved):
@GuiEpi I updated the wording a little bit, pulled the examples out of the info block, and listed the actual features of the two example workflows so it's more clear what the difference is.

Thanks for adding this for me!

PR #543, opened 2024-03-11
docs: Add documentation for Bun.sh

Pietrucci-Blacher:
Hello, here is my little proposal to add Bun.sh in the current documentation.

### Updated sections : 
- Installation 
- Publishing

I'll be happy to listen to any modifications, especially regarding the positioning in the code-group, where I'm not entirely sure.


aklinker1 (approved):
Thanks! Putting it last is fine

PR #550, opened 2024-03-15
chore: Use `@aklinker1/check` to simplify checks setup

aklinker1:


aklinker1:
Need to test on windows

PR #556, opened 2024-03-19
fix(init): Don't overwrite existing files when bootstrapping a new project

btea:
When initializing a project in a certain folder, if a `.gitignore` file exists in the project, a warning message will appear.

image

I still have a question, when the folder where we initialize the project contains other content, do we need to clear the content before downloading the template? 🤔

### DIFF
--- a/src/core/initialize.ts
+++ b/src/core/initialize.ts
@@ -63,7 +63,7 @@ export async function initialize(options: {
   if (isExists) {
     const isEmpty = (await fs.readdir(input.directory)).length === 0;
     if (!isEmpty) {
-      consola.warn(`The directory ${input.directory} is not empty. Aborted.`);
+      consola.error(`The directory ${path.resolve(input.directory)} is not empty. Aborted.`);
       process.exit(1);
     }
   }
--- a/src/core/initialize.ts
+++ b/src/core/initialize.ts
@@ -59,6 +59,14 @@ export async function initialize(options: {
   input.template ??= defaultTemplate;
   input.packageManager ??= options.packageManager;
 
+  const isExists = await fs.pathExists(input.directory);
+  if (isExists) {
+    const isEmpty = (await fs.readdir(input.directory)).length === 0;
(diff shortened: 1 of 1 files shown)


aklinker1 (approved):
Thanks. 

> I still have a question, when the folder where we initialize the project contains other content, do we need to clear the content before downloading the template? 🤔

giget will overwrite any files that already exist, but won't delete other files. So no, but it is still a dangerous operation.

We could add a check to see if the dir isn't empty, and confirm with the user before continuing if so.

btea:
It can be added to this PR. But if the directory is not empty and the user chooses to continue, should we clear the directory directly or keep giget's default behavior?

aklinker1:
> if the directory is not empty and the user chooses to continue, should we clear the directory directly or keep giget's default behavior?

@btea Looks like Nuxt just errors out if the directory has something in it, then they don't have to deal with that. So lets do the same. Just log an error and exit with code 1 if the directory isn't empty.

Screenshot 2024-03-20 at 7 52 37 AM

src/core/initialize.ts
```diff
@@ -59,6 +59,14 @@ export async function initialize(options: {
   input.template ??= defaultTemplate;
   input.packageManager ??= options.packageManager;
 
+  const isExists = await fs.pathExists(input.directory);
+  if (isExists) {
+    const isEmpty = (await fs.readdir(input.directory)).length === 0;
+    if (!isEmpty) {
+      consola.warn(`The directory ${input.directory} is not empty. Aborted.`);
```
aklinker1, line 66: Let's use `consola.error` and the absolute path instead.

```suggestion
      consola.error(`The directory ${path.resolve(input.directory)} is not empty. Aborted.`);
```

aklinker1 (approved):
Thanks, one small change

aklinker1:
Released in [v0.17.8](https://github.com/wxt-dev/wxt/releases/tag/v0.17.8)

PR #563, opened 2024-03-23
chore: Add a test covering `wxt init` in a non-empty directory

btea:
Supplement #556

### DIFF
--- a/e2e/tests/init.test.ts
+++ b/e2e/tests/init.test.ts
@@ -2,6 +2,7 @@ import { describe, it, expect } from 'vitest';
 import { TestProject } from '../utils';
 import { execaCommand } from 'execa';
 import glob from 'fast-glob';
+import { mkdir, writeJson } from 'fs-extra';
 
 describe('Init command', () => {
   it('should download and create a template', async () => {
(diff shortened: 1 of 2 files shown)


src/core/initialize.ts
```diff
@@ -155,9 +155,6 @@ async function cloneProject({
       .move(
         path.join(directory, '_gitignore'),
         path.join(directory, '.gitignore'),
-        {
-          overwrite: true,
-        },
       )
```
btea, line 158: If the directory is not empty, the code cannot be executed here, and the logic here can be removed.

e2e/tests/init.test.ts
```diff
@@ -42,4 +42,20 @@ describe('Init command', () => {
...
     `);
   });
+
+  it('should throw an error if the directory is not empty', async () => {
+    const project = new TestProject();
+
+    await execaCommand(`pnpm -s wxt init ${project.root} -t vue --pm npm`, {
+      env: { ...process.env, CI: 'true' },
+      stdio: 'ignore',
+    });
```
aklinker1, line 52:
```suggestion
    await fs.writeJson(project.resolve('package.json'), {});
```

Init E2E tests are slow, so rather than bootstrapping twice, creating a single file in the directory would speed them up.

Specifically, I've had problems with init tests timing out in GitHub actions when ran on windows. The install step is super slow.

aklinker1 (changes requested):

e2e/tests/init.test.ts
btea: Makes sense.  

However, the `project.resolve` method does not exist, I replaced it with `project.resolvePath`. And there is still a problem after the replacement, `fs.writeJSON` will throw an error, `No such file...`

aklinker1: Oh yeah, it's `resolvePath`. The no such file error was because the project directory didn't exist yet.

aklinker1 (approved):
I updated the test

aklinker1:
Released in [v0.17.8](https://github.com/wxt-dev/wxt/releases/tag/v0.17.8)

btea:
Thanks!

PR #564, opened 2024-03-24
feat(analysis): Open `stats.html` file automatically

### LINKED ISSUE #561: Ability to Auto Open Browser with `--analyze` Build Flag
### Feature Request

Like https://github.com/KusStar/vite-bundle-visualizer allows, it'd be nice if `wxt` could open our default browser to view the bundle analysis automatically, instead of having to open the generated HTML file manually.

Maybe via a `--open` flag in addition to the `--analyze` flag, or a seperate `--analyze-open` flag?

N/A

N/A

### Additional context

N/A

aklinker1:
This closes #561.

Either set `analysis.open` in `wxt.config.ts` or run `wxt build --analyze --analyze-open`.

### Todo

- [x] Don't open browser in CI


src/cli/commands.ts
```diff
@@ -66,7 +67,12 @@ cli
         manifestVersion: flags.mv3 ? 3 : flags.mv2 ? 2 : undefined,
         configFile: flags.config,
         debug: flags.debug,
-        analysis: flags.analyze ? { enabled: true } : undefined,
+        analysis: flags.analyze
+          ? {
+              enabled: true,
+              open: flags.analyzeOpen,
+            }
+          : undefined,
```
joealden, line 75: @aklinker1 how hard would it be to respect `analysis.open` from the config file if `--analyzeOpen` isn't passed?

aklinker1: It should already behave that way. The CLI flags override the relevant options in the config file

I would recommend setting the value in the config file if you always want to open it.

joealden: @aklinker1 this is not how things are behaving for me locally (on 0.17.8) - to clarify, my config looks like this:

```
analysis: { outputFile: "build/stats.html", open: true }
```

and when I run `wxt build --analyze`, `analysis.outputFile` is respected, but `analysis.open` is not.

FYI, I don't specify `analysis.enable` in my config as I only want to do this in a specific NPM script.

Could it be that the override logic isn't taking into account that `flags.analyzeOpen` is `undefined`, but `analysis.open` still exists if `flags.analyze` does (like how `"prop" in { prop: undefined }` is `true`)?

aklinker1: Interesting, let me take a look. You're setup is what I intended, so I'll fix it

aklinker1: Yup, the `undefined` was overriding the `true`:

Screenshot 2024-03-25 at 12 11 08 AM

> `userConfig` comes from the wxt.config.ts file, `inlineConfig` is the CLI flags, and merged is the output of merging them together.

Caused by the spread operator, replaced with `defu` and it's working: https://github.com/wxt-dev/wxt/pull/567

aklinker1: @joealden Released fix in [v0.17.9](https://github.com/wxt-dev/wxt/releases/tag/v0.17.9)

PR #566, opened 2024-03-24
feat: Add `{{mode}}` Template Variable

GuiEpi:
### Add `{{mode}}` Template Variable

- Fix do not [escape Vue interpolations](https://vitepress.dev/guide/using-vue#escaping)
- Add test correctly applies template variables for zip file names

### DIFF
--- a/e2e/tests/zip.test.ts
+++ b/e2e/tests/zip.test.ts
@@ -71,4 +71,34 @@ describe('Zipping', () => {
       }"
     `);
   });
+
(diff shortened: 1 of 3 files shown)


aklinker1 (approved):
Thanks for opening the PR! I simplified the test (removing the download package stuff, since that's already tested in the previous test).

> - Fix do not [escape Vue interpolations](https://vitepress.dev/guide/using-vue#escaping)

Ahh, nice, I completely missed that lol

PR #567, opened 2024-03-25
fix: Don't override `wxt.config.ts` options when CLI flags are not passed

aklinker1:
This closes https://github.com/wxt-dev/wxt/pull/564#discussion_r1536867522, see thread for more details.

Uses defu to apply default values instead of spread to not spread undefined values over real values. Replaced 4 spreads inside the config:

- `analysis` - This is the only object effected by CLI flags and is causing a problem today, the rest are just for safety.
- `alias`
- `experimental`
- `dev`

### DIFF
--- a/src/core/utils/building/resolve-config.ts
+++ b/src/core/utils/building/resolve-config.ts
@@ -54,11 +54,6 @@ export async function resolveConfig(
   // Merge it into the inline config
 
   const mergedConfig = mergeInlineConfig(inlineConfig, userConfig);
-  console.log({
-    userConfig: userConfig.analysis,
-    inlineConfig: inlineConfig.analysis,
-    mergedConfig: mergedConfig.analysis,
-  });
 
   // Apply defaults to make internal config.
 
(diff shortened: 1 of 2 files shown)


src/core/utils/building/resolve-config.ts
```diff
@@ -241,24 +241,15 @@ function mergeInlineConfig(
...
-    },
-    alias: {
-      ...userConfig.alias,
-      ...inlineConfig.alias,
-    },
-    experimental: {
-      ...userConfig.experimental,
-      ...inlineConfig.experimental,
-    },
+    analysis: defu(inlineConfig.analysis ?? {}, userConfig.analysis ?? {}),
```
aklinker1, line 244: Hmm, other than some one-off values, everything here is defu-able... Instead of doing individual defus, I could probably do a single one?

aklinker1: Released in [v0.17.9](https://github.com/wxt-dev/wxt/releases/tag/v0.17.9)

PR #577, opened 2024-03-29
feat: Configure dev server port with `dev.server.port`

zizhengtai:
Added the `dev.port` field to `InlineConfig` to allow configuring the dev server port.

### DIFF
--- a/src/cli/__tests__/index.test.ts
+++ b/src/cli/__tests__/index.test.ts
@@ -116,6 +116,20 @@ describe('CLI', () => {
       });
     });
 
+    it('should respect passing --port', async () => {
+      const expectedPort = 3100;
+      mockArgv('--port', String(expectedPort));
+      await importCli();
+
+      expect(createServerMock).toBeCalledWith({
+        dev: {
+          server: {
+            port: expectedPort,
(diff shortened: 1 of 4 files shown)


aklinker1:
Thanks for the PR! But I forgot, it's not that easy lol...

`inlineConfig` [comes from the CLI](https://github.com/wxt-dev/wxt/blob/main/src%2Fcli%2Fcommands.ts#L30-L38), so as it is now, unless we add a `--port` CLI flag, `inlineConfig.dev.port` is always `undefined`.

All config is merged and default values are applied inside `resolveConfig`, then we use the resolved config everywhere (accessible via `wxt.config`). So we should also add the port to the `ResolvedConfig` type and grab it from there. BUT to resolve the config, we need the server, and to create the server, we need the port. So there's kind of a dependency loop there.

Let me refactor a little and break that loop, and I'll commit my changes to this PR.

aklinker1:
Alright, fixed that issue and merged your PR with the latest. I ended up putting the config under `dev.server.port`, instead of just `dev.port`.

aklinker1 (approved):

Issue #506, opened 2024-02-29
Extract entrypoint options improvements


aklinker1:
This is an issue for addressing several issues with how entrypoints are imported to get their options.

- [x] #336
- [x] #504
- [x] #634
- [x] #635
- [x] #824

> Edit: As of `v0.19.0-alpha1`, all the above issues work with:
> ```ts
> export default defineConfig({
>   experimental: {
>     entrypointImporter: "vite-node",
>   },
> });
> ```

Will add to this list if other issues come up.

aklinker1:
Alright, I have a proof of concept using Vite's runtime API to load entrypoints: https://github.com/wxt-dev/wxt/pull/585

- Supports custom transformations via vite plugins
- Lets you import variables or use `import.meta.glob`, a vite feature

So this should work for everyone's use-case.

HOWEVER... it appears the vite runtime can't import CSS modules, which is kinda important. I've asked a question about it in the vite discord, so we'll se if I get a useful response. https://discord.com/channels/804011606160703521/1236142702785990677/1236142702785990677

aklinker1:
Alright, it was an issue on my side. Just needed to start the vite server lol. So it's working now. 

I need to clean up my changes before opening a real PR. Probably won't get to it this weekend... we'll see. But I got everything working :)

aklinker1:
Last problem to address... Module side effects.

Say you're using storage item in your content script:

```ts
// utils/storage.ts
export const isXyzEnabled = storage.defineItem(...);
```

```ts
// entrypoints/content.ts
import { isXyzEnabled } from '~/utils/storage';

export default defineContentScript({
  // ...
  async main() {
    console.log(await isXyzEnabled.getValue());
  }
});
```

In this example, `storage.defineItem` actually includes side effects, like checking if the storage item needs migrated. Should a side effect like this run? I don't think so, and I can't think of reason a side effect like this should ever run. That is, unless it's related to an imported variable used in the options.

So I have a decision to make. When importing entrypoints during the build to get their config, do I...
1. delete the main function and remove unused imports?
2. Leave everything as is and let all side-effects run

I think... I'm gonna go with option 1, and remove the main function. So I have to decide the best way to do that. Removing the main function should be simple... Use magicast or regex. To remove unused imports after that... I'm not sure how to do that. Regex again?

aklinker1:
[...]

PR #585, opened 2024-04-01
experimenting with vite runtime for loading entrypoint config

aklinker1:
Potential fix to #506

- [x] Load entrypoints with [Vite Runtime API](https://vitejs.dev/guide/api-vite-runtime)
- [x] use `magicast` to remove `main` function from entrypoints to speed up imports
   - [ ] ~~And maybe remove unused imports? Not sure the best way to handle side-effects...~~ No, let vite handle tree-shaking? That's beyond the scope of my knowledge lol.


aklinker1:
Will reference this PR in official PR.

PR #608, opened 2024-04-11
doc: Fix bad extension API examples

couriourc:
is it new version,but no doc?,there is no `browser.runtime.onInstall.addEventListener` or `browser.storage.local.setItem`,but there is `browser.runtime.onInstalled.addListener` and `browser.storage.local.set`,and it does work


aklinker1 (approved):
Thanks! I probably just typed it up by memory and got it wrong lol

PR #644, opened 2024-05-03
【draft】repro of issue #645

1natsu172:


aklinker1:
Great reproduction, thank you! You're the first person to use a test case, which is so nice. But like I said on the issue, this is intended.

PR #650, opened 2024-05-05
docs: Fix `wxt-vitest-plugin` reference

emmanuel-ferdman:
**PR Summary**:
PR #646 moved the location of the `wxt-vitest-plugin.ts` file. This PR adjusts sources to changes.


aklinker1 (approved):
Ah, thanks. Looks like that's the only reference that wasn't moved

PR #655, opened 2024-05-07
feat(runner): Add `keepProfileChanges` option

EdoanR:
This will add the [`keepProfileChanges`](https://extensionworkshop.com/documentation/develop/web-ext-command-reference/#keep-profile-changes) to the runner config.

I was trying to use the `--keep-profile-changes` in the `args` but it was not working, so I added as a option for the runner, just like `startUrls`.

Usage:

```ts
// wxt.config.ts
import { defineConfig } from 'wxt';

export default defineConfig({
  ...
  runner: {
    keepProfileChanges: true
  },
});
```

or

```ts
// web-ext.config.ts
import { defineRunnerConfig } from 'wxt';

export default defineRunnerConfig({
  keepProfileChanges: true
});
```

### DIFF
--- a/packages/wxt/src/core/runners/web-ext.ts
+++ b/packages/wxt/src/core/runners/web-ext.ts
@@ -35,6 +35,7 @@ export function createWebExtRunner(): ExtensionRunner {
         console: wxtUserConfig?.openConsole,
         devtools: wxtUserConfig?.openDevtools,
         startUrl: wxtUserConfig?.startUrls,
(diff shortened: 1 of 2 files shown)


aklinker1 (approved):
Oh good catch, not sure how I forgot that one initially

aklinker1:
Released in [`v0.18.2`](https://github.com/wxt-dev/wxt/releases/tag/v0.18.2)

PR #708, opened 2024-06-08
fix(storage): Add storage area typing for storage keys

TheOnlyTails:
This introduces new types that enforce the `area:key` format of storage keys by the WXT storage APIs.

### DIFF
--- a/packages/wxt/src/__tests__/storage.test.ts
+++ b/packages/wxt/src/__tests__/storage.test.ts
@@ -849,13 +849,15 @@ describe('Storage Utils', () => {
         });
         expectTypeOf(item).toEqualTypeOf<WxtStorageItem<number | null, {}>>();
       });
+    });
+  });
 
-      it('should not accept keys without a valid storage area prefix', () => {
-        // @ts-expect-error: Test passes if there is a type error here
-        storage.getItem('test');
-        // @ts-expect-error
-        storage.getItem('loca:test');
-      });
+  describe('types', () => {
+    it('should not accept keys without a valid storage area prefix', async () => {
+      // @ts-expect-error: Test passes if there is a type error here
+      await storage.getItem('test').catch(() => {});
+      // @ts-expect-error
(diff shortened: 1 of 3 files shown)


pnpm-lock.yaml
aklinker1, line 1: I don't think any of these changes are relevant, can you revert the changes to this file?

packages/wxt/src/__tests__/storage.test.ts
aklinker1, line 1: Could you add another test with a bad key to ensure types are working? Something like this should be good enough, can probably go outside the main describe for each storage area

```ts
it("should not accept keys without a valid storage area prefix", () => {
  // @ts-expect-error: Test passes if there is a type error here
  storage.getItem("test");
});
```

aklinker1 (changes requested):
Thanks for the PR! Good change.

A couple of changes and I can get this merged!

packages/wxt/src/__tests__/storage.test.ts
TheOnlyTails: Done!

pnpm-lock.yaml
TheOnlyTails: done

TheOnlyTails:
This is a bit of a nitpick, but do you know why the auto-generated docs show this in the definition of `StorageItemKey`?
image
https://deploy-preview-708--creative-fairy-df92c4.netlify.app/api/reference/wxt/storage/type-aliases/StorageItemKey.html#type-alias-storageitemkey

aklinker1 (approved):
> This is a bit of a nitpick, but do you know why the auto-generated docs show this in the definition of `StorageItemKey`?
> image

Haha nope, that's crazy. I probably need to upgrade the tool, but it's touchy and breaks horribly whenever I try.

aklinker1:
Released in [`v0.18.5`](https://github.com/wxt-dev/wxt/releases/tag/wxt-v0.18.5)

PR #709, opened 2024-06-10
docs: Simplify menu configuration

davestewart:
This is just a little stitch-in-time PR with some menu config cleanup:

- utility functions replace raw sidebar menu config
- utility functions replace raw head config

This makes it simpler to [grok and update the sidebar menu](https://github.com/davestewart/wxt/blob/feat/docs-cleanup/docs/.vitepress/config.ts#L73) structure:

```ts
sidebar: {
  '/guide/': menuRoot([
    menuGroup('Directory Structure', '/guide/directory-structure/', [
      menuItem('.output/', 'output'),
      menuItem('composables/', 'composables'),
      menuGroup('entrypoints/', '/guide/directory-structure/entrypoints/', [
        menuItem('background', 'background.md'),
        menuItem('bookmarks', 'bookmarks.md'),
        ...
      ]),
      ...
  ]),
}
```

Additionally, some minor style updates:

- main sidebar sections now use correct level 0 (this can be [styled](https://github.com/wxt-dev/wxt/pull/709/files#diff-096f06e4edeeaa293d906d7146e243667e1377d2133e35b9e240525bd79be5b5R88-R94)):<br>CleanShot 2024-06-10 at 13 32 37
- typedoc sidebar config now [shows badges](https://github.com/wxt-dev/wxt/pull/709/files#diff-702f6a4b41fe8e3d3094e575522859877a8cfb040263991f958de4768a6b3f8fR66-R69) for leaf-level collections:<br>CleanShot 2024-06-10 at 13 31 17

I note the original config used an interim `{ items: [ ... ] }` node to restyle the menu, but we can just do that in `theme/custom.css` if you want to tweak it further.

### DIFF
--- a/docs/.vitepress/config.ts
+++ b/docs/.vitepress/config.ts
@@ -116,7 +116,7 @@ export default defineConfig({
           ]),
           menuItem('hooks/', 'hooks'),
           menuItem('modules/', 'modules'),
-          menuGroup('public/', 'public', [
+          menuItem('public/', 'public/', [
             menuItem('_locales/', 'public/locales'),
           ]),
           menuItem('utils/', 'utils'),
--- a/docs/.vitepress/config.ts
+++ b/docs/.vitepress/config.ts
@@ -115,7 +115,6 @@ export default defineConfig({
             menuItem('*.ts', 'unlisted-scripts.md'),
           ]),
           menuItem('hooks/', 'hooks'),
-          menuItem('modules/', 'modules'),
           menuItem('public/', 'public/', [
             menuItem('_locales/', 'public/locales'),
           ]),
--- a/docs/.vitepress/config.ts
+++ b/docs/.vitepress/config.ts
@@ -119,7 +119,7 @@ export default defineConfig({
           menuGroup('public/', 'public', [
             menuItem('_locales/', 'public/locales'),
           ]),
-          menuItem('utils/', 'modules'),
+          menuItem('utils/', 'utils'),
 
           // Files
           menuItem('.env', 'env'),
--- a/docs/.vitepress/config.ts
+++ b/docs/.vitepress/config.ts
@@ -1,22 +1,16 @@
-import { DefaultTheme, defineConfig } from 'vitepress';
+import { defineConfig } from 'vitepress';
 import typedocSidebar from '../api/reference/typedoc-sidebar.json';
-
-const filteredTypedocSidebar = typedocSidebar.filter(
-  (item) => item.text !== 'API',
-);
-// Typedoc's markdown theme adds collapse: true to all our items, event ones without any children,
-// so they need to be removed.
-function removeCollapsedWithNoItems(items: DefaultTheme.SidebarItem[]) {
-  for (const item of items) {
-    if (item.items) removeCollapsedWithNoItems(item.items);
-    else delete item.collapsed;
-  }
-}
-removeCollapsedWithNoItems(filteredTypedocSidebar);
+import {
+  menuGroup,
+  menuItem,
+  menuRoot,
+  navItem,
+  prepareTypedocSidebar,
+} from './utils/menus';
+import { meta, script } from './utils/head';
 
 const title = 'Next-gen Web Extension Framework';
 const titleSuffix = ' – WXT';
-
 const description =
   "WXT provides the best developer experience, making it quick, easy, and fun to develop chrome extensions for all browsers. With built-in utilities for building, zipping, and publishing your extension, it's easy to get started.";
 const ogTitle = `${title}${titleSuffix}`;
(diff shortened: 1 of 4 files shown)


docs/.vitepress/config.ts
```diff
@@ -37,20 +31,16 @@ export default defineConfig({
...
-        'data-website-id': 'c1840c18-a12c-4a45-a848-55ae85ef7915',
-        src: 'https://umami.aklinker1.io/script.js',
-      },
-    ],
+    meta('og:type', 'website'),
+    meta('og:title', ogTitle),
+    meta('og:image', ogImage),
+    meta('og:url', ogUrl),
+    meta('og:description', description),
+    meta('twitter:card', 'summary_large_image', true),
```
aklinker1, line 39: Don't really like this third argument here, it's not clear just from looking at it what the third argument is doing. Not a huge problem, since it will probably never be used again.. but I would prefer two functions, `metaProperty` and `metaName` or just have the key be the first argument, like `meta('name', 'twitter:card', 'summary_large_image')`. Or make the third argument an object so the boolean can be labeled: `meta('twitter:card', 'summary_large_image', { useName: true })`

docs/.vitepress/utils/menus.ts
```diff
@@ -0,0 +1,81 @@
...
+  const filtered = items.slice(1);
+
+  // remove Typedoc's collapse: true from text nodes
+  const prepareItems = (items: DefaultTheme.SidebarItem[], depth = 0) => {
+    for (const item of items) {
+      if (item.items) {
+        prepareItems(item.items, depth + 1);
+        const hasLeaves = item.items.some((item) => !item.items);
+        if (hasLeaves) {
+          item.text += ` <span class="badge">${item.items.length}</span>`;
```
aklinker1, line 68: Didn't know this was interpreted as innerHTML, nice!

docs/.vitepress/config.ts
```diff
@@ -59,170 +49,116 @@ export default defineConfig({
...
+            menuItem('sidepanel', 'sidepanel.md'),
+            menuItem('*.html', 'unlisted-pages.md'),
+            menuItem('*.ts', 'unlisted-scripts.md'),
+          ]),
+          menuItem('hooks/', 'hooks'),
+          menuItem('modules/', 'modules'),
+          menuGroup('public/', 'public', [
+            menuItem('_locales/', 'public/locales'),
+          ]),
+          menuItem('utils/', 'modules'),
```
aklinker1, line 122: Mistake from my previous PR.

```suggestion
          menuItem('utils/', 'utils'),
```

docs/.vitepress/config.ts
```diff
@@ -59,170 +49,116 @@ export default defineConfig({
...
+            menuItem('options', 'options.md'),
+            menuItem('popup', 'popup.md'),
+            menuItem('sandbox', 'sandbox.md'),
+            menuItem('sidepanel', 'sidepanel.md'),
+            menuItem('*.html', 'unlisted-pages.md'),
+            menuItem('*.ts', 'unlisted-scripts.md'),
+          ]),
+          menuItem('hooks/', 'hooks'),
+          menuItem('modules/', 'modules'),
+          menuGroup('public/', 'public', [
```
aklinker1, line 119:
```suggestion
          menuGroup('public/', '/guide/directory-structure/public/', [
```

aklinker1 (changes requested):
I like the appearance of the sidebar much more now!

Couple of bugs I noticed:
1. The `guide/directory-structure/public/locales` link is broken. It takes me to `/publicpublic/locales.html` instead of `/guide/directory-structure/public/locales.html`
   Screenshot 2024-06-10 at 10 21 18 AM
2. Can't access `/guide/directory-structure/public/index.html` anymore. the menu item wasn't collapsible before, so you could click on it. Now I can only expand/collapse it's children. I'll let you decide how to add it back. This is likely one of the pages with duplicate content that will be moved, so it's fine if it's a temporary solution.

docs/.vitepress/utils/menus.ts
davestewart: Nor did I, I just poked to see what came out the other side!

docs/.vitepress/config.ts
davestewart: So it seems that as long as the parent config has leading and trailing slashes, that the full path will be properly built.

I don't fully understand how the theme is doing this, but it seems that the following picks up the parent `/guide/directory-structure/`:

https://github.com/wxt-dev/wxt/pull/709/files#diff-3def678deb1b1d5a53948eb8491817dde4dc881e032e785fc61cc93d334eefcdR119-R121

```ts
        menuGroup('Directory Structure', '/guide/directory-structure/', [
          ...
          menuItem('public/', 'public/', [
            menuItem('_locales/', 'public/locales'),
          ]),
```

I actually looked to do something within the helpers, but it seems it's not actually needed.

davestewart:
FWIW I find GitHub's resolve changes UI pretty clunky; I'm still not sure if I've clicked all the right buttons, or if the "changes requested" heading is indicating I still need to work, or you need to do work, or both of us need to do work!

LMK

aklinker1 (approved):
If it says requested changes, it's because I blocked the merging of the PR until I approve it. Individual suggestions in comments don't effect that status. So just re-requesting my review was the right thing to do!

Everything looks good now, thanks

PR #716, opened 2024-06-13
feat: Add `@wxt-dev/module-vue` package

aklinker1:
This is my first PR extracting a frontend framework's logic from `wxt` into a module.

Will update the vue template after this PR once the module is released to NPM

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -18,7 +18,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - uses: ./.github/actions/setup
-      - run: pnpm build-deps
+      - run: pnpm -r build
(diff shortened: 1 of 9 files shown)


aklinker1:
After writing the first moudle, a couple of take-aways:

- [ ] second arg options should be optional
- [ ] new helper: addImportPreset
- [ ] new helper: isPackageInstalled
- [ ] export WxtModule type from 'wxt/modules' to avoid "TS2742: The inferred type of 'default' cannot be named without a reference to '../node_modules/wxt/dist/index-q2ZwJBs6'. This is likely not portable. A type annotation is necessary."
- [ ] Double check that types augmentation works for module options without adding an import to the type

PR #727, opened 2024-06-14
 fix: Generalize react-refresh preamble logic to virtualize all inline scripts

aklinker1:
Basically, the react plugin was adding an inline script to the HTML page to enable HMR:

```html
<script type="module">
  import RefreshRuntime from "/@react-refresh"
  RefreshRuntime.injectIntoGlobalHook(window)
  window.$RefreshReg$ = () => {}
  window.$RefreshSig$ = () => (type) => type
  window.__vite_plugin_react_preamble_installed__ = true
</script>
```

However, this doesn't work for chrome extensions because inline scripts are blocked by the CSP, in dev and production.

Before this PR, I had written specific code to find and virtualize this react-refresh "preamble" inline script so it could be loaded from `http://localhost:3000/@wxt:virtualized-react-refresh`. So originally, I wanted to extract this logic into `@wxt-dev/module-react`. But since the `devServerPrerender` plugin (where this react-refresh logic is) is added in a kinda special place, there wasn't a hook for adding a similar plugin in the module yet.

Instead, I realized a better solution would be to generalize the logic in case another plugin adds inline scripts in the future, it would be handled automatically! So that's what this PR does.

### DIFF
--- a/packages/module-vue/README.md
+++ b/packages/module-vue/README.md
@@ -12,6 +12,7 @@ This plugin makes a few changes:
 ## Usage
 
 ```sh
+pnpm i vue
 pnpm i -D @wxt-dev/module-vue
 ```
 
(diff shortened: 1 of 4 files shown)


aklinker1:
Closing this, screwed up commits on this branch.

PR #751, opened 2024-06-18
docs: Improved example search

aklinker1:
Convert to simplified examples repo. Goal here is whenever I create an example for someone, I'll push it to https://github.com/wxt-dev/examples and it will show up in the docs automatically.

I'm not merging this until I move all the examples over to the new repo.

Screenshot 2024-06-18 at 3 57 41 PM


aklinker1:
@davestewart I would love your option on this new search, if it's more helpful than the last.

- New: https://deploy-preview-751--creative-fairy-df92c4.netlify.app/examples.html
- Old: https://wxt.dev/examples.html

davestewart:
Hey,

Interesting idea! Not sure how much value the search offers, though I do like the idea of a separate repo. Interestingly I just checked [Chrome's examples](https://developer.chrome.com/docs/extensions/samples), and they do the same!

I think a left-hand filter would work better. There's certainly enough space on the page to do that as well. Also wrapping checkboxes with labels would make for better UX.

FWIW, Nuxt and its Content module, along with `server/api` routes is another option to do search, vs a script + JSON update. Might be worth considering in the future.

aklinker1:
@davestewart It is based off Chrome's new examples! I really like how to you can quickly find examples that use a specific API, and see the "right" way to use them. I wanted to add the packages filter as well, so people can find examples using different NPM packages.

I can clean up the CSS and move it to the side, just wanted to get a POC out for some feedback.

aklinker1:
@davestewart Hows this?

Screenshot 2024-06-26 at 4 45 16 PM


> Also wrapping checkboxes with labels would make for better UX.

Not really sure what you meant by this, I was already wrapping the labels. This time I tried not wrapping (you can scroll horizontally)

davestewart:
Bedtime here; will take a look first thing!

PR #752, opened 2024-06-19
feat(dev): Reload extension when public files change

### LINKED ISSUE #735: Reload when localization changes
### Feature Request

When saving a change to any `public/_locales/*/messages.json`, the extension should either do a full reload, or it should support HMR somehow. It also needs to support any file pattern, so this also supports the changes in https://github.com/wxt-dev/wxt/pull/343.

A couple of ways to do this:

1. During dev, add a proxy around the `browser.i18n` API. It waits for messages from the dev server, caches the latest translations, and pulls the latest localization from the web. 
   - Couple of problems with this approach: won't automatically update text on page, but the next time you get the localization, it will be updated. That means you'd have to reload the HTML pages or content scripts, or re-trigger logic in the background. ~~The other problem, is how do I know which locale to pull text from.~~ Use [`@@ui_locale`](https://developer.chrome.com/docs/extensions/reference/api/i18n#overview-predefined)
2. Use Vite's HMR API
   - I don't really know how to do this. I would have to create a virtual module that imports all the localization files. It would automatically detect file changes because it would be imported... somewhere? Then I'd have to use the `import.meta.hot` API on the client side to setup changes? Once again, no idea how this would work.

I'm gonna go with option 1, using the web-socket to notify when code changes and cache the latest values client side. For UI's, we can add framework-specific utils around automatically re-rendering when the text changes. I don't think there's a single-solution-fits-all available here. Until those custom utils are implemented, which likely won't be in the first PR, you'll just have to refresh the HTML page.

N/A

Stopping the dev command and restarting it.

### Additional context

- https://github.com/wxt-dev/wxt/pull/343

aklinker1:
This closes #735. When you save, add, or delete a file in the public directory, the extension will reload. This includes localization files.


packages/wxt/src/core/utils/building/detect-dev-changes.ts
```diff
@@ -50,7 +50,20 @@ export function detectDevChanges(
...
+  if (changedSteps.size === 0) {
+    const hasPublicChange = some(changedFiles, (file) =>
+      file.startsWith(wxt.config.publicDir),
+    );
+    if (hasPublicChange) {
+      return {
+        type: 'extension-reload',
+        rebuildGroups: [],
+        cachedOutput: currentOutput,
+      };
```
aklinker1, line 62: TEST

packages/wxt/src/core/utils/building/detect-dev-changes.ts
```diff
@@ -50,12 +50,31 @@ export function detectDevChanges(
...
+  console.log({ changedSteps });
+  if (changedSteps.size === 0) {
+    const hasPublicChange = some(changedFiles, (file) =>
+      file.startsWith(wxt.config.publicDir),
+    );
+    console.log({
+      hasPublicChange,
+      changedFiles,
+      publicDir: wxt.config.publicDir,
+    });
```
aklinker1, line 62:
```suggestion
```

packages/wxt/src/core/utils/building/detect-dev-changes.ts
```diff
@@ -104,6 +116,7 @@ export function detectDevChanges(
     };
   }
 
+  console.log(1, unchangedOutput);
```
aklinker1, line 119:
```suggestion
```

PR #755, opened 2024-06-19
fix: Ignore .wxt file changes in dev

aklinker1:
I noticed when you first saved a change to the project, WXT would report `.wxt/tsconfig.json` as a changed file along with the actual file you changed. That's unnecessary, the directory doesn't effect any runtime code. So this PR ignores these files in the watcher.

| Before | After |
| --- | --- |
| Screenshot 2024-06-19 at 8 31 54 AM | Screenshot 2024-06-19 at 8 32 10 AM |

As you can see, this was causing an unnecessary extension reload, when really, the HTML pages just needed reloaded.


aklinker1:
Released in [`v0.18.8`](https://github.com/wxt-dev/wxt/releases/tag/wxt-v0.18.8)

PR #760, opened 2024-06-20
fix(modules): `addViteConfig` ignored user vite config

aklinker1:
If you used `addViteConfig` in a module, it cause WXT to ignore any `vite` config from your `wxt.config.ts` file. Because I forgot to await a promise 🤦


aklinker1:
Released in [`v0.18.8`](https://github.com/wxt-dev/wxt/releases/tag/wxt-v0.18.8)

Issue #710, opened 2024-06-11
Update generated ESLint config file for ESLint 9


KnightYoshi:
### Feature Request

Update ESLint config to support the ESLint 9 structure.

### What are the alternatives?

I have managed to get ESLint to recognize the WXT globals by importing it as:
```
import globals from 'globals';
import wxtimport from './.wxt/eslintrc-auto-import.json' assert { type: 'json' };

export default [
    {
        languageOptions: {
            globals: {
                ...globals.browser,
                ...wxtimport.globals,
            },
        },
    }
```

### Additional context

ESLint 9 has deprecated the `.eslintrc*` files in favor of eslint.config.js, eslint.config.mjs, and eslint.config.cjs and changed the config structure

https://eslint.org/docs/latest/use/configure/configuration-files

Looked through how it's generated and got lost trying to follow the trail 🤣

aklinker1:
Or, a much simpler alternative, just output a second file file along-side the existing file with the following content:

```ts
// .wxt/eslint-v9-config.js
import globals from 'globals';
import v8Config from './.wxt/eslintrc-auto-import.json' assert { type: 'json' };

export default {
        languageOptions: {
            globals: {
                ...globals.browser,
                ...v8Config.globals,
            },
        },
    }
```

And tell people to extend that file for v9 and up. Eventually, I'd like to just generate one file, but this could be a short-term solution.

KnightYoshi:
Thanks for pointing me in the right direction. I don't mind giving it a shot myself. I ended up in the [getUnimportOptions](https://github.com/wxt-dev/wxt/blob/34ea23030b932da7a66f0942d45025fd37bc236d/packages/wxt/src/core/utils/building/resolve-config.ts#L268C16-L268C34) function trying to find where it happens 🙃 

I've looked at how the _globals_ package does it, that package just has an `index.js` script that exports the `globals.json` file. I think we can do something similar here that way it's not backwards breaking.

I'll see about doing this a little later when I have some free

aklinker1:
Released in [`v0.18.11`](https://github.com/wxt-dev/wxt/releases/tag/wxt-v0.18.11)

PR #762, opened 2024-06-21
feat: add eslint 9 config support

KnightYoshi:
Add ESLint 9 support: https://github.com/wxt-dev/wxt/issues/710

With ESLint 9, the config has changed, this generates a ESLint JS file to work with the ESLint 9 config changes.

This also adds ESLint to the wxt-demo package.

There is just one issue to address.
IDE has an ESLint error
`SyntaxError: The requested module './.wxt/eslintrc-auto-import.js' does not provide an export named 'default'`

I think this is because it's not an `index.js` file and is doing a `module.exports` but in the demo package I'm using an `import` statement. If it's safe to now just use an `export` statement in the JS file I can update it to do that instead, or what would be the preferred approach to this?

I think everything else is correct, the Vitest was a bit of a learning curve, luckily there's plenty of other tests to learn from :)


aklinker1:
Thanks for the PR, will review this today!

packages/wxt-demo/wxt.config.ts
```diff
@@ -29,4 +29,9 @@ export default defineConfig({
     // @ts-expect-error: c is not defined, this should error out
     c: 'c',
   },
+  imports: {
+    eslintrc: {
+      enabled: true,
+    },
+  },
```
aklinker1, line 36: Remove this. You changes should be covered by tests.

packages/wxt-demo/package.json
```diff
@@ -23,6 +23,7 @@
   "devDependencies": {
     "@types/react": "^18.3.3",
     "@types/react-dom": "^18.3.0",
+    "eslint": "^9.5.0",
```
aklinker1, line 26: Same here. Create a E2E test that uses eslint 9 instead.

packages/wxt/src/core/utils/building/generate-wxt-dir.ts
```diff
@@ -31,7 +32,8 @@ export async function generateTypesDir(
     const unimport = createUnimport(wxt.config.imports);
     references.push(await writeImportsDeclarationFile(unimport));
     if (wxt.config.imports.eslintrc.enabled) {
-      await writeImportsEslintFile(unimport, wxt.config.imports);
+      const eslintVersion = await getEslintVersion();
+      await writeImportsEslintFile(unimport, wxt.config.imports, eslintVersion);
```
aklinker1, line 36: We can just move this variable into the function

packages/wxt/src/core/utils/building/generate-wxt-dir.ts
```diff
@@ -73,7 +77,26 @@ async function writeImportsEslintFile(
...
+  await fs.writeJson(jsonFilePath, eslintrc, { spaces: 2 });
+
+  await writeEslintNineFile(majorEslintVersion, jsonFilePath);
+}
+
+async function writeEslintNineFile(
+  majorEslintVersion: number,
+  jsonFilePath: string,
+) {
+  if (majorEslintVersion >= 9) {
```
aklinker1, line 91: This if statement should go outside this function. We should only call `writeEslintNineFile` if it is eslint 9. Small readibility thing.

packages/wxt/src/core/utils/eslint.ts
```diff
@@ -0,0 +1,16 @@
+import { resolve } from 'node:path';
+import fs from 'fs-extra';
+import { wxt } from '../wxt';
+
+export async function getEslintVersion(): Promise<string[]> {
+  try {
+    const packageJsonPath = resolve(
+      wxt.config.root,
+      'node_modules/eslint/package.json',
```
aklinker1, line 9: Unfortunately, there is no guarentee that this file exists... Monorepos for example, may not have eslint installed in this extention's package. Instead, we should import it and grab the version from the imports. Something like this:

```ts
const { ESLint } = await import('eslint');
console.log(ESLint.version)
```

aklinker1 (changes requested):
Couple of changes requested. I'll implement a E2E test for this, can you update it so it passes?

```ts
# To run tests:
pnpm test
```

aklinker1:
I see we're just re-exporting the JSON file in a JS file. How is this any different from this? Is a new file necessary?

```diff
// eslint.config.js
- import autoImports from './.wxt/eslintrc-auto-import.js';
+ import autoImports from './.wxt/eslintrc-auto-import.json' with { type: 'json' };

export default [
  {
    languageOptions: {
      globals: {
        ...autoImports.globals,
      },
      sourceType: 'module',
    },
  },
];
```

Instead, should we fully adopt the new config format and do something like this?

```js
import autoImports from './.wxt/eslintrc-auto-import.js';

export default [
  autoImports,
  {
    // User config here...
  },
];
```

```js
// ./.wxt/eslintrc-auto-import.js
import { globals } from './eslintrc-auto-import.json' with { type: 'json' };

export default [
  {
    languageOptions: {
      globals,
      sourceType: 'module',
    },
  }
]
```

aklinker1:
Doing some more testing, and it appears `typescript-eslint` handles this automatically for you. Still a problem if you use JS though, so this PR is still important.

aklinker1:
OK, got a basic example of using the new flat file format


https://github.com/wxt-dev/wxt/assets/10101283/afeca287-38c6-4e9d-b89d-984aff4bd99c

[wxt-eslint-9-example.zip](https://github.com/user-attachments/files/15945360/wxt-eslint-9-example.zip)

I think we should use this format instead.

aklinker1:
Also, I just merged #767, which will effect this PR. I knew it would cause conflicts with this PR when I merged it, but I needed it merged for some other work. Don't worry about fixing the conflicts yourself, I can do it since I broke it.

packages/wxt-demo/package.json
KnightYoshi: Do we even need ESLint included? I had added it for the demo stub, but since that config is being removed, I don't think we need ESLint even added to the demo package

packages/wxt/src/core/utils/building/generate-wxt-dir.ts
KnightYoshi: Fair enough, I looked at it and went "you know an early return would make this easier to read, or could have it outside the function"

packages/wxt/src/core/utils/eslint.ts
KnightYoshi: ESLint is not a friendly import with the `"type": "module"` 

Looked for an alternative option, found the `createRequire` flow, found out that `createRequire` is already declared, so looked to see if and how that was handled and mirrored that to import ESLint, let me know if that works for you

KnightYoshi:
> I see we're just re-exporting the JSON file in a JS file. ~If that's all we're doing, can't we just do this instead?~
> 
> _Edit: ESLint doesn't like the `with { type: 'json' }`_
> 
> ```diff
> // eslint.config.js
> - import autoImports from './.wxt/eslintrc-auto-import.js';
> + import autoImports from './.wxt/eslintrc-auto-import.json' with { type: 'json' };
> 
> export default [
>   {
>     languageOptions: {
>       globals: {
>         ...autoImports.globals,
>       },
>       sourceType: 'module',
>     },
>   },
> ];
> ```
> 
> What about fully adopting the new file config format and do something like this?
> 
> ```js
> import autoImports from './.wxt/eslint-auto-import-config.js';
> 
> export default [
>   autoImports,
>   {
>     // User config here...
>   },
> ];
> ```
> 
> ```js
> // ./.wxt/eslint-auto-import-config.js
> import { globals } from './eslintrc-auto-import.json' with { type: 'json' };
> 
> export default {
>   languageOptions: {
>     globals,
>     sourceType: 'module',
>   },
> }
> ```

The idea I had was keeping it BC, since jumping to ESLint 9+ might be bigger hurdle for some. There _was_ a package, [ESLint import plugin](https://www.npmjs.com/package/eslint-plugin-import) I had that's not ESLint 9 compatible and hasn't been updated in 6 months, and I wasn't sure when it was going to be updated so I opted not to move to ESLint 9 for now. 
I say "was" because the guy ljharb who is a big contributor insists on Node 0.4 compatibility and has caused a bit of an uproar over it, so I opted to remove it 🤣 However, that's not to say others can or will remove ESLint packages that prevent updating to the latest version.

What about splitting it so that it generates >=9+ config and <9 config depending on the ESLint version.

packages/wxt/src/core/utils/building/generate-wxt-dir.ts
KnightYoshi: moved! :)

aklinker1:
> I say "was" because the guy ljharb who is a big contributor insists on Node 0.4 compatibility and has caused a bit of an uproar over it, so I opted to remove it 🤣

... That's an interesting opinion.

> What about splitting it so that it generates >=9+ config and <9 config depending on the ESLint version.

I think this is what we should do. Checking out your latest code now.

packages/wxt/src/core/utils/eslint.ts
aklinker1: Hmm, I thought dynamic imports supported CJS modules. I see your `createRequire` code, if that's what's required, that's good with me 👍

KnightYoshi:
> > I say "was" because the guy ljharb who is a big contributor insists on Node 0.4 compatibility and has caused a bit of an uproar over it, so I opted to remove it 🤣
> 
> ... That's an interesting opinion.
> 
> > What about splitting it so that it generates >=9+ config and <9 config depending on the ESLint version.
> 
> I think this is what we should do. Checking out your latest code now.

Cool, if I have some time today I'll work on splitting it into separate dedicated configs based on the ESLint version, otherwise probably next weekend.

I already liked what the WXT framework has to offer. Working on this small part, and digging into the the internals of how it works has made me appreciate even more all the effort that has gone into it. It's non-trivial effort, so well done

packages/wxt-demo/package.json
aklinker1: No, we should be good. I'll remove the types package.

packages/wxt/src/core/utils/building/resolve-config.ts
```diff
@@ -302,17 +304,6 @@ async function getUnimportOptions(
 ): Promise<WxtResolvedUnimportOptions | false> {
   if (config.imports === false) return false;
 
-  const rawEslintEnabled = config.imports?.eslintrc?.enabled;
-  let eslintEnabled: boolean;
-  switch (rawEslintEnabled) {
-    case undefined:
-    case 'auto':
-      eslintEnabled = await isModuleInstalled('eslint');
```
aklinker1, line 310: Seems like `"auto"` was not working. Using your code though, it's working now!

packages/wxt/src/core/utils/building/resolve-config.ts
```diff
@@ -343,6 +330,40 @@ async function getUnimportOptions(
...
+      eslintEnabled = rawEslintEnabled;
+  }
+
+  return {
+    enabled: eslintEnabled,
+    filePath: path.resolve(
+      wxtDir,
+      eslintEnabled === 9
+        ? 'eslint-auto-imports.mjs'
+        : 'eslintrc-auto-import.json',
```
aklinker1, line 361: Needed to change the default file path for this file based on the version, so I moved the version detection code into the config resolution.

aklinker1 (approved):
OK, this is looking good. I need to test if this actually creates a valid config file, I'll add an E2E test for that later today maybe...

aklinker1:
Released in [`v0.18.11`](https://github.com/wxt-dev/wxt/releases/tag/wxt-v0.18.11)

PR #771, opened 2024-06-25
chore: Consolidate `unimport` code into a built-in module

aklinker1:
- Moves all the unimport code (except resolving config) into a built-in module, consolidating things.
- This PR also fixes an issue where default config was ignored because we were passing options into `scanImportsFromDir`, which was unnecessary (6bd2d16913b92ea106b415a763894949695eb81f)


packages/wxt/src/core/utils/building/generate-wxt-dir.ts
```diff
@@ -236,20 +182,19 @@ async function getGlobalsDeclarationEntry(): Promise<WxtDirFileEntry> {
...
-      return `/// <reference types="./${normalizePath(relativePath)}" />`;
-    }),
-  ].join('\n');
+      lines.push(`/// <reference types="./${normalizePath(relativePath)}" />`);
+    }
+  });
   return {
     path: 'wxt.d.ts',
-    text,
+    text: lines.join('\n') + '\n',
```
aklinker1, line 197: There were blank lines being generated when a file didn't have a reference, this fixes that.

PR #774, opened 2024-06-26
fix: Respect custom `outDir` when cleaning and zipping

AsakuraMizu:
1. `wxt clean` now removes output directory contents based on the config instead of `.output`, and also improvement on the command (config, logger, etc.); also a question here why we only remove its sub dirs but keep generated zips?
2. `wxt zip` now exclude output directory for firefox sources by default since user may specify a `outDir` that does not start with a dot and it's annoying to specify it again in `excludeSources` manually

### DIFF
--- a/packages/wxt/src/core/clean.ts
+++ b/packages/wxt/src/core/clean.ts
@@ -13,7 +13,24 @@ import { registerWxt, wxt } from './wxt';
  * @example
  * await clean();
  */
-export async function clean(config?: InlineConfig) {
+export async function clean(config?: InlineConfig): Promise<void>;
+/**
+ * Remove generated/temp files from the directory.
+ *
+ * @deprecated
+ *
+ * @param root The directory to look for generated/temp files in. Defaults to `process.cwd()`. Can be relative to `process.cwd()` or absolute.
+ *
+ * @example
+ * await clean();
+ */
+export async function clean(root?: string): Promise<void>;
+
+export async function clean(config?: string | InlineConfig) {
+  if (typeof config === 'string') {
+    config = { root: config };
+  }
+
   await registerWxt('build', config);
   wxt.logger.info('Cleaning Project');
 
--- a/packages/wxt/src/core/clean.ts
+++ b/packages/wxt/src/core/clean.ts
@@ -1,44 +1,48 @@
 import path from 'node:path';
 import glob from 'fast-glob';
 import fs from 'fs-extra';
-import { consola } from 'consola';
 import pc from 'picocolors';
+import { InlineConfig } from '~/types';
+import { registerWxt, wxt } from './wxt';
 
 /**
  * Remove generated/temp files from the directory.
  *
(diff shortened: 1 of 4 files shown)


packages/wxt/src/core/utils/building/resolve-config.ts
```diff
@@ -253,6 +253,8 @@ function resolveZipConfig(
       // Tests
       '**/__tests__/**',
       '**/*.+(test|spec).?(c|m)+(j|t)s?(x)',
+      // Output directory
+      `${mergedConfig.outDir ?? '.output'}/**`,
```
aklinker1, line 257: Pass the `outBaseDir` into this function like the `root` instead of recomputing it here.

packages/wxt/src/core/clean.ts
```diff
@@ -1,44 +1,48 @@
...
  *
- * @param root The directory to look for generated/temp files in. Defaults to `process.cwd()`. Can be relative to `process.cwd()` or absolute.
+ * @param config Optional config that will override your `<root>/wxt.config.ts`.
  *
  * @example
  * await clean();
  */
-export async function clean(root = process.cwd()) {
-  consola.info('Cleaning Project');
+export async function clean(config?: InlineConfig) {
```
aklinker1, line 16: This is a breaking change to the public API, but I agree it's probably better. Could you add an overload and mark the old implementation as deprecated?

I have one more breaking change to make for 0.19, I'll remove the deprecated overload when I'm preparing that release.

packages/wxt/src/core/clean.ts
```diff
@@ -1,44 +1,48 @@
...
 
   const tempDirs = [
     'node_modules/.vite',
     'node_modules/.cache',
     '**/.wxt',
-    '.output/*',
+    `${path.relative(root, wxt.config.outBaseDir)}/*`,
   ];
-  consola.debug('Looking for:', tempDirs.map(pc.cyan).join(', '));
+  wxt.logger.debug('Looking for:', tempDirs.map(pc.cyan).join(', '));
```
aklinker1, line 28: Woops, there's a couple places outside this file that need to use `wxt.logger` as well. I'll do that in a separate PR. Good find

aklinker1 (changes requested):
Most of this looks good, thanks for the PR! Just a couple of small changes.

> why we only remove its sub dirs but keep generated zips?

In my mind, clean commands are meant to reset the build environment and force a fresh build. Probably from when I was a mobile developer and I had to use `gradle clean` all the time.

Since ZIPs don't effect the TypeScript project, they don't need to be deleted. But I could be convinced otherwise... But it would have to be considered another breaking change at this point.

aklinker1 (approved):
Alright, looks good. Thanks for the PR!

aklinker1:
Released in [`v0.18.11`](https://github.com/wxt-dev/wxt/releases/tag/wxt-v0.18.11)

PR #793, opened 2024-07-02
perf: create zip using streams

Timeraa:
I noticed that zipping is not done using streams which results in the zip file being accumulated in the memory of the user which massively slows down the process of zipping and for some users (me included) throws an out of heap error for big projects.

before:
image
after:
image


aklinker1 (approved):
Nice, tests still pass, great find!

aklinker1:
Released in [`v0.18.12`](https://github.com/wxt-dev/wxt/releases/tag/wxt-v0.18.12)

PR #797, opened 2024-07-03
docs: rephrase a paragraph

sacrosanctic:
I found the previous wording clunky, so I tried to rephrase it while keeping the original voice.


aklinker1 (approved):
Thank you! I agree, much better.

PR #799, opened 2024-07-06
doc: detail issues with vite plugins

### LINKED ISSUE #798: Document issues with vite plugins
https://discord.com/channels/1212416027611365476/1258851568241868863/1258934891115057222

> Maybe something like, "due to the way wxt is implemented, some vite plugins do not work out of the box, and requires..."
> 
> And maybe point then to where they can get support like discord or GitHub issue?

sacrosanctic:
closes #798 

I removed the specifics, but I'm unsure if you want to go into more detail on why plugins wouldn't work as expected.

We should make searching easier by adding a `vite-plugin` tag or standardizing issue titles like "[VITE-PLUGIN] {name of plugin}"

### DIFF
--- a/docs/guide/go-further/vite.md
+++ b/docs/guide/go-further/vite.md
@@ -32,10 +32,6 @@ export default defineConfig({
 });
 ```
 
-:::info
-There are caveats with vite plugins due to the way WXT implement vite builds:
-- plugins may not work out of the box
-- plugins may be executed multiple times
-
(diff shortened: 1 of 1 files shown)


aklinker1 (approved):
Thanks! I added a label and updated the titles of plugin issues to make it easier to see what plugin was being discussed.

https://github.com/wxt-dev/wxt/issues?q=is%3Aissue+label%3A%22vite+plugin%22

Issue #809, opened 2024-07-07
Add XPath support for anchor property


sir-kokabi:
### Feature Request

I propose adding support for XPath selectors in the [anchor](https://wxt.dev/api/reference/wxt/client/interfaces/ContentScriptAnchoredOptions.html#anchor). 

Currently, the anchor property supports CSS selector, element, or function returning these. Adding XPath support would provide more flexibility in targeting elements, especially in complex DOM structures where XPath might be more suitable or convenient.

### Proposed type changes for the anchor property:

N/A

### Is your feature request related to a bug?

N/A

### What are the alternatives?

1. Continue using CSS selectors exclusively, potentially with more complex selectors to achieve similar results.
2. Implement a custom function that uses [document.evaluate()](https://developer.mozilla.org/en-US/docs/Web/API/Document/evaluate) to find elements using XPath, and return the result as the anchor.

```ts
function getElementByXPath = (xpath, parent = document) => {
  const result = document.evaluate(xpath, parent, null, XPathResult.FIRST_ORDERED_NODE_TYPE, null);
  return result.singleNodeValue;
};
```
However, built-in XPath support would provide a more straightforward and maintainable solution for users who prefer or require XPath selectors. This also helps migrate a project that uses a lot of XPath selectors to WXT more easily.

### Additional context

This feature can be a competitive advantage over Plasmo, which selects anchors based on CSS selectors. XPath, especially useful when the DOM structure is dynamically generated or faced with nested and complex structures, provides more powerful selection capabilities compared to CSS selectors in certain cases such as selecting elements based on textual content or relative positions.

**Let me add this feature to the codebase.**

aklinker1:
Could you share an example of using XPath vs CSS and the benefits of it? I'm not familiar with XPath tbh.

So you're suggesting to check if the anchor starts with `"//"` an then use `document.evaluate` (as shown in your example) instead of `document.querySelector`?

sir-kokabi:
For example:
`//div[@data-testid='sidebarColumn']//div[@aria-label='Timeline: Trending now']/ancestor::*[3]`

You can find more examples in my project [Twifiner](https://github.com/sir-kokabi/twifiner/blob/main/src/contents/index.ts). I developed it using Plasmo. XPath has significantly shortened long code snippets and made it possible to achieve many tasks in just a few lines, tasks that were difficult with CSS selectors and previously required multiple lines of code.

Yes, a relative XPath begins with `//` and an absolute XPath begins with `/`.

[...]

PR #813, opened 2024-07-08
feat: add XPath support to getAnchor()

sir-kokabi:
#809

### DIFF
--- a/packages/wxt/src/client/content-scripts/ui/index.ts
+++ b/packages/wxt/src/client/content-scripts/ui/index.ts
@@ -202,7 +202,7 @@ function getAnchor(options: ContentScriptAnchoredOptions): Element | undefined {
 
   if (typeof resolved === 'string') {
     // If the string is an XPath expression (starts with '//' or '/')
-    if (resolved.startsWith('//') || resolved.startsWith('/')) {
+    if (resolved.startsWith('/')) {
       // Evaluate the XPath and return the first ordered node
       const result = document.evaluate(
         resolved,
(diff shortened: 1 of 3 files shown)


packages/wxt/src/client/content-scripts/ui/index.ts
```diff
@@ -199,8 +199,25 @@ function getAnchor(options: ContentScriptAnchoredOptions): Element | undefined {
 
   let resolved =
     typeof options.anchor === 'function' ? options.anchor() : options.anchor;
-  if (typeof resolved === 'string')
-    return document.querySelector<Element>(resolved) ?? undefined;
+
+  if (typeof resolved === 'string') {
+    // If the string is an XPath expression (starts with '//' or '/')
+    if (resolved.startsWith('//') || resolved.startsWith('/')) {
```
aklinker1, line 205: can this be simplified?

```suggestion
    if (resolved.startsWith('/')) {
```

packages/wxt/src/client/content-scripts/ui/types.ts
```diff
@@ -194,7 +194,7 @@ export type ContentScriptPositioningOptions =
 
 export interface ContentScriptAnchoredOptions {
   /**
-   * A CSS selector, element, or function that returns one of the two. Along with `append`, the
+   * A CSS selector, XPath expression, element, or function that returns one of the two. Along with `append`, the
```
aklinker1, line 197:
```suggestion
   * A CSS selector, XPath expression, element, or function that returns one of the three. Along with `append`, the
```

packages/wxt/src/client/content-scripts/ui/index.ts
sir-kokabi: Yes, you are correct. It has been resolved.

packages/wxt/src/client/content-scripts/ui/types.ts
sir-kokabi: This one too

aklinker1 (approved):

PR #826, opened 2024-07-17
docs: fix link to unimport

eetann:
The link to unimport was vue.ts, so corrected it to svelte.ts.


aklinker1 (approved):
Thanks! I double checked the rest of the modules, this was the only one with the copy/paste error.

PR #758, opened 2024-06-19
feat: New `@wxt-dev/i18n` package

aklinker1:
### Usage

```yml
# <srcDir>/locales/en.yml
simpleMessages: No need to use an object and provide a message or description field
chromeMessages:
  message: But you still can
  description: Same as the regular i18n setup, this field shouldn't be translated.
nested:
  messages: You can access nested messages using a "."
substitutions: You can include substitutions like $1
plurals:
  0: 0 items
  1: 1 item
  n: $1 items
```

> You can use `.json`, `.jsonc`, `.json5`, `.yaml`, `.yml`, or `.toml` files for your messages

```ts
// Auto-imports from:
// import { i18n } from 'wxt/client';

i18n.t("simpleMessage")              // No need to use an object and provide a message or description field
i18n.t("chromeMessages")             // But you still can
i18n.t("nested.messages")            // You can access nested messages using a "."
i18n.t("substitutions", ["this!"])   // You can include substitutions like this!
i18n.t("plurals", 0)                 // 0 items
i18n.t("plurals", 1)                 // 1 item
i18n.t("plurals", 2)                 // 2 items
i18n.t("plurals", 2, ["Two"])        // Two items
```

### Overview

This closes #327, newer version of PR #343. It's out of date, and now that the repo is a monorepo, it can be it's own package.

- [x] Untyped client (`i18n.t(...)`)
   - [x] Plan for future pluralization resolvers (like arabic's few/many plural forms)
      ```diff
      -createI18n()
      +createI18n({ resolvePlurals: (forms, count, locale) => ... })
      ```
   - [x] Plan for future substitution formatters (like money or numbers)
      ```diff
      -i18n.t("it-costs", [dollars])
      +i18n.t("it-costs", [i18n.money(dollars)])
      ```
- [x] Allow customizing types (`createI18n<MyI18n>(...)`)
- [x] Detect number of substitutions and make it type-safe
- [x] Generate declaration file for messages file
- [x] Allow passing options into `t(..., options)`
- [x] Utils for transforming custom JSON/JSON5/YAML to standard chrome extension json
   - [x] Generate plurals in correct format
   - [x] Support array keys
   - [x] Use dots for nested objects/keys
   - ~~Add `@see` links in JSDoc to link to original messages file?~~ Not something all editors support, so skipping
- [x] WXT Module
   - [x] Generate and integrate types
   - [x] Throw error if `default_locale` is missing
   - [x] Regenerate and reload extension when saving localization file
[...]

PR #832, opened 2024-07-21
fix: Don't throw error if localization file is missing

aklinker1:
Before, if you listed a `default_locale` in your manifest, WXT would throw an error if you didn't provide a `public/_locales/{default_locale}/messages.json` file. This PR removes this requirement.

Instead of having WXT do this validation, we'll just let the browser report the problem when the extension is installed instead. This is because  it could be perfectly valid to not have a file there but because a module or some other custom build step adds translations to the output directory.

### DIFF
--- a/packages/wxt/src/core/utils/building/generate-wxt-dir.ts
+++ b/packages/wxt/src/core/utils/building/generate-wxt-dir.ts
@@ -127,14 +127,14 @@ declare module "wxt/browser" {
 }
 `;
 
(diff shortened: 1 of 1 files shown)


aklinker1:
CC @Timeraa - broke this out of #758 for you.

aklinker1:
Released in [`v0.18.15`](https://github.com/wxt-dev/wxt/releases/tag/wxt-v0.18.15)

PR #827, opened 2024-07-18
feat(storage): `init` option and rename `defaultValue` to `fallback`

aklinker1:
This closes https://github.com/wxt-dev/wxt/issues/823

PR #846, opened 2024-07-23
feat(storage): initialize item with default value

veryCrunchy:


aklinker1:
@veryCrunchy Why do you need to re-initialize a value? What's your use-case?

veryCrunchy:
@aklinker1 Since I seem to have trouble explaining things in a way that's clear and easy to understand, I've let ai sum it up for me;

Your explanation captures the essence of `reInit` quite well. Here’s a refined version that maintains clarity and emphasizes the key benefits:

---

The `reInit` method simplifies the process of resetting and synchronizing stored values by leveraging the initialization logic defined within the item itself. Here's why this is beneficial:

1. **Streamlined State Management**:
   - By defining the initialization logic directly within the item, you can easily manage state changes. For instance, when a user logs in with a new account, you can call `reInit` to update the stored user ID without manually handling the reset logic elsewhere.

2. **Seamless Synchronization**:
   - If your stored data needs to align with external sources (such as a database or API), `reInit` ensures that the stored value is updated according to the latest external data. This avoids discrepancies and keeps your application state in sync with external changes.

3. **Encapsulation of Initialization Logic**:
   - Instead of having separate functions for resetting or initializing values, the logic is encapsulated within the item definition. This simplifies your codebase by reducing the need for additional imports and separate reset functions.

### Example Use-Cases

- **User Session**: Resetting user-related data when a new account is logged in.
- **Feature Flags**: Updating feature flags based on configuration changes or external settings.

By using `reInit`, you streamline state management and ensure that your stored values remain consistent with their intended initialization logic.

---

This explanation underscores the convenience and effectiveness of `reInit` in managing and synchronizing data.

aklinker1:
> Example Use-Cases
> - User Session: Resetting user-related data when a new account is logged in.
> - Feature Flags: Updating feature flags based on configuration changes or external settings.

This is what I was looking for, thanks 😄

Neither of these cases make sense to me. Both are easily solved by just calling `setValue` instead of putting things like network requests inside the `init` option...

From an code organization perspective, think of the storage items as your persistence layer, not the service layer. You don't put business logic in repos, you put business logic in services.

For example, for user session account management, this is how I would implement it at a high level:

```ts
const user = storage.defineItem<User>('local:user');
const preferences = storage.defineItem<Preferences>('local:preferences')

async function login(username, password) {
  const newUser = await fetchUser(username, password)
  const newPreferences = await fetch preferences(newUser);

  await user.setValue(newUser)
  await preferences.setValue(newPreferences)
}
```

Or feature flags fetched from an API:

```ts
const featureFlags = storage.defineItem('local:feature-flags', {
  fallback: {...},
})

scheduleJob('every 30m', async () => {
  const latestFlags = await fetchFeatureFlags();
  await featureFlags.setValue(latestFlags);
})
```

In both these cases, set value works just fine, and the code is organized with all related logic together. I don't see how `reInit` would simplify this code or make it more organized.

aklinker1:
One potential use-case I see is analytics session ids. Having a way to quickly regenerate a new ID at the beginning of a session would be useful. But in that case, I'm either using the `browser.storage.session` or `sessionStorage`; the first, `init` would work as-is and the second isn't extension storage.

aklinker1:
@veryCrunchy I already have a PR open for this change, minus `reInit`. I would hold off on doing more work until you convince me the feature is necessary. https://github.com/wxt-dev/wxt/pull/827

aklinker1:
@veryCrunchy I merged [my PR](https://github.com/wxt-dev/wxt/pull/827), putting together a big release and wanted to include that one. Feel free to keep this open and add `reInit` if you feel strongly about it :)

Issue #784, opened 2024-06-30
Don't use `webextension-polyfill` by default


aklinker1:
### Feature Request

Currently, WXT's `browser` is the default export from `webextension-polyfill`. However, now that Chrome doesn't support MV2 extensions, and Firefox has MV3 support, it's not really doing anything anymore... It's main features were:

1. Add promise-based API to chrome's MV2 implementation
2. Providing a single global variable so you don't have to use both `chrome` and `browser`

However, those two features are no longer necessary:

1. Add promise-based API to chrome's MV2 implementation
   - Firefox and other browsers all support MV3 now, and Firefox's MV2 APIs are already promise-based. So you can write promise-based API calls and it will work on all browsers
2. Providing a single global variable so you don't have to use both `chrome` and `browser`
   - All browsers now support the `chrome` global, but WXT will continue to re-export it's own `browser` variable.

So how is WXT going to remove the polyfill?

First, it will provide some flag in config to continue using it if you want, but the polyfilll will not be the default. It will be moved to an optional peer-dependency so it's not even installed if it doesn't have to be.

Second, we'll be replacing `@types/webextension-polyfill` with `@types/chrome`. This will address lots of the problems with missing API and manifest types, at the cost of not being able to use type errors to know what's supported by all browsers and what's not. Either way, both ways have problems, but I think `@types/chrome` will provide a better DX.

Finally, we'll need to update `wxt/testing` to make sure it properly fakes the API. Not sure if `@webext-core/fake-browser` is still the best option for providing a fake version during development. _Edit: Been looking and can't find anything else on NPM_

So in summary, you shouldn't have to make any changes to your source code or tests. That's the goal, to make this transition as seemless as possible. You'll be able to opt into using the polyfill from `wxt.config.ts`. There will be breaking changes to types, don't think I can get around that.

If you use a library that includes the polyfill, it will work just fine, the polyfill will still be included.

#### Is your feature request related to a bug?

https://github.com/wxt-dev/wxt/issues/506 - The `vite-node` solution works well, but the polyfill is getting in the way.

https://github.com/wxt-dev/wxt/issues/521 - Improve manifest typing

https://github.com/wxt-dev/wxt/issues/299 - Improve runtime API typing

### What are the alternatives?

N/A

### Additional context

- Discord: https://discord.com/channels/1212416027611365476/1253736416425410560

aiktb:
Does the default build target for `wxt build -b firefox` change to MV3?

[...]

PR #847, opened 2024-07-23
feat(experimental): First-class support for excluding `webextension-polyfill`

aklinker1:
> [!WARNING]
> If you were previously using `disableWebextensionPolyfill`, here's how to migrate to the new setting:
>
> ```diff
> experimental: {
> - disableWebextensionPolyfill: true,
> + extensionApi: "chrome",
> }
> ```

Apart of #784, this PR refactors the experimental flag for removing the `webextension-polyfill` from your bundled extension.

Before, it was a blanket replacement - even if a NPM package depended on the polyfill, the polyfill would be stripped out of the bundle, often times breaking the package's functionality.

Now, WXT will use `wxt/browser/chrome` instead of `wxt/browser`. This new module just exports the `chrome` or `browser` global based on the runtime browser. If a NPM package imports the polyfill, the polyfill will be included and the NPM package will function normally!

To enable this new functionality, add the following option to your `wxt.config.ts`:

```ts
export default defineConfig({
  extensionApi: "chrome",
});
```

Note that this is an experimental feature, and I need help testing it out!


packages/wxt/src/__tests__/storage.test.ts
```diff
@@ -1,6 +1,6 @@
 import { fakeBrowser } from '@webext-core/fake-browser';
 import { describe, it, expect, beforeEach, vi, expectTypeOf } from 'vitest';
-import { browser } from '../browser';
+import { browser } from 'wxt/browser';
```
aklinker1, line 3: Because we're using `mkdist` now, we can import from `wxt/browser` and it will be resolved from the NPM module instead of using the local file, and this will be replaced with `wxt/browser/chrome` when using `extensionApi: "chrome"`

packages/wxt/src/core/builders/vite/plugins/resolveExtensionApi.ts
```diff
@@ -0,0 +1,27 @@
+import { ResolvedConfig } from '../../../../types';
+import type * as vite from 'vite';
+
+/**
+ * Apply the experimental config for which extension API is used. This only
+ * effects the extension API included at RUNTIME - during development, types
+ * depend on the import.
+ *
+ * NOTE: this only works if we import `wxt/browser` instead of using the relative path.
```
aklinker1, line 9: TODO: Add eslint rule to ensure this works.

aklinker1:
Tested new mode here: https://github.com/aklinker1/github-better-line-counts/pull/34

Working like a charm!

PR #848, opened 2024-07-23
perf(size)!: Switch from `tsup` to `unbuild` for building WXT

aklinker1:
BREAKING CHANGE: Drop CJS support

---

I was having some weird behavior with dev mode module resolution in #847, and wanted to simplify the build output to see if that would have any effect. Haven't tested it yet, but as you can see below, it's a good change anyways:

| | TSUP | Unbuild |
| --- | --- | --- |
| Packed size | 431 kB | 114 kB |
| Unpacked size | 1.89 mb | 656 kB |
| Build time | 6.419 s | 3.002 s |

Other changes:

- Removed `~` typscript path, doesn't work with `mkdist`
- Dropped support for CJS


packages/wxt/src/core/utils/building/resolve-config.ts
```diff
@@ -374,12 +374,11 @@ async function getUnimportEslintOptions(
...
-  const requireResolve =
-    require?.resolve ??
-    (await import('node:module')).default.createRequire(import.meta.url)
-      .resolve;
+  const require = (await import('node:module')).default.createRequire(
+    import.meta.url,
+  );
   // require.resolve returns the wxt/dist/index file, not the package's root directory, which we want to return
-  return path.resolve(requireResolve('wxt'), '../..');
+  return path.resolve(require.resolve('wxt'), '../..');
```
aklinker1, line 381: Figure... this out

packages/wxt/scripts/build.ts
```diff
@@ -1,126 +0,0 @@
...
-    ...preset,
-    entry: {
-      cli: 'src/cli/index.ts',
-    },
-    format: ['esm'],
-    banner: {
-      // Fixes dynamic require of nodejs modules. See https://github.com/wxt-dev/wxt/issues/355
-      // https://github.com/evanw/esbuild/issues/1921#issuecomment-1152991694
-      js: "import { createRequire } from 'module';const require = createRequire(import.meta.url);",
-    },
```
aklinker1, line 104: Another point of interest

aklinker1: We don't seem to need this anymore.

PR #849, opened 2024-07-23
fix: wrong module hook type

Timeraa:
I noticed that if you define the hooks object the type requires you to set all hooks which isn't correct


aklinker1 (approved):
Oh shoot, woops lol. Thanks

PR #851, opened 2024-07-24
feat: Auto Icons Module

### LINKED ISSUE #850: Automatically generate logos in correct sizes
### Feature Request

Was thinking it would be a nice feature if wxt can take a path to the logo and automatically generated images in the correct sizes for all the stores (512x512, 128x128 etc.) 

Just a small discussion point because I would definitely appreciate it but not sure how others / how much value it would add to the project

Timeraa:
Adds a new module to wxt to automatically generate icons for the manifest `icons` field using sharp

Closes #850

For the structure (package.json, tsconfig) I copied the one from the vue module, had to add a build config for fs-extra


Timeraa:
@aklinker1 I noticed a small issue with the wxt module system, is there a way to get the module options in hooks? Because right now you would need to restart the whole process in order for the options to update.

aklinker1:
Are you saying if someone saves the `wxt.config.ts` file with new options, they aren't updated? Hmmm... Yeah I can see why that would be the case. Let me think on the best way to fix that.

Timeraa:
> Are you saying if someone saves the `wxt.config.ts` file with new options, they aren't updated? Hmmm... Yeah I can see why that would be the case. Let me think on the best way to fix that.

Yep exactly that! It totally makes sense why but probably something to think about!

packages/module-auto-icons/package.json
```diff
@@ -0,0 +1,42 @@
...
+  "exports": {
+    ".": {
+      "import": {
+        "types": "./dist/index.d.mts",
+        "default": "./dist/index.mjs"
+      },
+      "require": {
+        "types": "./dist/index.d.cts",
+        "default": "./dist/index.cjs"
+      }
```
aklinker1, line 17: v0.19.0 is queued up and about to be released, the breaking change being dropping CJS support. So you can remove the "require". I should remove it from the other modules as well...

```suggestion
      "types": "./dist/index.d.mts",
      "default": "./dist/index.mjs"
```

packages/module-auto-icons/build.config.ts
```diff
@@ -0,0 +1,5 @@
+import { defineBuildConfig } from 'unbuild';
+
+export default defineBuildConfig({
+  externals: ['fs-extra'],
```
aklinker1, line 4: Isn't unbuild already marking `fs-extra` as external? It should see it in the `dependencies` and know to not bundle it.

aklinker1 (changes requested):
Code looks good, just a couple of small changes.

What are your thoughts on the `module-` prefix for the package names? I wanted it to be consistent, but now I'm regretting it... `@wxt-dev/auto-icons` is good too 😵‍💫

packages/module-auto-icons/build.config.ts
aklinker1: Yeah, tested and it's the same output without this file, I removed it in my commit.

packages/module-auto-icons/package.json
aklinker1: Added in my commit

aklinker1:
Looks like the generated icons aren't added to the internal output structure, and are causing some issues: 

Screenshot 2024-07-25 at 10 10 22 AM

And the icons don't show up in the build output.

Screenshot 2024-07-25 at 10 11 42 AM

Might need to add some more hooks to WXT to fix these issues.

aklinker1:
For example, a `prepare:publicPaths` hook here to fix the `browser.runtime.getURL` issue: https://github.com/wxt-dev/wxt/blob/ab234c0f458b4fb62de353130d6971701576e946/packages/wxt/src/core/utils/building/generate-wxt-dir.ts#L78-L90

For the build output issue... I think we can just add the icon files to the `output.publicAssets` file in the `build:done` hook.

```
wxt.hooks.hook("build:done", (wxt, output) => {
  output.publicAssets.push({ type: "asset", file: "icons/16.png");
});
```

Timeraa:
Oh gosh what did I do to the branch

Timeraa:
Implemented the requested changes @aklinker1, I also updated the background.ts so that the /icons path is correct in the geturl

aklinker1 (approved):
> Implemented the requested changes @aklinker1, I also updated the background.ts so that the /icons path is correct in the geturl

Perfect, thanks. Everything looks good now.

> I can make it just auto-icons if ya want, just did fit for consistency haha

Hmmmmmmmmmmmmmmmmmm... Yeah, let's rename it to just `@wxt-dev/auto-icon`

PR #880, opened 2024-07-29
fix(auto-icons): path option

Timeraa:


Timeraa:
@aaronklinker-st I noticed an issue, the overwriting of the manifest does not work

Timeraa:
@aklinker1  I noticed an issue, the overwriting of the manifest does not work

packages/auto-icons/src/index.ts
```diff
@@ -11,18 +11,19 @@ export default defineWxtModule<AutoIconsOptions>({
   async setup(wxt, options) {
     const parsedOptions = defu(options, {
```
aklinker1, line 12: You can defu type safe with:

```suggestion
    const parsedOptions = defu<AutoIconOptioms, AutoIconOptioms>(options, {
```

packages/auto-icons/src/index.ts
```diff
@@ -11,18 +11,19 @@ export default defineWxtModule<AutoIconsOptions>({
...
       enabled: true,
-      baseIconsPath: resolve(wxt.config.srcDir, 'assets/icon.png'),
+      baseIconPath: resolve(wxt.config.srcDir, 'assets/icon.png'),
       grayscaleOnDevelopment: true,
       sizes: [128, 48, 32, 16],
     });
 
+    const resolvedPath = resolve(process.cwd(), parsedOptions.baseIconPath);
```
aklinker1, line 19: If the option defaults to resolving from the src dir, should we also resolve from there instead of cwd?

Maybe also update the options jsdoc to specify where the path is relative to?

Timeraa: Changed it to resolve from srcDir and added a comment in the options.

packages/auto-icons/src/index.ts
Timeraa: Says to prevent further typos suggests a fix with typos 😂

aklinker1 (approved):
Sorry about the typo 😬

packages/auto-icons/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@wxt-dev/auto-icons",
-  "version": "1.0.0",
+  "version": "1.0.1",
```
aklinker1, line 3: Oh wait, don't manually bump the version, this is handled in the release workflow.

aklinker1: You're good! I should have caught this in review. When a project uses conventional commits, and often times with other release methods, there's usually a specific commit for bumping the version, so there's a nice clean change to apply a tag to. With conventional commits, release tools do the bumping automatically for you.

PR #885, opened 2024-07-31
chore: add more metadata for npm

Timeraa:
This PR adds more metadata to make the modules more complete in terms of "required" data that's shown on npm, perhaps we should add more metadata @aklinker1?

### DIFF
--- a/packages/auto-icons/package.json
+++ b/packages/auto-icons/package.json
@@ -1,5 +1,28 @@
 {
   "name": "@wxt-dev/auto-icons",
+  "description": "WXT module for automatically generating extension icons in different sizes",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/wxt-dev/wxt/tree/main/packages/auto-icons"
+  },
+  "homepage": "https://wxt.dev",
+  "keywords": [
+    "wxt",
+    "module",
+    "icons",
+    "manifest"
+  ],
+  "author": {
+    "name": "Florian Metz",
+    "email": "me@timeraa.dev"
+  },
+  "contributors": [
+    {
+      "name": "Aaron Klinker",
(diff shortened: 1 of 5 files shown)


packages/auto-icons/package.json
```diff
@@ -1,5 +1,28 @@
 {
   "name": "@wxt-dev/auto-icons",
+  "description": "WXT module for automatically generating extension icons in different sizes",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/wxt-dev/wxt/tree/main/packages/auto-icons"
+  },
```
aklinker1, line 7: This doesn't look right. See Vite's for reference. I think `packages/wxt` is also out of date.

https://github.com/vitejs/vite/blob/0f56e1724162df76fffd5508148db118767ebe32/packages/vite/package.json#L63-L67

aklinker1 (commented):
Rest looks good. Thanks for add this. 

_PS: I won't touch this code so you get full credit lol_

packages/module-react/package.json
```diff
@@ -1,5 +1,21 @@
 {
   "name": "@wxt-dev/module-react",
+  "description": "WXT module to enable React support",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/wxt-dev/wxt/tree/main/packages/module-react"
+  },
+  "homepage": "https://wxt.dev",
```
aklinker1, line 8: Thoughts on having the modules point to their README instead of the main docs website? There isn't any docs specific to these modules on it.

https://github.com/wxt-dev/wxt/tree/main/packages/module-react/README.md

Timeraa: Think that's better yeah

aklinker1 (approved):

PR #886, opened 2024-08-01
docs: Improve `prepare:types` hook JSDoc

mehimanshupatil:


aklinker1 (approved):
Thanks, that info is definitely important and was definitely missing!

Issue #738, opened 2024-06-16
`wxt zip` including `.git/` folder in `sources` zip


uncenter:
### Describe the bug

Turns out https://github.com/aklinker1/publish-browser-extension/issues/21 is actually due to the `sources` zip being 30x larger (6mb to 190mb) after upgrading from `0.17.3` to `0.18.x`, and Mozilla refusing to accept such a wildly large file. I believe this is due to https://github.com/wxt-dev/wxt/pull/674 somehow leading to my `.git/` folder being included in the `sources` zip, confirmed after downloading it from the [release assets of my project](https://github.com/catppuccin/github-file-explorer-icons/releases/tag/v0.6.3) and listing the contents of it:

![CleanShot 2024-06-16 at 14 54 43@2x](https://github.com/wxt-dev/wxt/assets/47499684/cb774318-a53d-4341-ad5b-4cf3120a42fe)

Also either that, zipping so many `.git/` files, or https://github.com/wxt-dev/wxt/pull/501, is causing `FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory` when I try to run `wxt zip` on `0.18.x`.

### Reproduction

Not minimal unfortunately but easily demonstrates the issue by changing the version of `wxt` installed: https://github.com/catppuccin/github-file-explorer-icons.

### Steps to reproduce

1. `git clone --recurse-submodules https://github.com/catppuccin/github-file-explorer-icons.git`
2. `pnpm install` in the project directory.
3. `pnpm zip:firefox`, note that WXT will list the size of the sources zip in the terms of 100s of MB.
4. `pnpm add -D wxt@0.17.3`
5. Delete the `dist/` directory for good measure (I think this bug was even including its own output zips in later zips?).
6. `pnpm zip:firefox`, note that WXT will list the size of the sources zip as 5-6 MB.

### System Info

```shell
System:
    OS: macOS 15.0
    CPU: (8) arm64 Apple M1
    Memory: 85.48 MB / 8.00 GB
    Shell: 3.7.1 - /run/current-system/sw/bin/fish
  Binaries:
    Node: 20.14.0 - ~/.local/state/fnm_multishells/10829_1718488770533/bin/node
    npm: 10.7.0 - ~/.local/state/fnm_multishells/10829_1718488770533/bin/npm
    pnpm: 9.3.0 - ~/.local/state/fnm_multishells/5285_1718466025130/bin/pnpm
[...]

PR #902, opened 2024-08-04
feat: allow zipping hidden files by listing them explicitly in `includeSources`

hikiko4ern:
I omitted `dot: true` and didn't touch the default `excludeSources` (mentioned in https://github.com/wxt-dev/wxt/issues/738#issuecomment-2192251866) to avoid filtering all files inside hidden directories (including those in `.git`). For example, in my repo:
- `dot: true` filters 4657 files
- only 1028 files are filtered with this PR

In theory, there should be no difference for the end user - both approaches will not include files inside the hidden directory if only that directory is specified in `includeSources` (i.e. to include all files inside `.hidden-dir`, you should specify `.hidden-dir/**`, not `.hidden-dir`). Perhaps this specific aspect should be mentioned somehow in the `includeSources` JSDoc?

### DIFF
--- a/packages/wxt/e2e/tests/zip.test.ts
+++ b/packages/wxt/e2e/tests/zip.test.ts
@@ -97,8 +97,7 @@ describe('Zipping', () => {
     expect(await project.fileExists(sourcesZip)).toBe(true);
   });
 
-  // TODO: Fix #738 and re-enable
-  it.skip('should not zip hidden files into sources by default', async () => {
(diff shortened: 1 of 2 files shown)


aklinker1 (approved):
Well that was an easy fix lol. Thanks for updating the tests too!

aklinker1:
Released in [`v0.19.5`](https://github.com/wxt-dev/wxt/releases/tag/wxt-v0.19.5)

PR #906, opened 2024-08-06
feat: allow excluding files when zipping

Timeraa:
Allows the user to exclude files when zipping the folder, real world use case for this would be excluding sourcemaps for example when using the sentry plugin you need to have them but when zipping the actual extension I don't want to have them to not "expose" the structure of my codebase.

I didn't add the include option since you can do that using the public folder

I am not quite sure about the JSDoc, had to escape it like this, anyone know a proper fix for that?

### DIFF
--- a/packages/wxt/src/core/utils/building/resolve-config.ts
+++ b/packages/wxt/src/core/utils/building/resolve-config.ts
@@ -268,6 +268,7 @@ function resolveZipConfig(
     includeSources: [],
     compressionLevel: 9,
     ...mergedConfig.zip,
+    exclude: mergedConfig.zip?.exclude ?? [],
     excludeSources: [
       '**/node_modules',
       // WXT files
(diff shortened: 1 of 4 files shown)


aklinker1 (commented):
JSDoc looks fine. When you add an option, you also need to add it to `resolve-config.ts`: https://github.com/wxt-dev/wxt/blob/36cfcd0acb57a1958f605ff14f2310cae3645ae7/packages/wxt/src/core/utils/building/resolve-config.ts#L36

See the failing checks, just fix the type errors and this will be good to go.

Timeraa:
> JSDoc looks fine. When you add an option, you also need to add it to `resolve-config.ts`: https://github.com/wxt-dev/wxt/blob/36cfcd0acb57a1958f605ff14f2310cae3645ae7/packages/wxt/src/core/utils/building/resolve-config.ts#L36
> 
> 
> 
> See the failing checks, just fix the type errors and this will be good to go.

Oh I actually did that but I think I removed that by accident before pushing, my bad

aklinker1 (approved):

aklinker1:
Released in [`v0.19.5`](https://github.com/wxt-dev/wxt/releases/tag/wxt-v0.19.5)

PR #908, opened 2024-08-09
fix: #907 move wxt devDeps execa to dependencies

1natsu172:
fix #907 

I was looking about #907 and found that the error was caused by dynamic import `execa`.

In non-monorepo, I think it was by chance that `execa@8` was found and working so far (What a surprise, up until now, execa was running on `@8`!). But in monorepo, there are cases where `execa@5` is found before `execa@8`, due to pnpm hoist and module resolution.

> Until 0.19.4: not found the expect `execa@9.3.0`
image

Here is a screenshots of the patch in `node_modules/wxt` and logging of `import.meta.resolve` etc.

| Monorepo#wxt@0.19.4 | Standalone#wxt@0.19.4 |
|--------|--------|
|  found `execa@5.1.1` | found `execa@8.0.1`  |
| image | image |

PR #910, opened 2024-08-10
chore: Resolve version regression issues related to optional peer deps.

1natsu172:
related: #908

With simple solution. 

If want an advanced solution to this problem, we can have the renovate-bot do `lockfilemaintenance`, but at the cost of having a bot.

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,6 +11,7 @@
     "test:coverage": "pnpm -r --sequential run test:coverage",
     "prepare": "simple-git-hooks",
     "prepublish": "pnpm -s build",
+    "postinstall": "pnpm i --resolution-only",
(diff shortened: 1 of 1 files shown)


1natsu172:
If try merge this PR, we will have a lockfile conflict with #908. We will need to rebase to HEAD after #908, then generate the latest lockfile and add the commit to this PR.

aklinker1:
I'm gonna close this in favor of just running `pnpm i --fix-lockfile` if something breaks. The post-install already takes a lot of time in this repo.

1natsu172:
Yes you are right. This `postinstall` method is simple but have a performance problem and it's over-engineering to try to do something about it 😄

PR #916, opened 2024-08-14
fix(types): Don't report type errors when using string templates with `browser.i18n.getMessage`

aklinker1:
This fixes: https://discord.com/channels/1212416027611365476/1272982191764144248/1272982191764144248

> I inspected the generated types and it's a bunch of function overloads where `messageName` is basically a literal string that can be passed through as argument to `.getMessage()`. It's nice for development and auto-completion, but it doesn't allow me to pass a template string with an argument that is typed correctly.
> 
> It allows me to write it as:
> ```ts
> browser.i18n.getMessage('mode_upscale')
> ```
> 
> Also this works:
> ```ts
> const myMode = 'upscale'
> browser.i18n.getMessage(`mode_${myMode}`)
> ```
> 
> But this does not (on type level):
> ```ts
> const modeOptions = modes.map((mode) => ({
>   label: browser.i18n.getMessage(`mode_${mode}`),
> }))
> ```
> 
> Am I doing something wrong?

### DIFF
--- a/packages/wxt/src/core/utils/building/generate-wxt-dir.ts
+++ b/packages/wxt/src/core/utils/building/generate-wxt-dir.ts
@@ -164,7 +164,7 @@ declare module "wxt/browser" {
     const comment =
       commentLines.length > 0
         ? `    /**\n${commentLines.map((line) => `     * ${line}`.trimEnd()).join('\n')}\n     */\n`
-        : '';
+        : '    ';
     return `${comment}getMessage(
       messageName: ${keyType},
       substitutions?: string | string[],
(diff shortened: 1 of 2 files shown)


packages/wxt/src/core/utils/building/generate-wxt-dir.ts
```diff
@@ -150,18 +150,43 @@ declare module "wxt/browser" {
...
+    const commentLines: string[] = [];
+    if (description) commentLines.push(...description.split('\n'));
+    if (translation) {
+      if (commentLines.length > 0) commentLines.push('');
+      commentLines.push(`"${translation}"`);
+    }
+    const comment =
+      commentLines.length > 0
+        ? `    /**\n${commentLines.map((line) => `     * ${line}`).join('\n')}\n     */\n`
+        : '';
```
aklinker1, line 167:
```suggestion
        : '    ';
```

PR #919, opened 2024-08-16
perf: Ignore non-source code from the file watcher

1natsu172:
related #866 

I noticed an over-firing of `resolveConfig` due to `reloadOnChange`, so I will fix that.

At first, I thought need to cache the `resolveConfig`, but don't need to cache it now. This change will reduce build time. And I haven't figured out the cause……but after repeating the launch a few times, launch speed also faster. (The 1st launch still a little slow) 🤷 

##### About the repeating launch

1. `cd wxt-demo`
2. `pnpm i` (rebuild the project)
3. run 1st: `node ./node_modules/wxt/bin/wxt.mjs`
    - browser launch duration about 1.3s
4. repeat run 2nd..3rd..: `node ./node_modules/wxt/bin/wxt.mjs`
    - browser launch duration about 600ms
    - decreased

### DIFF
--- a/packages/wxt/src/core/create-server.ts
+++ b/packages/wxt/src/core/create-server.ts
@@ -145,8 +145,6 @@ function createFileReloader(server: WxtDevServer) {
   const changeQueue: Array<[string, string]> = [];
 
   return async (event: string, path: string) => {
-    await wxt.reloadConfig();
-
     changeQueue.push([event, path]);
 
     await fileChangedMutex.runExclusive(async () => {
@@ -156,6 +154,8 @@ function createFileReloader(server: WxtDevServer) {
         .splice(0, changeQueue.length)
         .map(([_, file]) => file);
       if (fileChanges.length === 0) return;
+
+      await wxt.reloadConfig();
 
       const changes = detectDevChanges(fileChanges, server.currentOutput);
       if (changes.type === 'no-change') return;
--- a/packages/wxt/src/core/create-server.ts
+++ b/packages/wxt/src/core/create-server.ts
@@ -147,9 +147,6 @@ function createFileReloader(server: WxtDevServer) {
   return async (event: string, path: string) => {
     await wxt.reloadConfig();
 
-    // Here, "path" is a non-normalized path (ie: C:\\users\\... instead of C:/users/...)
-    if (path.startsWith(wxt.config.outBaseDir)) return;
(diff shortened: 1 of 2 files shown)


1natsu172:
I honestly don't know if removing the `outDir`from the target is a good solution. If it affects HMR etc, we need to consider a different approach.

aklinker1:
> I honestly don't know if removing the `outDir`from the target is a good solution. If it affects HMR etc, we need to consider a different approach.

Yup, this is fine. I thought I already had code inside the listener to check this. Are you on windows?

aklinker1:
Yeah, I do. Here:

https://github.com/wxt-dev/wxt/blob/main/packages%2Fwxt%2Fsrc%2Fcore%2Fcreate-server.ts#L151-L152

It should definitely go in the watcher settings directly though, so I like this PR. Can you also add the `wxtDir` to the ignored list? Then remove those if statements?

1natsu172:
I hadn't noticed the guards just below it 😅 Ok, just change and committed. Please check it.

packages/wxt/src/core/create-server.ts
```diff
@@ -147,9 +147,6 @@ function createFileReloader(server: WxtDevServer) {
   return async (event: string, path: string) => {
     await wxt.reloadConfig();
```
aklinker1, line 148: Last change - let's move this down to right before line 160 (before `detectDevChanges` is called) so it only runs once if multiple files are changed.

1natsu172: Yeah, it's nice at call the queue execution timing. Commit pushed.

aklinker1 (approved):
Alright, ran the benchmark, and looks like this solved the majority of the issues! Great find.

Screenshot 2024-08-16 at 2 51 59 PM

There's probably more issues, and I'd still like to investigate the other commits, because I don't think this code has changed for a long time. So there are potentially more performance improvements to be had.

aklinker1:
Alright, ran the benchmark for `wxt-v0.18.4` after cherry-picking this PR onto that reference. It had no effect. Not really sure what that means, still gonna look into the other bad commits.

1natsu172:
Yes, this PR made it dramatically faster, but it is possible that just eliminated something of the side effects. 

As I said in the discussion, something in the userModules codes or something dependency modules...vite, vite-node…umm…? 

Let's continue the conversation at #866.

PR #924, opened 2024-08-18
perf: Ignore output directories in all `vite.watcher`

1natsu172:
From my research on https://github.com/wxt-dev/wxt/discussions/866#discussioncomment-10373242. Add the watcher config to the base config. This will make it ignored on `importEntrypoint`'s vite server. The performance is slightly more stable because vite is internally handled by the watcher.

### DIFF
--- a/packages/wxt/src/core/builders/vite/index.ts
+++ b/packages/wxt/src/core/builders/vite/index.ts
@@ -59,6 +59,11 @@ export async function createViteBuilder(
       config.build.sourcemap = 'inline';
     }
 
+    config.server ??= {};
+    config.server.watch = {
+      ignored: [`${wxtConfig.outBaseDir}/**`, `${wxtConfig.wxtDir}/**`],
+    };
+
     config.plugins ??= [];
     config.plugins.push(
       wxtPlugins.download(wxtConfig),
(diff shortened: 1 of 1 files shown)


1natsu172:
BTW, the own bench.
image

aklinker1 (commented):
Should we just disable file watching when importing TS files?

1natsu172:
Yup, be able to `server.watch = null`, but even after doing that, there was no further improve in performance. 
image

aklinker1 (approved):
Interesting, I don't get that initial spike in my benchmarks. What OS do you use @1natsu172?

1natsu172:
@aklinker1 Oh really? I use MacOS Sonoma 14.2.1 and chrome(127.0.6533.120). My mac mini is an early model with poor memory(8GB), so this may affect it… 🤔 

Also, I'm curious if the same bench trend exists for Windows and Linux. I don't have them, so we need to ask someone to help.

aklinker1:
I'll run them on Linux, I just uninstalled windows the other week though, I got so fed up with it. Was only using it for gaming, but I was able to get steam up and running on Arch lol.

1natsu172:
Good timing uninstall lol. Please try it on Arch Linux!

aklinker1:
Yeah, I don't see the initial slowdown on linux. My PC is old and slow, but the startup times are very consistent

```
165ea51 1174 1135 1132 1131 1133 1134 1132 1133 1133 1143 1133 1135 1131 1130 1137 1130 1133 1130 1132 1144 
```

aklinker1:
I'm gonna go ahead and merge this!

1natsu172:
Great, very stable. If so, then the initial slow down and spikes could be due to MacOS and specs. Thank you check on Linux and merge!

PR #955, opened 2024-09-01
feat: continuous publish using pkg.pr.new

kongmoumou:
This PR will enable continuous publish wxt packages for every commit & PR on main automatically, instead of publishing it manually using npm command. It would be super easy and fast for testing locally, just update package version pointed to specific commit. It works in my fork, result like this ⬇️

image

All u need to do is [install the Github Application](https://github.com/apps/pkg-pr-new).

Detail see https://github.com/stackblitz-labs/pkg.pr.new

### DIFF
--- a/.github/workflows/pkg.pr.new.yml
+++ b/.github/workflows/pkg.pr.new.yml
@@ -9,22 +9,9 @@ on:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
-
+    runs-on: ubuntu-22.04
     steps:
-      - name: Checkout code
-        uses: actions/checkout@v4
-
-      - run: corepack enable
-      - uses: actions/setup-node@v4
-        with:
-          node-version: 20
-          cache: 'pnpm'
-
-      - name: Install dependencies
-        run: pnpm install
-
-      - name: Build
-        run: pnpm -r build
-
-      - run: pnpx pkg-pr-new publish './packages/*'
+      - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


aklinker1:
Seems intriguing, would simplify releasing test versions. Will review this when I have time!

.github/workflows/pkg.pr.new.yml
```diff
@@ -0,0 +1,30 @@
...
+          node-version: 20
+          cache: 'pnpm'
+
+      - name: Install dependencies
+        run: pnpm install
+
+      - name: Build
+        run: pnpm -r build
+
+      - run: pnpx pkg-pr-new publish './packages/*'
```
aklinker1, line 30: Lets get this more inline with the rest of the workflows in the repo. Also added `--compact`.

```suggestion
jobs:
  build:
    runs-on: ubuntu-22.04
    steps:
      - uses: actions/checkout@v4
      - uses: ./.github/actions/setup
      - run: pnpm buildc all
      - run: pnpx pkg-pr-new publish --compact './packages/*'
```

aklinker1 (changes requested):
Alright, this looks good. Cleanup the workflow a bit and we'll merge!

.github/workflows/pkg.pr.new.yml
kongmoumou: done~

aklinker1 (approved):
Thanks for the PR!

PR #966, opened 2024-09-03
fix: Properly overload `import.meta.env` with WXT's own environment globals

### LINKED ISSUE #935: WXT specific `import.meta.env` constants not showing up in editor
https://discord.com/channels/1212416027611365476/1224507127863967784/1277527012726145156

1natsu172:
fix #935

I copied vite's implement style https://github.com/vitejs/vite/blob/main/packages/vite/types/importMeta.d.ts. It seems that module augmentation is not allowed because `import` is a top-level scope.

TS docs:
> You can’t declare new top-level declarations in the augmentation — just patches to existing declarations.

**completion screenshot**
image


packages/wxt/vitest.globalSetup.ts
```diff
@@ -0,0 +1,27 @@
...
+  const e2eDistPath = './e2e/dist/';
+  if (existsSync(e2eDistPath)) {
+    try {
+      await rm(e2eDistPath, { recursive: true, force: true });
+      consola.info(`cleaned up ${e2eDistPath}`);
+    } catch (error) {
+      consola.error(error);
+    }
+  }
+}
```
1natsu172, line 27: The globalSetup and teardown was prepared as a side note. `e2e/dist` had swelled up to nearly GB.

packages/wxt/src/core/utils/building/generate-wxt-dir.ts
```diff
@@ -201,7 +201,6 @@ async function getGlobalsDeclarationEntry(): Promise<WxtDirFileEntry> {
     path: 'types/globals.d.ts',
     text: [
       '// Generated by wxt',
-      'export {}',
```
aklinker1, line 204: Very interesting, I think I copied this from Nuxt? I had always questioned why they had the export there, but I remember it working at some point lol.

packages/wxt/vitest.globalSetup.ts
aklinker1: Let's extract this change to a separate PR, but it's a good fix. I thought I was doing this already though :/ maybe I'm thinking of a different project lol

aklinker1 (changes requested):

aklinker1:
Merged this PR with `main` to fix the checks as well.

packages/wxt/vitest.globalSetup.ts
1natsu172: Haha common things. Okay, I'll separate this and submit!

1natsu172:
Haha, I missed the local rebase but it's maybe ok now 😅

aklinker1 (approved):

PR #968, opened 2024-09-03
chore: Cleanup leftover E2E test artifacts 

1natsu172:
Separated from #966 .

Cleanup the `e2e/dist` after all tests.

### DIFF
--- a/packages/wxt/vitest.globalSetup.ts
+++ b/packages/wxt/vitest.globalSetup.ts
@@ -1,27 +1,19 @@
-import { rm } from 'fs/promises';
-import { existsSync } from 'fs';
-import consola from 'consola';
+import { exists, rm } from 'fs-extra';
 
-let teardownHappened = false;
+let setupHappened = false;
 
 export async function setup() {
+  if (setupHappened) {
+    throw new Error('setup called twice');
+  }
+
+  setupHappened = true;
+
   // @ts-expect-error
   globalThis.__ENTRYPOINT__ = 'test';
-}
-
-export async function teardown() {
-  if (teardownHappened) {
-    throw new Error('teardown called twice');
-  }
-  teardownHappened = true;
 
   const e2eDistPath = './e2e/dist/';
-  if (existsSync(e2eDistPath)) {
-    try {
-      await rm(e2eDistPath, { recursive: true, force: true });
-      consola.info(`cleaned up ${e2eDistPath}`);
-    } catch (error) {
-      consola.error(error);
-    }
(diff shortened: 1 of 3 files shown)


packages/wxt/vitest.globalSetup.ts
```diff
@@ -0,0 +1,27 @@
+import { rm } from 'fs/promises';
+import { existsSync } from 'fs';
```
aklinker1, line 2:
```suggestion
import { rm, exists } from 'node:fs/promises';
```

packages/wxt/vitest.globalSetup.ts
```diff
@@ -0,0 +1,27 @@
...
+}
+
+export async function teardown() {
+  if (teardownHappened) {
+    throw new Error('teardown called twice');
+  }
+  teardownHappened = true;
+
+  const e2eDistPath = './e2e/dist/';
+  if (existsSync(e2eDistPath)) {
```
aklinker1, line 19:
```suggestion
  if (await exists(e2eDistPath)) {
```

packages/wxt/vitest.globalSetup.ts
```diff
@@ -0,0 +1,27 @@
...
+    try {
+      await rm(e2eDistPath, { recursive: true, force: true });
+      consola.info(`cleaned up ${e2eDistPath}`);
```
aklinker1, line 22:
```suggestion
```

Don't think we need this log.

aklinker1 (changes requested):
Some small changes. Should we delete the files during `setup` instead of `teardown`? I often find I need to explore the files after running tests to figure out what's going on.

packages/wxt/vitest.globalSetup.ts
1natsu172: Alright, following your style. I'm not picky about it.

packages/wxt/vitest.globalSetup.ts
1natsu172: Oh sorry fs-extra already existed. I didn't notice.

1natsu172:
> Some small changes. Should we delete the files during `setup` instead of `teardown`? I often find I need to explore the files after running tests to figure out what's going on.

Given that use case, it might be better that way, I'll check if it's ok to cleanup with setup.

1natsu172:
> I'll check if it's ok to cleanup with setup.

There was no problem. Let's do this.

aklinker1 (approved):

PR #977, opened 2024-09-12
chore(package): set type to "module" for Svelte template

rxliuli:
When adding shadcn-svelte, I discovered that the template project's type was not set to "module," causing errors when using "export default" in postcss.js. Since ESM is now widely popular, shouldn't setting "type: module" by default be fine?


aklinker1 (approved):
The template was based on Vite's svelte template, and looks like they also update it to use ESM. So let's do it!

Issue #938, opened 2024-08-28
Content script is reloaded twice when the file is saved.


kazu-ux:
### Describe the bug

Content script is reloaded twice when the file is saved.

### Reproduction

https://github.com/user-attachments/assets/433eb77a-6924-435d-bdd5-04a8a899d7f9

### Steps to reproduce

Run `npm install` followed by `npm run dev`

### System Info

```shell
System:
    OS: Windows 10 10.0.19045
    CPU: (8) x64 AMD Ryzen 5 3400G with Radeon Vega Graphics 

    Memory: 1.61 GB / 13.94 GB
  Binaries:
    Node: 20.11.1 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.22 - C:\Program Files\nodejs\yarn.CMD
    npm: 10.2.4 - C:\Program Files\nodejs\npm.CMD
    pnpm: 9.9.0 - C:\Program Files\nodejs\pnpm.CMD
    bun: 1.1.10 - ~\AppData\Local\reflex\bun\bin\bun.EXE     
    Watchman: 20230806.083715.0 - D:\watchman\bin\watchman.EXE
  Browsers:
    Edge: Chromium (127.0.2651.74)
    Internet Explorer: 11.0.19041.4355
  npmPackages:
    wxt: ^0.19.7 => 0.19.7
```

### Used Package Manager

npm

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/wxt-dev/wxt/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://wxt.dev/guide/installation.html).
- [X] Check that there isn't [already an issue](https://github.com/wxt-dev/wxt/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/wxt-dev/wxt/discussions) or join our [Discord Chat Server](https://discord.gg/ZFsZqGery9).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

aklinker1:
<details>
<summary>Can't reproduce:</summary>

https://github.com/user-attachments/assets/5fab4a7f-24b1-4f0c-976d-c6bb03b536c6

</details>

Can you apply this patch and share the logs?

```diff
[...]

PR #978, opened 2024-09-12
perf: Reduce hypersensitive onChange of watcher

1natsu172:
refer #938 

Debounce the watcher's onChange.

### DIFF
--- a/packages/wxt/package.json
+++ b/packages/wxt/package.json
@@ -82,6 +82,7 @@
   },
   "dependencies": {
     "@aklinker1/rollup-plugin-visualizer": "5.12.0",
+    "@dotenvx/dotenvx": "^1.14.1",
     "@types/chrome": "^0.0.269",
     "@types/webextension-polyfill": "^0.10.7",
     "@webext-core/fake-browser": "^1.3.1",
--- a/packages/wxt/package.json
+++ b/packages/wxt/package.json
@@ -81,8 +81,8 @@
     "prepack": "pnpm build"
   },
   "dependencies": {
-    "@types/chrome": "^0.0.269",
     "@aklinker1/rollup-plugin-visualizer": "5.12.0",
+    "@types/chrome": "^0.0.269",
     "@types/webextension-polyfill": "^0.10.7",
     "@webext-core/fake-browser": "^1.3.1",
     "@webext-core/isolated-element": "^1.1.2",
@@ -116,6 +116,7 @@
     "ohash": "^1.1.3",
     "open": "^10.1.0",
     "ora": "^8.1.0",
+    "perfect-debounce": "^1.0.0",
     "picocolors": "^1.0.1",
     "prompts": "^2.4.2",
     "publish-browser-extension": "^2.1.3",
--- a/packages/wxt/package.json
+++ b/packages/wxt/package.json
@@ -82,7 +82,6 @@
   },
   "dependencies": {
     "@aklinker1/rollup-plugin-visualizer": "5.12.0",
-    "@dotenvx/dotenvx": "^1.14.1",
     "@types/chrome": "^0.0.269",
     "@types/webextension-polyfill": "^0.10.7",
     "@webext-core/fake-browser": "^1.3.1",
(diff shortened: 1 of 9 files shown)


packages/wxt/package.json
```diff
@@ -81,8 +81,8 @@
     "prepack": "pnpm build"
   },
   "dependencies": {
-    "@types/chrome": "^0.0.269",
     "@aklinker1/rollup-plugin-visualizer": "5.12.0",
+    "@types/chrome": "^0.0.269",
```
1natsu172, line 85: auto sorted by pnpm

packages/wxt/src/core/create-server.ts
```diff
@@ -130,7 +131,9 @@ export async function createServer(
   });
 
   // Listen for file changes and reload different parts of the extension accordingly
-  const reloadOnChange = createFileReloader(server);
+  const reloadOnChange = debounce(createFileReloader(server), 3e2, {
+    leading: true,
+  });
```
1natsu172, line 136: I'm expecting unrealistic and excessive changes by machine. Not sure if the wait-time should be shorter or longer.... 🤔

pnd280: 300ms still causes the double reloads issue, I have to bump it to at least 3s

1natsu172: @pnd280 I turned off trailing, so 300ms might have been OK. Please check. If it's still NG, please send us a captured video of your screen.

1natsu172:
## NOTE: perfect-debounce behavior

<details><summary>Details</summary>
<p>

### `{ leading: true, trailing: true }`
image

### `{ leading: false, trailing: false }`
image


### `{ leading: true, trailing: false }`
image


### `{ leading: false, trailing: true }`
image

</p>
</details> 

`maxWait`
**There is no `maxWait`, so `leading` has priority. unlike lodash, even `{ leading: true, trailing: true }` does not fire a total of two times.**

packages/wxt/src/core/wxt.ts
```diff
@@ -31,6 +32,8 @@ export async function registerWxt(
   // See https://github.com/wxt-dev/wxt/issues/873#issuecomment-2254555523
   process.env.NODE_ENV ??= command === 'serve' ? 'development' : 'production';
 
+  loadDotenv({ path: ['.env.wxt', '.env'], quiet: true });
```
1natsu172, line 35: Is this the right place to load? More top level? (cli.mjs?)

packages/wxt/src/types.ts
```diff
@@ -904,6 +904,10 @@ export interface ConfigEnv {
    * details.
    */
   manifestVersion: 2 | 3;
+  /**
+   * WatchDebounce passed in from `.env.wxt` via the `WXT_WATCH_DEBOUNCE`. Which can suppress cases where dev server reloads are oversensitive.
+   */
+  watchDebounce: number;
```
1natsu172, line 910: Is this the right place to set in? Or should separate to another type and config property?

1natsu172 (commented):
@aklinker1 I implemented env variable loading and set the default value to a tentative 800ms (based on their logs). Please review when you have time!

aklinker1 (approved):
@1natsu172 sorry for the slow review on this! I made a couple of changes:

1. Move config from `ConfigEnv` to `ResolvedConfig` because environment variables isn't what the "env" is referencing in "ConfigEnv". That object is more for storing information about the current build setup than storing all environment variables
2. Updated .env loading behavior to match Vite's. See https://github.com/wxt-dev/wxt/blob/ef30d9aac1e7353a96ecf30c53df39b74a707516/packages/wxt/src/core/utils/env.ts
3. Moved loading of environment variables into `resolveConfig` so it refreshes the environment variables on save
4. Replaced `@dotenvx/dotenvx` with `dotenv` since we don't need the CLI, just the env loading behavior.

Otherwise it looks good! Thanks for fielding this one!

1natsu172:
@aklinker1 Oh you've already made a lot of changes and adjustments. Thanks!

aklinker1:
@1natsu172 want me to revert any of that? I figured since it took me a while to review, I'd make the changes myself to spend things up

1natsu172:
@aklinker1 No, it's okay! I was concerned about the type of env file, timing of loading, etc., but everything is clear now 👍

PR #990, opened 2024-09-23
feat(storage): Support storage items in batch functions

Timeraa:
This pull request includes significant updates to the storage API documentation and implementation to support bulk operations and improve clarity. The most important changes include the addition of bulk operation methods in the documentation, corrections to grammar and phrasing, and enhancements to the `WxtStorage` interface to support these new methods.

### Documentation Updates:
* Added a new section on bulk operations in `docs/guide/extension-apis/storage.md`, detailing methods like `getItems`, `getItemMetas`, `setItemValues`, `setItemMetas`, and `removeItems`.
* Corrected grammar and phrasing issues in various parts of the storage API documentation, improving readability and accuracy. [[1]](diffhunk://#diff-289e6b83a68b7127245ccd599012a905e62eb103e37764c7149aaf0cbb29f3cfL233-R340) [[2]](diffhunk://#diff-289e6b83a68b7127245ccd599012a905e62eb103e37764c7149aaf0cbb29f3cfL281-R390) [[3]](diffhunk://#diff-289e6b83a68b7127245ccd599012a905e62eb103e37764c7149aaf0cbb29f3cfL300-R403)

### Code Enhancements:
* Enhanced the `createStorage` function in `packages/wxt/src/storage.ts` to include new methods for bulk operations: `getItemMetas`, `setItemValues`, `setItemMetas`, and `removeItems`. [[1]](diffhunk://#diff-0acc00d70515e675fbae7f4724a31981a7b394328490d24e0d56008ef9522005R392-R502) [[2]](diffhunk://#diff-0acc00d70515e675fbae7f4724a31981a7b394328490d24e0d56008ef9522005L474-R612) [[3]](diffhunk://#diff-0acc00d70515e675fbae7f4724a31981a7b394328490d24e0d56008ef9522005L526-R667) [[4]](diffhunk://#diff-0acc00d70515e675fbae7f4724a31981a7b394328490d24e0d56008ef9522005R717-R767)
* Updated the `WxtStorage` interface to support the new bulk operation methods, ensuring type safety and consistency. [[1]](diffhunk://#diff-0acc00d70515e675fbae7f4724a31981a7b394328490d24e0d56008ef9522005L474-R612) [[2]](diffhunk://#diff-0acc00d70515e675fbae7f4724a31981a7b394328490d24e0d56008ef9522005L526-R667) [[3]](diffhunk://#diff-0acc00d70515e675fbae7f4724a31981a7b394328490d24e0d56008ef9522005R717-R767)
* Improved the `getItems` method to maintain the order of keys and handle different key types more efficiently.

These changes collectively enhance the functionality and usability of the storage API, making it more powerful and easier to use for developers.


docs/guide/extension-apis/storage.md
```diff
@@ -139,6 +139,87 @@ const unwatch = showChangelogOnUpdate.watch((newValue) => {
...
+// {
+//   item1: { ...metadata of item1 },
+//   item2: { ...metadata of item2 },
+//   item3: { ...metadata of item3 }
+// }
+```
+
+#### Setting Values of Multiple Storage Items
+
+You can set values for multiple storage items in a single operation using `setStorageItemValues`:
```
aklinker1, line 192:
```suggestion
You can set values for multiple storage items in a single operation using `setItemValues`:
```

docs/guide/extension-apis/storage.md
```diff
@@ -139,6 +139,87 @@ const unwatch = showChangelogOnUpdate.watch((newValue) => {
...
+```ts
+await storage.setItemValues(
+  { item1, item2, item3 },
+  { item1: 'new value', item2: 42, item3: true },
+);
+```
+
+#### Setting Metadata of Multiple Storage Items
+
+To update metadata for multiple storage items, use `setStorageItemMetas`:
```
aklinker1, line 203:
```suggestion
To update metadata for multiple storage items, use `setStorageItemMetas`. Similar to `setMeta`, this overwrites the provided metadata keys for each item, leaving others untouched:
```

docs/guide/extension-apis/storage.md
```diff
@@ -139,6 +139,87 @@ const unwatch = showChangelogOnUpdate.watch((newValue) => {
...
+  },
+);
+```
+
+#### Deleting Values of Multiple Storage Items
+
+You can remove the values (and optionally metadata) of multiple storage items efficiently using `deleteItemValues`:
+
+```ts
+await storage.deleteItemValues({ item1, item2 }, { removeMeta: true });
```
aklinker1, line 220: I forget, is meta not deleted by default?

aklinker1 (changes requested):
Can you add tests? Specifically making sure the minimal number of storage calls are added.

Otherwise this is looking great! Thanks for adding this

docs/guide/extension-apis/storage.md
Timeraa: Will have a look

Timeraa:
> Can you add tests? Specifically making sure the minimal number of storage calls are added.
> 
> Otherwise this is looking great! Thanks for adding this

Will do! _Did the PR on a sleepless night_

aklinker1:
Haha I noticed you got it out like as soon as you asked about it on discord lol.

docs/guide/extension-apis/storage.md
```diff
@@ -139,6 +139,87 @@ const unwatch = showChangelogOnUpdate.watch((newValue) => {
...
+```ts
+await storage.setItemValues(
+  { item1, item2, item3 },
+  { item1: 'new value', item2: 42, item3: true },
+);
+```
+
+#### Setting Metadata of Multiple Storage Items
+
+To update metadata for multiple storage items, use `setStorageItemMetas`. Similar to `setMeta`, this overwrites the provided metadata keys for each item, leaving others untouched:
```
aklinker1, line 203:
```suggestion
To update metadata for multiple storage items, use `setItemMetas`. Similar to `setMeta`, this overwrites the provided metadata keys for each item, leaving others untouched:
```

packages/wxt/src/__tests__/storage.test.ts
```diff
@@ -896,6 +896,243 @@ describe('Storage Utils', () => {
     });
   });
 
+  describe('Batch Item Operations', () => {
+    beforeEach(() => {
+      fakeBrowser.reset();
+      storage.unwatch();
+    });
+
```
aklinker1, line 904: Top level describe is already running these.

```suggestion
```

packages/wxt/src/__tests__/storage.test.ts
```diff
@@ -896,6 +896,243 @@ describe('Storage Utils', () => {
...
+        expect(localGetSpy).toHaveBeenCalledTimes(1);
+        expect(sessionGetSpy).toHaveBeenCalledTimes(1);
+      });
+    });
+
+    describe('setItemValues', () => {
+      it('should set the values of multiple storage items efficiently', async () => {
+        const item1 = storage.defineItem<number>('local:item1');
+        const item2 = storage.defineItem<string>('session:item2');
```
aklinker1, line 1005: Add a third item for local storage here as well.

packages/wxt/src/__tests__/storage.test.ts
```diff
@@ -896,6 +896,243 @@ describe('Storage Utils', () => {
...
+
+    describe('setItemMetas', () => {
+      it('should set the metadata of multiple storage items efficiently', async () => {
+        const item1 = storage.defineItem<number, { v: number }>('local:item1');
+        const item2 = storage.defineItem<string, { date: number }>(
+          'session:item2',
+        );
+
+        const localSetSpy = vi.spyOn(fakeBrowser.storage.local, 'set');
+        const sessionSetSpy = vi.spyOn(fakeBrowser.storage.session, 'set');
```
aklinker1, line 1040: Also need to make sure the storage is only gotten once.

packages/wxt/src/__tests__/storage.test.ts
```diff
@@ -896,6 +896,243 @@ describe('Storage Utils', () => {
...
+        const value2 = await item2.getValue();
+
+        expect(value1).toBeNull();
+        expect(value2).toBeNull();
+
+        expect(localRemoveSpy).toHaveBeenCalledTimes(2);
+        expect(sessionRemoveSpy).toHaveBeenCalledTimes(2);
+      });
+
+      it('should not delete metadata when removeMeta option is true', async () => {
```
aklinker1, line 1093:
```suggestion
      it('should not delete metadata when removeMeta option is false', async () => {
```

packages/wxt/src/__tests__/storage.test.ts
```diff
@@ -896,6 +896,243 @@ describe('Storage Utils', () => {
...
+        await storage.deleteItemValues({
+          item1,
+          item2,
+        });
+
+        const value1 = await item1.getValue();
+        const value2 = await item2.getValue();
+
+        expect(value1).toBeNull();
+        expect(value2).toBeNull();
```
aklinker1, line 1087: Check meta as well

aklinker1 (commented):
Now... I was reviewing the code, and noticed something. We already have some batch APIs:

docs/guide/extension-apis/storage.md
aklinker1: Oh, no, looks like I was wrong. We don't delete meta by default: https://github.com/wxt-dev/wxt/blob/4a1bd41df9f26de337e1c9573e5794518be0a726/packages/wxt/src/__tests__/storage.test.ts#L193-L204

Timeraa:
> Now... I was reviewing the code, and noticed something. We already have some batch APIs: 
> 
> 
> 
> ```ts
> 
> await storage.getItems(["local:installDate", "session:someCounter"]);
> 
> ```
> 
> 
> 
> What about instead of adding new functions, just adding them to the existing APIs:
> 
> 
> 
> ```ts
> 
> const installDate = storage.defineItem(...);
> 
> 
> 
> await storage.getItems([installDate, "session:someCounter"]);
> 
> ```
> 
> 
> 
> That way they use the same style, arrays instead of objects.

Yeah that sounds like a better idea

docs/guide/extension-apis/storage.md
aklinker1: For this PR, let's leave it as is and update this PR match. I don't recall the exact reason I choose to do that, but I know I chose it for some reason lol.

If we need to change it, it will have to be a breaking change.

packages/wxt/src/__tests__/storage.test.ts
```diff
@@ -384,6 +384,56 @@ describe('Storage Utils', () => {
...
+          await storage.setItem('local:key1', 'newer value1');
+          await storage.setItem('session:key2', 'new value2');
+
+          expect(cb1).toBeCalledTimes(2); // Still 2, not called again
+          expect(cb2).toBeCalledTimes(1); // Still 1, not called again
+
+          expectTypeOf(unwatchers).toEqualTypeOf<{
+            'local:key1': Unwatch;
+            'session:key2': Unwatch;
+          }>();
```
aklinker1, line 421: If you're watching a group of keys in a single call, we should return a single `unwatch` function to remove the listener for all of them at once. If someone needs to manage watches individually, they should just call watch with a single item at a time.

```suggestion
          expectTypeOf(unwatchers).toEqualTypeOf<Unwatch>();
```

packages/wxt/src/storage.ts
```diff
@@ -158,7 +172,27 @@ function createStorage(): WxtStorage {
...
+              any
+            >;
+            returnObj[itemKey.storageKey] = value;
+          }
+        });
+        return returnObj;
+      } else {
+        // Otherwise, return the array of results (original behavior)
+        return flatResults;
+      }
```
aklinker1, line 195: I think regardless of the input types, we should return the same thing here. What if we need to add support for mixing keys and items? Should we support that right now? Probably.

packages/wxt/src/storage.ts
```diff
@@ -242,17 +276,33 @@ function createStorage(): WxtStorage {
...
-    watch: (key, cb) => {
-      const { driver, driverKey } = resolveKey(key);
-      return watch(driver, driverKey, cb);
-    },
+    watch: ((
+      keyOrWatchers:
+        | StorageItemKey
+        | Record<StorageItemKey, WatchCallback<any>>,
+      cb?: WatchCallback<any>,
+    ): any => {
```
aklinker1, line 284: Hmm, we use arrays everywhere else... Using an object here breaks that consistency. Not opposed to this, but let me think about it.

packages/wxt/src/storage.ts
```diff
@@ -364,6 +416,148 @@ function createStorage(): WxtStorage {
         migrate,
       };
     },
+    getItemMetas: async <T extends Record<string, WxtStorageItem<any, any>>>(
+      items: T,
+    ) => {
```
aklinker1, line 421: Once again here, we use arrays everywhere else, so I think we should keep using them.

aklinker1 (commented):
@Timeraa this is a big PR and your latest changes don't really match what I was asking for in this comment: https://github.com/wxt-dev/wxt/pull/990#pullrequestreview-2328292648

Would you mind if I implemented the changes for `getItems` so you have a better idea what direction I think we should go in?

Timeraa:
> @Timeraa this is a big PR and your latest changes don't really match what I was asking for in this comment: [#990 (review)](https://github.com/wxt-dev/wxt/pull/990#pullrequestreview-2328292648)
> 
> Would you mind if I implemented the changes for `getItems` so you have a better idea what direction I think we should go in?

Go ahead! Was just thinking that made the most sense and is the least breaking in that regard

packages/wxt/src/__tests__/storage.test.ts
Timeraa: Yeah that makes more sense to be fair, shall I change it or will you in your commit?

packages/wxt/src/storage.ts
Timeraa: Looking back at it makes more sense to do it as an array yeah, since you can just get the array index you need

packages/wxt/src/storage.ts
Timeraa: Hmm, but how would you watch multiple keys at once if you were to pass an array? Since you have to tell the watch what you want to watch which wouldn't really work nicely if you had to pass an array of arrays. I could make it a separate function `watchMany` or so if you like?

packages/wxt/src/storage.ts
Timeraa: Ditto

aklinker1:
Here's what my changes would look like if I updated the `getItems` and `removeItems` functions: https://github.com/wxt-dev/wxt/compare/batch-item-change-suggestions

The setter and metadata functions will be more complex, I'm not entirely sure how to type them to continue using the array syntax, but we should keep using arrays when possible, since that's how all the existing functions are written.

Only exception to that would be `watch`... where arrays vs objects would have different behaviors (like in vue's `watch` function).

```ts
// Watch multiple values with a single callback:
storage.watch([item1, item2], ([newItem1, newItem2], [oldItem1, oldItem2]) => {
  // ...
});

// Watch multiple values, each with their own callbacks
storage.watch({
  [item1.key]: (newValue, oldValue) => {
    // ...
  },
  "local:item2": (newValue, oldValue) => {
    // ...
  },
});
// or
storage.watch([
  [item1, (newValue, oldValue) => {
    // ...
  }],
  ["local:item2", (newValue, oldValue) => {
    // ...
  }],
]);
```

Not sure how I want the watch function to work, so I would put that in it's own PR where we can discuss it separately.

Timeraa:
> Here's what my changes would look like if I updated the `getItems` and `removeItems` functions: https://github.com/wxt-dev/wxt/compare/batch-item-change-suggestions
> 
> The setter and metadata functions will be more complex, I'm not entirely sure how to type them to continue using the array syntax, but we should keep using arrays when possible, since that's how all the existing functions are written.
> 
> Only exception to that would be `watch`... where arrays vs objects would have different behaviors (like in vue's `watch` function).
> 
> ```ts
> // Watch multiple values with a single callback:
> storage.watch([item1, item2], ([newItem1, newItem2], [oldItem1, oldItem2]) => {
>   // ...
> });
> 
> // Watch multiple values, each with their own callbacks
> storage.watch({
>   [item1.key]: (newValue, oldValue) => {
>     // ...
>   },
>   "local:item2": (newValue, oldValue) => {
>     // ...
>   },
> });
> // or
> storage.watch([
>   [item1, (newValue, oldValue) => {
>     // ...
>   }],
>   ["local:item2", (newValue, oldValue) => {
>     // ...
>   }],
> ]);
> ```
> 
> Not sure how I want the watch function to work, so I would put that in it's own PR where we can discuss it separately.



> Here's what my changes would look like if I updated the `getItems` and `removeItems` functions: https://github.com/wxt-dev/wxt/compare/batch-item-change-suggestions
> 
> The setter and metadata functions will be more complex, I'm not entirely sure how to type them to continue using the array syntax, but we should keep using arrays when possible, since that's how all the existing functions are written.
> 
> Only exception to that would be `watch`... where arrays vs objects would have different behaviors (like in vue's `watch` function).
> 
> ```ts
> // Watch multiple values with a single callback:
> storage.watch([item1, item2], ([newItem1, newItem2], [oldItem1, oldItem2]) => {
>   // ...
> });
> 
> // Watch multiple values, each with their own callbacks
> storage.watch({
>   [item1.key]: (newValue, oldValue) => {
>     // ...
>   },
>   "local:item2": (newValue, oldValue) => {
>     // ...
>   },
> });
> // or
> storage.watch([
>   [item1, (newValue, oldValue) => {
>     // ...
>   }],
>   ["local:item2", (newValue, oldValue) => {
>     // ...
>   }],
> ]);
> ```
> 
> Not sure how I want the watch function to work, so I would put that in it's own PR where we can discuss it separately.

Thanks, I'll implement it like you described and split the watch into another PR later!

Timeraa:
Looking closer at it @aklinker1, wouldn't it make more sense if getItems returns an array of just the values since you can just use the array index since the order should be the same as the one how you passed it

aklinker1:
Yes, but two comments:

1. That's a breaking change. Lets wait to do that until storage has been pulled out of wxt, and is a separate package.
2. Even though the jsdoc I wrote says "The return order is guaranteed to be the same as the order requested"... I don't think that's true looking at the current implementation. So looking up the key you want to use is kinda required right now.

Timeraa:
> Yes, but two comments:
> 
> 1. That's a breaking change. Lets wait to do that until storage has been pulled out of wxt, and is a separate package.
> 2. Even though the jsdoc I wrote says "The return order is guaranteed to be the same as the order requested"... I don't think that's true looking at the current implementation. So looking up the key you want to use is kinda required right now.

Okay! I'll fix the 2. point though!

Timeraa:
Okay I think I implemented it right now? Hope I didn't over-read something

aklinker1 (approved):
Alright, made lots of changes. Mostly around consistency with the other APIs, using arrays instead of objects.

Left comments around other changes.

aklinker1:
Gonna merge, if there are problems, let's fix in another PR.

aklinker1:
Released in https://github.com/wxt-dev/wxt/releases/tag/wxt-v0.19.14

Issue #1005, opened 2024-09-27
Change `getItem`'s default to `undefined`


avi12:
#### Is your feature request related to a bug?
N/A

### Feature request
Suppose I want to fetch multiple storage values, currently, it'll look like:
```ts
const [value1, value2] = await Promise.all([
  storage.getItem<Type1>("local:key1").then(val => val ?? "foo"),
  storage.getItem<Type2>("local:key2").then(val => val ?? "bar")
]);;
```
I prefer having this code instead:
```ts
const [value1 = "foo", value2 = "bar"] = await Promise.all([
  storage.getItem<Type1>("local:key1"),
  storage.getItem<Type2>("local:key2"),
]);
```
which can be achieved without triggering a TypeScript error by changing `null` to `undefined` in
https://github.com/wxt-dev/wxt/blob/e3f00d279781c03dfd0903b41aada100ae5b1d55/packages/wxt/src/storage.ts#L309

aklinker1:
The storage API has a built-in way of providing default values.

```ts
const [value1, value2] = await Promise.all([
  storage.getItem<Type1>("local:key1", { fallback: "foo" }),
  storage.getItem<Type2>("local:key2", { fallback: "bar" }),
]);
```

The storage APIs were designed to feel similar to `localStorage`, which is why it currently returns `null` instead of `undefined`. At this point, it would be a pretty disruptive breaking change to make.

I'm not necessarily opposed to this, `undefined` does make sense in many ways, but this would be really annoying to migrate to. So many little edge cases and checks (like someone accidentally used `=== null` insetad of `== null`, etc).

avi12:
Huh, I was not ware of the `{ fallback: ... }` second argument

avi12:
But does it actually affect `getItem`'s return type to get rid of `null`?

baraich:
@avi12, I have fixed the type issue and made the pull request ([9516324](https://github.com/baraich/wxt/commit/95163243d3b768670cb84a110804d68cd6ac8669)) and once merged the issue would resolved in production.

PR #1007, opened 2024-09-28
fix: #1005 fixed, by updating type-definations to getItem method.

baraich:
## Changes
Updated the `getItem` function to check if the fallback parameter is provided. If it is, the function will return a non-null type.

### DIFF
--- a/packages/wxt/src/storage.ts
+++ b/packages/wxt/src/storage.ts
@@ -461,7 +461,10 @@ export interface WxtStorage {
    * @example
    * await storage.getItem<number>("local:installDate");
    */
-  getItem<T>(key: StorageItemKey, opts?: GetItemOptions<T>): Promise<T | null>;
+  getItem<T, O extends GetItemOptions<T> = GetItemOptions<T>>(
+    key: StorageItemKey,
+    opts?: O,
+  ): Promise<O extends { fallback: infer F } ? F : T | null>;
   /**
    * Get multiple items from storage. The return order is guaranteed to be the same as the order
    * requested.


Timeraa:
Maybe add a type test into the storage tests for that

Timeraa:
Using https://vitest.dev/guide/testing-types in https://github.com/wxt-dev/wxt/blob/main/packages/wxt/src/__tests__/storage.test.ts

I think @aaronklinker-st needs to add the --typecheck flag to the test commands but you can use `vitest --typecheck` in the `packages/wxt` package

baraich:
> Using https://vitest.dev/guide/testing-types in https://github.com/wxt-dev/wxt/blob/main/packages/wxt/src/__tests__/storage.test.ts
> 
> I think @aklinker1 needs to add the --typecheck flag to the test commands but you can use `vitest --typecheck` in the `packages/wxt` package

@Timeraa Hi, according to your guidance I rank the typecheck command and all the tests in storage.test.ts are passing and I am sure it's safe to merge.
image

baraich:
@aklinker1 or @Timeraa, please merge the pull request asap, I am unable to concentrate on my tomorrows' board examination or if you are not planning to merge please provide appropriate feedback.

Thanks in advance.

baraich:
Also, here is the flow of types.
- When we haven't provided a type hint, or fallback parameter the return type in `unknown`.
image
- Now, I have given a type hint.
image
- And when I have given fallback, then the following behaviour is expected.
image


- Moreover, it has type-safety built in, if I hint the function that type is `number` then I cannot pass anything else than number to fallback parameter.
image

aklinker1 (changes requested):
Can you make the PR editable by maintainers? I can't seem to push the type tests.

https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork

```ts
    it('should return a nullable type when getItem is called without a fallback', async () => {
      const res = await storage.getItem<string>('local:test');
      expectTypeOf(res).toBeNullable();
    });

    it('should return a nullable type when getItem is called without a fallback', async () => {
      const res = await storage.getItem<string>('local:test', {
        fallback: 'test',
      });
      expectTypeOf(res).not.toBeNullable();
    });
```

baraich:
@aklinker1 Hi, I have allowed edits by maintainers.

baraich:
@aklinker1 Hi, I think there is some problem with your code.
image

aklinker1:
No, the test code is correct, your change doesn't work when the first type parameter is passed in.

baraich:
@aklinker1, Yes, you are write, when we pass the type hint and fallback. It defaults to typehint and null condition. Will fix an update. Thanks

baraich:
@aklinker1 Why is it should changes required, I have made the changes please review them. Your reviewing part is left. Thanks.

aklinker1 (approved):
Thanks!

Timeraa (approved):

baraich:
Thanks, for approving the changes. As this is my first time contribution to open source, I am not aware of the steps involved. Are there more steps to this, as the pull request doesn't seem as it has been merged (please correct if wrong) even after you have approved.

aklinker1:
Just had to wait for checks to pass, just getting back to this now!

packages/wxt/src/storage.ts
```diff
@@ -461,7 +461,16 @@ export interface WxtStorage {
...
-  getItem<T>(key: StorageItemKey, opts?: GetItemOptions<T>): Promise<T | null>;
+  getItem<TValue>(
+    key: StorageItemKey,
+    opts: GetItemOptions<TValue> & { fallback: TValue },
+  ): Promise<TValue>;
+
+  getItem<TValue>(
+    key: StorageItemKey,
+    opts?: GetItemOptions<TValue>,
+  ): Promise<TValue | null>;
```
avi12, line 472: I'm pretty sure that to solve #1005 it needs to return `Promise<TValue | null>`

baraich: @avi12 The issue the pull request has solved is the proper type hinting. `null` is more suitable option as this tell the developer explicitly that the is **"no value"** and we should leave `undefined` to JavaScript itself as that indicates something unexpected has happened.

As a developer, we should always go for `null` and JS core will default to `undefined` and this is really important when finding out where the issue arose as one is by developer and another is my accident. Hope you get that now.

Thanks.

avi12: I guess my alternative can be
```ts
const [value1 = "foo", value2 = "bar"] = await Promise.all([
  storage.getItem<string>("local:key1"),
  storage.getItem<string>("local:key2")
]);
```

baraich: Your alternative would only work if the get item function returns undefined, otherwise, the code you wrote would not work as the get function would return for the item you are trying to get in the value for your variables would also be null.

avi12: You're right, my alternative can only be
```ts
const [value1 = "foo", value2 = "bar"] = await Promise.all([
  storage.getItem<string>("local:key1"),
  storage.getItem<string>("local:key2")
]) as [string, string];
```

baraich: This won't work as the value you provided after _=_ sign will only be assigned if the value returned by promise is undefined. The promise returns a null so the value1 would be null not the string you specified as fallback with equals sign.

avi12: My point is that `getItem` not returning `undefined` is very unfortunate for my use case and I'm sure I'm not the only one who holds this opinion

baraich: That is what I am saying return `undefined` should be only kept for compiler as a developer we should return `null`

avi12: I don't think the Chromium team agrees with you, considering that
```ts
const { key1 } = await chrome.storage.local.get("key1");
```
will either have `undefined` or a non-`undefined` value, which lets me do
```ts
const { key1 = "foo" } = await chrome.storage.local.get("key1");
```
which is very maintainable and readable for me

baraich: It's up to the developers liking, I am no one to specify what is right or wrong. The repo owner can guide us further. Thanks.

PR #1041, opened 2024-10-04
[UPDATE] error friendly messages for null and undefined values

windmillcode0:
You may run your i18n files through ai and ai may not be perfect it may introduce undefined or null values this changes gives more human friendly output for the developer to more easily fix the problem

### DIFF
--- a/packages/i18n/src/build.ts
+++ b/packages/i18n/src/build.ts
@@ -166,6 +166,11 @@ function _parseMessagesObject(
       ];
     }
     case 'object':
+      if ([null, undefined].includes(object)) {
+        throw new Error(
+          `While parsing the @wxt-dev/i18n package has found that one of the values in one of your translation files is null or undefined. You may have wanted this value to be transalted from the original please test as necessary`,
+        );
+      }
       if (Array.isArray(object))
         return object.flatMap((item, i) =>
           _parseMessagesObject(path.concat(String(i)), item, depth + 1),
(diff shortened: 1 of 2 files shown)


packages/i18n/src/build.ts
```diff
@@ -166,6 +166,11 @@ function _parseMessagesObject(
       ];
     }
     case 'object':
+      if ([null, undefined].includes(object)) {
+        throw new Error(
+          `While parsing the @wxt-dev/i18n package has found that one of the values in one of your translation files is null or undefined. You may have wanted this value to be transalted from the original please test as necessary`,
+        );
```
aklinker1, line 172: Only the source language should be parsed by the module. That means this isn't a problem with your translation service missing a translation, but a formatting problem with a file maintained by a human. I think we should use a simpler error message that indicates that:

```suggestion
        throw new Error(`Messages file should not contain \`${object}\` (found at "${path.join('.')}"`);
```

aklinker1 (changes requested):
Thanks for the PR, can you add a test? 

https://github.com/wxt-dev/wxt/blob/main/packages/i18n/src/__tests__/build.test.ts

```sh
cd packages/18n
pnpm test
```

packages/i18n/src/build.ts
windmillcode0: Added the test and updated the error message

aklinker1 (approved):

aklinker1:
Will merge next time I'm on github once the checks pass!

aklinker1:
Released in https://github.com/wxt-dev/wxt/releases/tag/i18n-v0.1.1

PR #1042, opened 2024-10-04
perf: Replace `execa` with `nano-spawn` for smaller package size

### LINKED ISSUE #1037: Replace `execa` with `nano-spawn`
### Feature Request

Made by the same devs, just lighter weight and zero dependencies. WXT has lots of sub-dependencies, so reducing them as much as possible makes sense.

https://github.com/sindresorhus/nano-spawn

Timeraa:
Closes #1037


packages/wxt/src/core/package-managers/bun.ts
```diff
@@ -11,8 +11,8 @@ export const bun: WxtPackageManagerImpl = {
     if (options?.all) {
       args.push('--all');
     }
-    const { execa } = await import('execa');
-    const res = await execa('bun', args, { cwd: options?.cwd });
+    const { default: spawn } = await import('nano-spawn');
```
aklinker1, line 14: Let's move this to a regular import as well, was doing dynamic because execa wasn't ESM before. Not sure about now, but nano-spawn is ESM-only.

packages/wxt/e2e/tests/init.test.ts
```diff
@@ -1,18 +1,22 @@
...
-    await execaCommand(`pnpm -s wxt init ${project.root} -t vue --pm npm`, {
-      env: { ...process.env, CI: 'true' },
-      stdio: 'ignore',
-      cwd: WXT_PACKAGE_DIR,
-    });
+    await spawn(
+      'pnpm',
+      ['-s', 'wxt', 'init', project.root, '-t', 'vue', '--pm', 'npm'],
+      {
+        env: { ...process.env, CI: 'true' },
```
aklinker1, line 15: Seems like we don't need to spread process.env into this anymore:

> `options.env`
>
> Type: `object`
> Default: `{}`
>
> Override specific [environment variables](https://en.wikipedia.org/wiki/Environment_variable). Other environment variables are inherited from the current process ([process.env](https://nodejs.org/api/process.html#processenv)).
>
> &ndash; https://www.npmjs.com/package/nano-spawn/v/0.1.0#optionsenv

aklinker1 (commented):
Nice, thanks for hopping on this so fast lol. If you wanna look into any other package size optimizations like this, that'd be great! I know you also mentioned locking down the esbuild versions?

packages/wxt/e2e/tests/init.test.ts
Timeraa: Oh, overread that oopsies, will fix

packages/wxt/src/core/package-managers/bun.ts
Timeraa: Will change

Timeraa:
Dang it check failed, worked locally

aklinker1:
> @Timeraa [chore: idk why you downgraded rollup™](https://github.com/wxt-dev/wxt/pull/1042/commits/213f6314196ee31521266ddccec243a782e4ee0a)

Haha if this happens again, run `pnpm i --fix-lockfile` and it should resolve itself.

aklinker1 (approved):
@Timeraa this good to go? Code looks good to me, and windows tests pass, so I think it's ready to go!

Timeraa:
> @Timeraa this good to go? Code looks good to me, and windows tests pass, so I think it's ready to go!

If you approve then yeah

Issue #1026, opened 2024-10-03
Using @wxt-dev/module-vue together with UnoCSS


mklueh:
### Feature Request / Example Provision
Hello,

first of all thank you for the wxt plugin, it makes things really easy so far.

However, I'm usually using Vue together with UnoCSS and I'm failing to do so with WXT, especially using @wxt-dev/module-vue.

Is this in general possible and could you provide an example if this is already working?

I'm currently running into diverse issues like this

>  ERROR  ../../../node_modules/pkg-types/dist/index.mjs (1:9): "statSync" is not exported by "__vite-browser-external:node:fs", imported by "../../../node_modules/pkg-types/dist/index.mjs".

or this, depending on my config

```
- Preparing...
[plugin:unocss:global:build:bundle] [plugin unocss:global:build:bundle] [unocss] Entry module not found. Did you add `import 'uno.css'` in your main entry?
x Build failed in 2.72s
 ERROR  ../../../node_modules/pkg-types/dist/index.mjs (1:9): "statSync" is not exported by "__vite-browser-external:node:fs", imported by "../../../node_modules/pkg-types/dist/index.mjs".
file: C:/Users/MyPc/Workspace/my-project/node_modules/pkg-types/dist/index.mjs:1:9

1: import { statSync, promises } from 'node:fs';
            ^
2: import { resolve, join, isAbsolute, dirname } from 'pathe';
3: import { resolvePath } from 'mlly';

  file: C:/Users/MyPc/Workspace/my-project/node_modules/pkg-types/dist/index.mjs:1:9

  1: import { statSync, promises } from 'node:fs';
  ^
  2: import { resolve, join, isAbsolute, dirname } from 'pathe';
  3: import { resolvePath } from 'mlly';

  at getRollupError (/C:/Users/MyPc/Workspace/my-project/node_modules/rollup/dist/es/shared/parseAst.js:392:41)
  at error (/C:/Users/MyPc/Workspace/my-project/node_modules/rollup/dist/es/shared/parseAst.js:388:42)
  at Module.error (/C:/Users/MyPc/Workspace/my-project/node_modules/rollup/dist/es/shared/node-entry.js:14900:16)
  at Module.traceVariable (/C:/Users/MyPc/Workspace/my-project/node_modules/rollup/dist/es/shared/node-entry.js:15347:29)
[...]

Issue #596, opened 2024-04-03
UnoCSS Vite plugin failed to pick up import from entrypoint's `index.ts`


HigherOrderLogic:
### Describe the bug

UnoCSS failed to pickup `uno.css` import from `index.ts`.

I open this issue here because the plugin works perfectly fine in vanilla Vite project. Will open an issue upstream if the problem doesnt lie in WXT.

### To Reproduce

https://stackblitz.com/edit/github-c9syzq

You can also try https://github.com/Deuscx/starter-vue-wxt

To reproduce the bug, just run the build script and see the warning pop up at console.

### Expected behavior

No warning

### Screenshots

![image](https://github.com/wxt-dev/wxt/assets/73709188/5e6c455c-1778-4144-abaf-27a6e3c8cb47)

### Environment

```
  System:
    OS: Linux 5.0 undefined
    CPU: (8) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
    Node: 18.18.0 - /usr/local/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 10.2.3 - /usr/local/bin/npm
    pnpm: 8.15.3 - /usr/local/bin/pnpm
  npmPackages:
    wxt: ^0.17.0 => 0.17.11 
```

### Additional context

N/A

aklinker1:
UnoCSS is working just fine. You can see the `mt-2` class is applied correctly in the github repo you shared:

<img width="862" alt="Screenshot 2024-04-03 at 12 45 05 PM" src="https://github.com/wxt-dev/wxt/assets/10101283/b65eb74f-d4d0-41e3-b49d-0c32bb2470dc">

WXT orcastrates multiple vite builds. The two warnings come from building the background and content script, neither of which include the `uno.css` import. So those warnings are working as expected.

I can add a util that helps scope vite plugins to specific entrypoints.

aklinker1:
As of [`v0.17.12`](https://github.com/wxt-dev/wxt/releases/tag/v0.17.12), you can now conditionally add vite plugins based on the entrypoints being built. This will let you exclude unocss for the background and content scripts:

```ts
export default defineConfig({
  // ...

  vite: () => ({
    plugins: [vue()],
  }),
  hooks: {
    "vite:build:extendConfig": (entries, config) => {
[...]

Issue #522, opened 2024-03-04
use unocss 


Idiot-Alex:
### Feature Request

I want to use UnoCSS in the WXT project, but my attempts have been unsuccessful.

I have updated the wxt.config.ts and uno.config.ts files, the contents are as follows. (see Additional context)

But when I launch the project, I receive the following error message:
<img width="876" alt="image" src="https://github.com/wxt-dev/wxt/assets/16541166/b394139b-1de8-47cf-aa28-eacd6745c7d1">

I am able to launch another Vite project normally using the same approach.

#### Is your feature request related to a bug?

N/A

### Additional context

There is my project configs:
1. wxt.config.ts
```
import { defineConfig } from 'wxt';
import vue from '@vitejs/plugin-vue';
import UnoCSS from 'unocss/vite';

// See https://wxt.dev/api/config.html
export default defineConfig({
  imports: {
    addons: {
      vueTemplate: true,
    },
  },
  vite: () => ({
    plugins: [
      vue(),
      UnoCSS(),
    ],
  }),
});

```

2. uno.config.ts
```
// uno.config.ts
import { defineConfig, presetAttributify, presetUno } from 'unocss'

export default defineConfig({
  presets: [
    presetAttributify({ /* preset options */}),
    presetUno(),
    // ...custom presets
  ],
})
```

aklinker1:
Seems to work just fine for me. Followed the vite setup, and I don't get that error.

[uno-css-example.zip](https://github.com/wxt-dev/wxt/files/14482167/uno-css-example.zip)


https://github.com/wxt-dev/wxt/assets/10101283/3d4ce8a3-18ae-43d2-a145-159d83f1bbca


Something seems messed up about your setup, there's that auto-import warning that shouldn't be showing up either, which makes me think you're not running the application correctly? Can you share a repro?

aklinker1:
Also see #596

PR #1043, opened 2024-10-04
feat: unocss module

Timeraa:
This PR should make the use of unocss with WXT a fair share easier, see reference issues: #522 #1026
Makes the need for this obsolete: #596

@aklinker1 Any way we can have the `vite:devServer:extendConfig` have entrypoints as well because then I can get rid of the "Entry module not found." warning you will get when it runs over entrypoints you don't want to use.

Also I am a bit uncertain but I think that `unocss.entrypoints` should be required so that you don't run it on files where it's not used, what's your opinion?


aklinker1:
@Timeraa I made some changes to simplify things:

1. Convert `entrypoints` to `excludeEntrypoints` and use entrypoint names instead of paths
2. Add `configOrPath` option to pass config into the vite plugin
3. Added plugin to demo extension for testing.

> Any way we can have the vite:devServer:extendConfig have entrypoints as well because then I can get rid of the "Entry module not found." warning you will get when it runs over entrypoints you don't want to use.

There isn't unfortunately :/. I see the warning you're talking about, don't know how to remove it.

> Also I am a bit uncertain but I think that `unocss.entrypoints` should be required so that you don't run it on files where it's not used, what's your opinion?

I think leaving it as optional is fine. I added a default of `["background"]` for it, so at least the background will be automatically removed. The majority of the time, people will use unocss in most entrypoints. The main reason to add a CSS framework is because you've got lots of CSS, and you want to use it. So I'm not too concerned about that. Someone will create an issue, we'll point them to the line in the docs about it, and everyone else will see that as well.

I have plans to clean up Vite's logs in general, after that it should be possible to improve the warning message and tell them to add the entrypoint to `excludedEntrypoints`.

aklinker1:
Ignore the failing `Continuous Publish / build` check, that happens for new packages.

aklinker1 (approved):

Timeraa:
Everything looks fine to me!

PR #1045, opened 2024-10-04
perf: Downgrade `esbuild` so a single version is shared between sub-dependencies

aklinker1:
Should reduce the overall size of WXT's install by 22% (110mb &rarr; 86mb):

https://pkg-size.dev/https:%2F%2Fpkg.pr.new%2Fwxt@3c3eb77

I also added a test to make sure this it is valid.


aklinker1:
Released in [`v0.19.12`](https://github.com/wxt-dev/wxt/releases/tag/wxt-v0.19.12)

PR #1076, opened 2024-10-18
feat: Add support for `WXT_` environment variable prefix

aklinker1:


aklinker1:
Released in [`v0.19.12`](https://github.com/wxt-dev/wxt/releases/tag/wxt-v0.19.12)

aklinker1:
@hussion No, not unless you patch WXT with `patch-package`. Why would you need to remove this prefix? I don't see how this would cause harm

PR #1078, opened 2024-10-18
feat(env): Load env from `.env.[browser]` variants

sleekslush:
Adds ability to prioritize browser-specific environment settings when resolving configuration.

The `mode` and `browser` files will have highest precedence, followed by the rest as they exist today.
```
.env.[mode].[browser]
.env.[mode].[browser].local
```

### DIFF
--- a/docs/guide/essentials/config/environment-variables.md
+++ b/docs/guide/essentials/config/environment-variables.md
@@ -9,6 +9,8 @@ WXT supports [dotenv files the same way as Vite](https://vite.dev/guide/env-and-
 .env.local
 .env.[mode]
 .env.[mode].local
+.env.[browser]
+.env.[browser].local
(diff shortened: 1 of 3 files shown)


aklinker1:
Nice, do you think we also need the `.env.[browser]` and `.env.[browser].local` variants?

sleekslush:
> Nice, do you think we also need the `.env.[browser]` and `.env.[browser].local` variants?

I considered about doing that but thought it might be weird if someone did `--mode=chrome`. I don't think it necessarily is a problem, so I could go either way.

aklinker1:
I added them. I'll merge and do a release with this soon!

aklinker1:
Released in https://github.com/wxt-dev/wxt/releases/tag/wxt-v0.19.13

PR #1091, opened 2024-10-19
Add multiple icon support

ultimateshadsform:
I have tested it with the wxt demo and it seems to work as I expected.

I also tried with .jpg and it works it converts to a .png. (But with lower quality as you said).
But that's up to the user. Just use a .png from the start or a high quality one.

I also added support for placing image in public folder but if there is both in public and in assets it will prioritize the one from the assets folder.

Does everything look good? Anything needed to change?

Could you test it on your end too?


ultimateshadsform:
I think that is all.

aklinker1 (commented):
@Timeraa thoughts on this?

packages/auto-icons/src/index.ts
```diff
@@ -42,20 +85,28 @@ export default defineWxtModule<AutoIconsOptions>({
...
     wxt.hooks.hook('build:done', async (wxt, output) => {
-      const image = sharp(resolvedPath).png();
+      let image: sharp.Sharp;
+
+      if (extname(resolvedPath).toLowerCase() === '.png') {
+        // If the file is already PNG, just read it without conversion
+        image = sharp(resolvedPath).png();
+      } else {
+        // If it's not PNG, convert it
+        image = sharp(resolvedPath).toFormat('png').png();
```
1natsu172, line 95: As for the framework, if it's not png (transparent), it might be better to log a warning. Not a sure opinion, though.

ultimateshadsform: That's a good addition!

ultimateshadsform:
I'm not sure how to use this review thing. But it's in the latest commit down there.

packages/wxt-demo/wxt.config.ts
```diff
@@ -13,6 +13,9 @@ export default defineConfig({
       },
     ],
   },
+  autoIcons: {
+    grayscaleOnDevelopment: true,
+  },
```
Timeraa, line 18: Didn't I make `grayscaleOnDevelopment` true by default?

Timeraa:
> @Timeraa thoughts on this?

Everything looks fine, I do think that we should error on a non transparent image instead of converting it to prevent people from leaving it there because "It's just a warning"

ultimateshadsform:
> > @Timeraa thoughts on this?
> 
> Everything looks fine, only my comment above, I do think that we should error on a non transparent image instead of converting it to prevent people from leaving it there because "It's just a warning"

I don't know. Not every logo/image have transparency.

ultimateshadsform:
> Icons should generally be in PNG format, because PNG has the best support for transparency. They can, however, be in any raster format supported by Blink, including BMP, GIF, ICO, and JPEG.

https://developer.chrome.com/docs/extensions/reference/manifest/icons

Timeraa:
I still think we should disallow the use of non transparent formats as it's generally speaking a bad practice for extensions. @aklinker1?

ultimateshadsform:
> I still think we should disallow the use of non transparent formats as it's generally speaking a bad practice for extensions. @aklinker1?

I think think we should. because this is ONLY for Chrome. I don't think Firefox store enforces this rule.

https://developer.mozilla.org/en-US/docs/Web/Manifest/icons#specifications

They even put svg etc.

image

ultimateshadsform:
I think we should separate between Chrome and Firefox. 

We should have png as default but should at least not block it.

ultimateshadsform:
But I think you may be right about Chrome. They do say it's required for extension.

https://developer.chrome.com/docs/webstore/images#icons

But that is ONLY for Chrome.As far as I can see Firefox or edge addons doesn't say that they MUST be in in that format.

But yeah.

ultimateshadsform:
Closing this. I'm just gonna make my own version.

PR #1093, opened 2024-10-20
Prevent toggling of csp just to work with firefox development

windmillcode0:
instead of     content_security_policy:"script-src 'self'; object-src 'self'; img-src 'self' blob: data: unsafe-eval", just to only work with firefox can leave as     // content_security_policy: {
    //   "extension_pages": ...
    // }, and the framework will know what to do
annoying to go back and forth between firefox development and the rest


aklinker1:
This isn't the correct place to add this logic. The CSP util is a pure utility around parsing and manipulating CSP strings, it doesn't have any web extension or manifest logic in it.

Instead, we should update the types so you always declare the MV3 content_security_policy, and automatically downgrade it to the MV2 format

Here's where we do that for other manifest fields: https://github.com/wxt-dev/wxt/blob/main/packages%2Fwxt%2Fsrc%2Fcore%2Futils%2Fmanifest.ts#L118-L120

windmillcode0:
Greetings I have made the desired changes

aklinker1:
@windmillcode0 Sorry for taking so long to review this. I've refactored the code a bit, but can't push. Can you edit this PR to allow maintainer's to push to it?

windmillcode0:
Greetings what would you need me to do? I tried pulling but it indicated its already up to date there is a merge conflict and am not sure which branch you updated

aklinker1:
https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork#enabling-repository-maintainer-permissions-on-existing-pull-requests

windmillcode0:
I dont see that option here 
image

aklinker1:
Whelp. I'm gonna close this PR and open the one with my changes in it then. It has all your commits and you'll still be a contributor.

PR #1096, opened 2024-10-20
enhance(docs): comparison to real footnotes

Malix-Labs:
## Documentation

https://www.markdownguide.org/extended-syntax/#footnotes


aklinker1:
Alright, got it working. Patched the package to get rid of `id:subId` syntax when the footnote is used multiple times (`[3.1]` &rarr; `[3]`). Also converted from using numbers to using letters because the footnote plugin changes them so the first one seen is 1. That was confusing me, so I switched to letters for the references instead.

Screenshot 2024-10-22 at 8 21 45 AM

aklinker1 (approved):
Thanks for the PR! I was looking for a better way to do this, so glad you found this!

PR #1103, opened 2024-10-22
fix: Don't use `#private` member variables in `ContentScriptContext`

aklinker1:
Noticed in a few of my extensions I was getting an error "Couldn't access property 'signal' of undefined" in production builds. Was fixed by patching WXT to remove the #private members of the context class.

Every time I use classes in JS, I end up regretting it SMH.


aklinker1:
Released in https://github.com/wxt-dev/wxt/releases/tag/wxt-v0.19.13

PR #1156, opened 2024-11-11
fix: Throw when config file does not exist

kongmoumou:
Currently, if provide a non-existent config file(like typo or something else), it would fallback to default config and may throw some misleading err. Maybe we should handle this case and abort it explicitly.

## Before

image

## After

image


aklinker1 (approved):
Thanks! Will merge once checks pass

aklinker1:
Released in https://github.com/wxt-dev/wxt/releases/tag/wxt-v0.19.14

PR #1175, opened 2024-11-14
fix: Add extension ID to event used to invalidate `ContentScriptContext`

aklinker1:
Before, all WXT extensions shared the same event name, meaning if you had multiple installed they would invalidate each other.

Now, the event includes the extension ID, so it only invalidates itself.

### DIFF
--- a/packages/wxt/src/client/content-scripts/content-script-context.ts
+++ b/packages/wxt/src/client/content-scripts/content-script-context.ts
@@ -35,7 +35,9 @@ import { createLocationWatcher } from './location-watcher';
  * ```
  */
 export class ContentScriptContext implements AbortController {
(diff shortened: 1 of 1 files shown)


packages/wxt/src/client/content-scripts/content-script-context.ts
```diff
@@ -35,7 +35,9 @@ import { createLocationWatcher } from './location-watcher';
  * ```
  */
 export class ContentScriptContext implements AbortController {
-  private static SCRIPT_STARTED_MESSAGE_TYPE = 'wxt:content-script-started';
+  private static SCRIPT_STARTED_MESSAGE_TYPE = getUniqueEventName(
+    'wxt:content-script-started',
+  );
```
aklinker1, line 40: I thought I was already using this function for this event... but I was using it for `wxt:locationchange`, must have misremembered where it was used.

PR #1184, opened 2024-11-17
chore: Update Svelte template to Svelte 5

PatatoBit:
using `new App()` no longer works in Svelte 5. I updated the example template to Svelte 5. Though there is still a type error which can be fixed using this. But I just reckoned the code would've been too complicated.

```ts
const target = document.getElementById("app");
let app;
if (target) {
  app = mount(App, { target });
} else {
  throw new Error("Target element not found");
}
```

I found this after finding where it doesn't work. This is actually my first time using wxt so I took a while for it as I really want to use Svelte.


Timeraa (approved):
lgtm, I don't use Svelte so I'll let @aklinker1 merge this

PR #1197, opened 2024-11-22
chore(vue-template): remove references to Vetur and TypeScript Vue Plugin (Volar)

aiktb:
TypeScript Vue Plugin (Volar) has been deprecated and is no longer required for Vue 3 projects.
Vetur only works with Vue2 and has not been updated for 1 year.

There should be no need to mention them in templates.


aklinker1 (approved):
Thanks, this file was copied from Vite's template.

Issue #537, opened 2024-03-10
Mount and unmount with MutationObserver


hexpl0it:
For my extension I need to mount my app immediately after a precise div. However, this div is not immediately available when the page is loaded.

To do this I used MutationObserver: 

```javascript
import ReactDOM from "react-dom/client";
import Toolbar from "./components/Toolbar";
import { useEffect } from "react";

function waitForElm(selector) {
  return new Promise((resolve) => {
    if (document.querySelector(selector)) {
      return resolve(document.querySelector(selector));
    }
    const observer = new MutationObserver((mutations) => {
      if (document.querySelector(selector)) {
        observer.disconnect();
        resolve(document.querySelector(selector));
      }
    });
    observer.observe(document.body, {
      childList: true,
      subtree: true,
    });
  });
}

export default defineContentScript({
  main(ctx) {
    const ui = createIntegratedUi(ctx, {
      position: "inline",
      onMount: async (container) => {
        const elm = await waitForElm('[class^="ToolbarContainer__StyledHeader"]');
        const rootNode = document.createElement("div");
        elm.insertAdjacentElement("afterend", rootNode);
        const root = ReactDOM.createRoot(rootNode);
        root.render(<Toolbar />);
        return root;
      },
      onRemove: (root) => {
        root.unmount();
      },
    });

    ui.mount()
  },
});
``` 

However, this div can be removed following certain actions on the page. I would need to reassemble my component as soon as this div reappears. How can I do this?

aklinker1:
You'll want to use the awaited element as the `anchor` tag.

Something like this would work. (I typed this up in github comments, so it probably isn't valid JS, and I haven't tested it).

```ts
let anchor;
const ui  = createIntegratedUi(ctx, {
  position: "inline",
  anchor: () => anchor,
  onMount: async (container) => {
    const root = ReactDOM.createRoot(container);
    root.render(<Toolbar />);
    return root;
  },
  onRemove: (root) => {
    root.unmount();
  },
});
watchDomChanges(ctx, '[class^="ToolbarContainer__StyledHeader"]', {
  onAdd: (newAnchor) => {
    anchor = newAnchor;
    ui.mount();
  },
  onRemove: () => {
    ui.remove();
  },
});
```

```js
function watchDomChanges(ctx, selector, callbacks) {
  const addMountListener = () => {
    const observer = new MutationObserver((mutations) => {
      const el = document.querySelector(selector)
      if (el) {
        observer.disconnect();
[...]

PR #1210, opened 2024-11-26
feat: `autoMount` content script UIs

1natsu172:
refer: https://github.com/wxt-dev/wxt/issues/537

## implementation

Basically the style of the existing implementation, the `ui.remove` method is internally split. Because what autoMount callback and what the user calls are different in behavior.

## testing
Tested for all built-in UIs. 

The API spec makes it a floating-promise, so I'm hacking to get over the asynchronous processing. We must call `runMicrotasks` after every dom change and floating-promise process. Otherwise, only the vitest process would proceed and result in a false positive test.

## wxt-demo

Added new `automount.content`.  The gif I posted on Discord, with a delay in between for directional purposes.

## docs

It's pretty temporary, so feel free to fix it so that it's consistent with the other pages. The same is true for JSDoc, please correct any areas of concern.


1natsu172:
@aklinker1 PR is ready for review. But I'm sure your V1 milestone is a priority for you 😃

packages/wxt/src/client/content-scripts/ui/index.ts
```diff
@@ -27,16 +34,38 @@ export function createIntegratedUi<TMounted>(
...
+  let autoMountInstance: AutoMount | undefined = undefined;
   const mount = () => {
     applyPosition(wrapper, undefined, options);
     mountUi(wrapper, options);
     mounted = options.onMount?.(wrapper);
   };
+  const autoMount = (autoMountOptions?: AutoMountOptions) => {
+    if (autoMountInstance) {
+      throw Error('autoMount is already set.');
+    }
```
aklinker1, line 46: Should we throw an error here, or just return and do nothing?

packages/wxt/src/client/content-scripts/ui/types.ts
```diff
@@ -216,3 +220,24 @@ export interface ContentScriptAnchoredOptions {
    */
   append?: ContentScriptAppendMode | ((anchor: Element, ui: Element) => void);
 }
+
+/**
+ *
+ */
```
aklinker1, line 226: Delete or add a comment

packages/wxt/src/client/content-scripts/ui/types.ts
```diff
@@ -216,3 +220,24 @@ export interface ContentScriptAnchoredOptions {
...
+export type AutoMountOptions = {
+  /**
+   * Executes only a once mount and a once unmount.
```
aklinker1, line 229:
```suggestion
   * When true, only mount and unmount a UI once.
```

packages/wxt/src/client/content-scripts/ui/types.ts
```diff
@@ -216,3 +220,24 @@ export interface ContentScriptAnchoredOptions {
...
+  once?: boolean;
+  /**
+   * The callback triggered when `StopAutoMount` is called.
+   */
+  onStop?: () => void;
+};
+export type StopAutoMount = () => void;
+export interface AutoMount {
+  /**
+   * The observing by AutoMount stops, but the UI is not removed.
```
aklinker1, line 240:
```suggestion
   * Stop watching the anchor element for changes, but keep the UI mounted.
```

packages/wxt/src/client/content-scripts/ui/types.ts
```diff
@@ -41,6 +41,10 @@ export interface ContentScriptUi<TMounted> {
    * Function that mounts or remounts the UI on the page.
    */
   mount: () => void;
+  /**
+   * Function that automatically mounts or remounts the UI in response to dynamic anchor.
```
aklinker1, line 45:
```suggestion
   * Call `ui.autoMount()` to automatically mount and remove the UI as the anchor is dynamically added/removed by the webpage.
```

docs/guide/essentials/content-scripts.md
```diff
@@ -571,6 +571,44 @@ For MV3, `injectScript` is synchronous and the injected script will be evaluated
...
+    // Call autoMount to observe anchor element for add/remove.
+    // You can stop it by calling the returned function.
+    const stopAutoMount = ui.autoMount();
+  },
+});
+```
+
+:::tip About the `once` option
+The `once` option executes a once mount and a once unmount.
+:::
```
aklinker1, line 604: I think we can remove this tip, we link to the API reference below, where this same description is present, and we don't mention the `once` option on this page.

packages/wxt/src/client/content-scripts/ui/index.ts
```diff
@@ -67,16 +97,38 @@ export function createIframeUi<TMounted>(
...
+    autoMountInstance = autoMountUi(
+      { mount, unmount, stopAutoMount },
+      {
+        ...options,
+        ...autoMountOptions,
+      },
+    );
+    return stopAutoMount;
+  };
```
aklinker1, line 118: There's a bit of code duplication here, doing the same thing for each of the three UI functions... Would it be better if `autoMount` was a separate utility function exported from `wxt/client` rather than an instance function?

```diff
- ui.autoMount()
+ autoMount(ui)
```

That would make it easier to reduce code duplication... Not sure if that'd be better than/easier to use or see compared to what you're currently doing, just throwing out an idea.

_Edit: the anchor is private to the instance, so this probably wouldn't be possible without exposing it... Hmm, is there a better way to reduce code duplication? Maybe a function like this?_

```ts
interface AutoMountable {
  mount(): void;
  remove(): void;
}

interface AutoMount {
   autoMount(...): StopAutoMount
}

function addAutoMount<T extends AutoMountable>(
  anchor: ...,
  ui: T,
): T & AutoMount {
  ...
}
```

_These interface names probably conflict with what you've already defined, just using them as an example. They might already exist in your branch too, can't remember._

aklinker1 (commented):
Couple of recommendations, but overall, code looks good and seems to work when I tested out the demo app.

packages/wxt/src/client/content-scripts/ui/index.ts
1natsu172: Yeah...if we needed a fourth type of built-in UI, I think it would be tough. Let's see if your helper function pattern will work. But, it is necessary to let the user set the anchor twice.... 🤔

packages/wxt/src/client/content-scripts/ui/index.ts
1natsu172: @aklinker1 Hmm, but If don't throw it, the return value to be `StopAutoMount | undefined`. Or maybe warn logger is better.

packages/wxt/src/client/content-scripts/ui/index.ts
aklinker1: Yeah, I would rather not have users pass the anchor twice.

packages/wxt/src/client/content-scripts/ui/index.ts
aklinker1: Yeah, I like this better.

packages/wxt/src/client/content-scripts/ui/types.ts
```diff
@@ -216,3 +205,43 @@ export interface ContentScriptAnchoredOptions {
...
+   * Custom data returned from the `options.mount` function.
+   */
+}
+
+export interface MountFunctions extends BaseMountFunctions {
+  /**
+   * Call `ui.autoMount()` to automatically mount and remove the UI as the anchor is dynamically added/removed by the webpage.
+   */
+  autoMount: (options?: AutoMountOptions) => StopAutoMount;
+}
```
1natsu172, line 229: Abstraction of the user-side interface to `MountFunctions`.

packages/wxt/src/client/content-scripts/ui/index.ts
```diff
@@ -256,6 +287,109 @@ function mountUi(
...
+
+  return {
+    mount,
+    remove,
+    autoMount,
+  };
+}
```
1natsu172, line 332: I made the factory function because I thought the abstraction was the mount, not the UI.

aklinker1 (commented):
When I clicked the button to stop auto mounting, I expected it to remove the UI. Was that intentional? In my mind, the UI should be shown when auto mount is running, and removed when it's not.

<details>
<summary>Video</summary>

https://github.com/user-attachments/assets/7ecf74d6-8d23-4724-b1f3-17bdbf2dba4e

</details>

Everything else looks good, and it works well. ~~Only other comment is where the 1 second delay is coming from? Aren't MutationObserver's nearly instant?~~ 🤦 missed the timeout you added lol

aklinker1:
> Only other comment is where the 1 second delay is coming from? Aren't MutationObserver's nearly instant? 🤦 missed the timeout you added lol

Did the timeout cause https://github.com/wxt-dev/wxt/pull/1219?

aklinker1 (approved):

1natsu172:
@aklinker1 

> When I clicked the button to stop auto mounting, I expected it to remove the UI. Was that intentional? In my mind, the UI should be shown when auto mount is running, and removed when it's not.

This is intentional as it is implemented now(https://github.com/1natsu172/wxt/blob/001e42fb1fc602ae3bd7401b36b2d5726b80dcbb/packages/wxt/src/client/content-scripts/ui/index.ts#L296-L299). `stopAutoMount` stops the autoMount mechanism. It does not control the UI.

Just in case, the user can remove manually it using `onStop`.

```ts
const stopAutoMount = autoMountUi.autoMount({
      onStop: () => {
        autoMountUi.remove()
        console.log('Auto mount stopped.');
      },
    });
```

But is it preferable if the behavior of `stopAutoMount` is _stopAutoMountAndRemove_? 🤔 

> Did the timeout cause https://github.com/wxt-dev/wxt/pull/1219? In that case, you should return the timeout and clear it in onRemove. But I removed the timeout from the demo extension.

It was just put in for visual clarity. No problem to remove it.

aklinker1:
> This is intentional as it is implemented now(https://github.com/1natsu172/wxt/blob/001e42fb1fc602ae3bd7401b36b2d5726b80dcbb/packages/wxt/src/client/content-scripts/ui/index.ts#L296-L299). stopAutoMount stops the autoMount mechanism. It does not control the UI.

Hmm, I don't see a use-case for keeping the UI mounted when the anchor is removed... I think we should change it so the UI gets removed.

aklinker1:
Also, seems like windows tests are broken?

aklinker1:
Hmm, the stop function returned from `ui.autoMount` won't really have a purpose anymore after this change, will it? Because it's the same thing as `ui.remove`? So is the solution just to remove the return value and tell people to use `remove` instead?

aklinker1:
Actually, maybe there is value in keeping the stop function. Maybe we should just tell people to call remove in the docs, and optionally mention that they can stop watching but keep the UI mounted if they want.

1natsu172:
Hmm but now I feel it is better to stipulate the operation strictly. In the original Issue we had a conversation about covering only one scenario, and the `once` option also includes `remove`. It is less confusing to make the operation consistent.

1natsu172:
If we are confused, the users will be more confused.

1natsu172:
> Also, seems like windows tests are broken?

It was an error that the document object was not found in the background, but there may be a flaky test. It passes now. 🤔

aklinker1 (commented):
OK, then let's remove the return value since `remove` and `stop` are doing the same thing.

aklinker1:
Hmmm background test is still failing. Why would the auto mount library be getting ran for that test?

1natsu172:
Hmm, I thought the cleanup was inadequate, but I guess not. 

Actually, regardless of this PR (even in the `main` branch), getting `Content script “test” context invalidated` and Happy-dom errors in other tests. It might be related. 🤔 But why Windows?

Run `pnpm run test ui`

```bash
stderr | src/client/content-scripts/ui/__tests__/index.test.ts > Content Script UIs > mounted value > iframe > should set the mounted value based on the onMounted return value
DOMException [NotSupportedError]: Failed to fetch from "chrome-extension://test-extension-id/": URL scheme "chrome-extension" is not supported.
    at Function.validateSchema (file:///Users/1natsu/ghq/github.com/wxt-dev/wxt/node_modules/.pnpm/happy-dom@15.11.4/node_modules/happy-dom/src/fetch/utilities/FetchRequestValidationUtility.ts:120:10)
```

```bash
stdout | src/client/content-scripts/ui/__tests__/index.test.ts
[wxt] Content script "test" context invalidated
[wxt] Content script "test" context invalidated
[wxt] Content script "test" context invalidated
.
.
```

1natsu172:
Huh, if I look closely, it's a `wxt-demo` test file. And I reproduced it on my local macOS by running `pnpm test` with monorepo root. I'll investigate.

aklinker1:
Ok, thanks for investigating. I'll check it out after work today

aklinker1:
I see, simple problem. `useAppConfig` is auto-imported from `wxt/client`, which includes the UI functions, so your library gets loaded, and it accesses `document` as soon as it's imported... I suppose `useAppConfig` should be imported from `wxt/sandbox` instead. Hmm... gonna have to wait for `0.20.0` for this, that's definitely a breaking change.

@1natsu172 this is caused because in `@1natsu/wait-element`, you're accessing `document` as a side effect. To make it safe across different runtimes, I'd recommend using `globalThis.document` instead (which evaluates to `undefined` in environments without the global instead of throwing an error), or not accessing it directly until an API is called. That said, this is WXT's fault, so I'll come up with a fix for it on my side. Just a compatibility recommendation.

1natsu172:
@aklinker1 

> which includes the UI functions, so your library gets loaded, and it accesses document as soon as it's imported

I see...thank you for the investigation.

> To make it safe across different runtimes, I'd recommend using `globalThis.document` instead

I didn't expect it to be referenced outside of the browser environment, but you're right about cases like this one. I'll fix soon my library and publish it too.

1natsu172:
Okay, I added a commit updating upstream(my library). The test passes on my local.

aklinker1:
Awesome, I've got the WXT change working, but there's a bit more cleanup to do. But we can merge this PR!

s-bessing:
@aklinker1 @1natsu172 is it possible to add the element multiple times? If the anchor is present more the once then also the elements should be mounted to every element on the page. Is this possible?

PR #1206, opened 2024-11-25
Provide a way to reopen the browser like vite.

nishu-murmu:
I was looking around the code. I implemented this.
We can map more features for certain keystrokes.

Do let me know your thoughts?
This PR fixes #348

PR #1211, opened 2024-11-26
feat: Keyboard shortcut to reopen the browser without restarting the dev command

nishu-murmu:
Thanks for the insight @aklinker1, still getting used to the code structure and APIs.
First time actually contributing to open source project. 😓
Will look into it.

But I was also suggesting that, we can map few keys for various server related functionality.

Like the reference you gave
https://github.com/vitejs/vite/blob/bf1e9c2fd7b05f84d05e59f72b3fc26ca22807bb/packages/vite/src/node/shortcuts.ts

Had various key mapped to certain functionality.

We can do so?

I've added the code for removing and registering the listeners.
Can you check if it correct implementation?

**Edit:**
Sorry, Accidently closed the actual PR.


aklinker1:
No worries, if you push code to the same branch, you don't need to open a new PR with every change. We'll just use this one now instead of #1206. I'll to go this later today, woke up and going to work now lol.

packages/wxt/src/core/create-server.ts
```diff
@@ -107,10 +108,15 @@ async function createServerInternal(): Promise<WxtDevServer> {
...
       // Listen for file changes and reload different parts of the extension accordingly
       const reloadOnChange = createFileReloader(server);
+      const keyboardsShortCuts = createKeyBoardShortCuts(server);
       server.watcher.on('all', reloadOnChange);
+      keyboardsShortCuts.start();
     },
+
     async stop() {
       wasStopped = true;
+      const keyboardsShortCuts = createKeyBoardShortCuts(server);
```
Timeraa, line 118: Shouldn't it be Shortcut and Keyboard? Because I don't think we're Short Cutting something lol

sh.exe.stackdump
aklinker1, line 1: Not sure where this came from, but let's delete it.

packages/wxt/src/types.ts
```diff
@@ -535,6 +535,11 @@ export interface WxtDevServer
   restartBrowser: () => void;
 }
 
+export interface KeyboardShortcutWatcher {
+  start(): void;
+  stop(): void;
+}
+
```
aklinker1, line 542: This file is for public types, so lets put the interface inside the new `src/core/keyboard-shortcuts.ts` file, next to `createKeyboardShortcuts`.

aklinker1 (changes requested):
Couple of recommendations:

1. Let's move `createKeyboardShortcuts` into it's own file for better organization, `src/core/keyboard-shortcuts.ts`
2. Instead of using a global variable, we should put the `isWatching` variable inside the `createKeyboardShortcuts` function so it's not global, but state for a single instance of the `keyboardShortcuts` object.
3. Related to 2, we should only create 1 keyboard shortcut object outside the `server` object. Then you can just call `.start()` and `.stop()` inside the functions. You mentioned you weren't sure where to put it so it has access to the server? I'd recommend putting it here: https://github.com/nishu-murmu/wxt/blob/88d15a2b715c93e89b09923b70875bf46dcb3bd4/packages/wxt/src/core/create-server.ts#L153

packages/wxt/src/types.ts
nishu-murmu: Thanks for the clarifications and recommendations.
I'll quickly make those changes.

packages/wxt/src/core/create-server.ts
nishu-murmu: If it is confusing I will change it.
What name do you suggest?

aklinker1:
```diff
- createKeyBoardShortCuts
+ createKeyboardShortcuts
```

"keyboard" and "shortcuts" are each one word, so you shouldn't capitalize the "b" and "c".

aklinker1 (changes requested):
Alright, code looks good, other than the capitalization @Timeraa mentioned. 

Tested it out on Mac, but I can no longer use `ctrl+C` to stop the server. It does nothing when I press those keys, when it should stop the process.

I did some snooping around Vite's code, and they don't use raw input mode:

https://github.com/vitejs/vite/blob/51a5569e66bd7f0de79ac14b9e902d1382ccd0aa/packages/vite/src/node/shortcuts.ts#L90-L92

Should we also use `readline.createInterface`?

nishu-murmu:
> Alright, code looks good, other than the capitalization @Timeraa mentioned.
> 
> Tested it out on Mac, but I can no longer use `ctrl+C` to stop the server. It does nothing when I press those keys, when it should stop the process. Reloading works really well though!
> 
> I did some snooping around Vite's code, and they don't use raw input mode:
> 
> https://github.com/vitejs/vite/blob/51a5569e66bd7f0de79ac14b9e902d1382ccd0aa/packages/vite/src/node/shortcuts.ts#L90-L92
> 
> Should we also use `readline.createInterface`?

The first thing I did I saw the vite's code.

Initially I tried that code itself
But we have to additionally press enter for the keypress to work

Then I tried the raw input mode which was working.

Do you any idea regarding this?

aklinker1:
> The first thing I did I saw the vite's code.
> 
> Initially I tried that code itself
> But we have to additionally press enter for the keypress to work
> 
> Then I tried the raw input mode which was working.
> 
> Do you any idea regarding this?

My main concern here is there's more than just `ctrl+C`... We need to continue handling all other signals as well, `ctrl+X`, `ctrl+~`, etc. I don't think adding if-statements is the right way to go... Vitest's keyboard shortcuts don't require pressing enter. Seems like they use if-statements, but it seems pretty complex...

https://github.com/vitest-dev/vitest/blob/1fa16f941a651c1af0c24124ec4533d432002236/packages/vitest/src/node/stdin.ts#L52-L75

Is there a library we can use for this?

aklinker1:
> Is there a library we can use for this?

Doesn't seem like there is. I think I would prefer to use Vite's appraoch and require pressing enter. Their help message makes it really easy to see that you need to press enter, which I think is fine. We should probably add a log telling people to "Type *o + enter* to open the browser".

nishu-murmu:
@aklinker1 I've reverted back to `readline` code, removing the raw input mode
And I've also added an info log for this functionality.

packages/wxt/src/core/keyboard-shortcuts.ts
```diff
@@ -0,0 +1,50 @@
...
+  };
+
+  return {
+    start() {
+      if (isWatching) return;
+
+      rl = readline.createInterface({
+        input: process.stdin,
+        output: process.stdout,
+        terminal: false, // Don't intercept ctrl+C, ctrl+Z, etc
```
aklinker1, line 31: Specifically, I had to add this line, `terminal: false`, for it to work on Mac. Without it, the dev server only stopped after pressing ctrl+C twice. With it, I only have to press ctrl+C once, like before.

aklinker1 (approved):
Awesome, thanks! I made some small changes, but this is working well now!

What operating system do you use @nishu-murmu ? Can you test my changes to make sure it works for you too?

aklinker1:
@Timeraa can you test this out too?

Timeraa:
> @Timeraa can you test this out too?
> 
> 
> 
> > If you have the github CLI installed, you can checkout PRs from forks with `gh pr checkout 1211`. No idea how to do it without `gh`. I guess clone the fork and test there?

Will do

packages/wxt/src/core/runners/web-ext.ts
```diff
@@ -81,6 +82,7 @@ export function createWebExtRunner(): ExtensionRunner {
 
       const duration = Date.now() - startTime;
       wxt.logger.success(`Opened browser in ${formatDuration(duration)}`);
+      wxt.logger.info(`To reopen the browser, press ${pc.bold('o + enter')}`);
```
Timeraa, line 85:
```suggestion
      wxt.logger.info(`Reopen browser using ${pc.bold('o')} + ${pc.bold('ENTER')}`);
```

Timeraa: Tested, works fine for me on mac with CTRL + C too, I do have this one change just to make the text look a bit better. @aklinker1 thoughts?

Timeraa (changes requested):

packages/wxt/src/core/runners/web-ext.ts
aklinker1: Looks good, is that the same wording and casing Vite uses? We should just match what they do... Away from my computer, can't check.

Timeraa: image

~~Well if you want it to look like vite it has a bit of work to go jk~~

Timeraa: image
Ah, that's what it looks like. I think this should be fine for now, once we add more tools and stuff we can make the ui look better? Should I change my suggestion and we stick with it @aklinker1 or do we keep it like I suggested it for now?

Timeraa (approved):

Issue #1214, opened 2024-11-26
Vite 6


aklinker1:
### Feature Request

Double check that it works without issues, and look into using the new environment API to improve build speed.

#### Is your feature request related to a bug?

N/A

### What are the alternatives?

N/A

### Additional context

https://vite.dev/blog/announcing-vite6.html

aklinker1:
Wrote a simple test to see how to use the environment APIs to build multiple bundles at once. The API is much nicer to use, but performance wise, `vite.createBuilder(...).buildApp()` negligibly faster than just running `vite.build(...)` multiple times with different config. Hard to test on a real extension without actually implementing it in WXT... so we'll have to do that to benchmark it against a real extension with lots of dependencies.

Did not profile memory usage.

eMerzh:
Hello,

not sure you had that in your radar, but we were prompted with vite6 dep update, and it seems that the vite (https://wxt.dev/guide/essentials/config/vite.html) method to define the config isn't really type compatible anymore. 

properties like resolve alias & other 
🤔

aklinker1:
@eMerzh You probably have multiple versions of Vite installed. Type errors like this are usually a sign on if. Just released a new version of WXT that adds support for Vite 6: [`v0.19.17`](https://github.com/wxt-dev/wxt/releases/tag/wxt-v0.19.17)

Delete your lockfile or run `pnpm i vite@latest && pnpm rm vite` to update vite. Then run `pnpm why vite` again to make sure there's only one version installed in the project. If you use a different package manager... there are similar commands, but I don't know them off the top of my head.

remusris:
A vite template for using React Compiler would be a great addition or at least one example showing how to do it.

aklinker1:
Huh, does the regular react vite plugin not include the compiler already?

remusris:
You have to turn it on and when I tried turning it on it was giving me errors

![image](https://github.com/user-attachments/assets/25b3a207-e729-442b-bb0d-89ff155d68d6)

https://react.dev/learn/react-compiler

below was my attempt of integrating it

```typescript
import { defineConfig } from 'wxt';
import { defineRunnerConfig } from 'wxt';
import { defineWxtModule } from 'wxt/modules';

export interface CustomCommandOptions {
  // Add any options your command needs
  commandName?: string;
}

// See https://wxt.dev/api/config.html
export default defineConfig({
  extensionApi: 'chrome',
[...]

PR #1215, opened 2024-11-26
fix: Update `vite` dependency range to support v6

aklinker1:
Apart of #1214, everything seems in order. Vanilla template works in dev mode and production builds with no errors or warnings.

Plugins haven't updated their peer dependencies yet, so it's hard to test on full projects. Will let people open issues if it doesn't work for them.


aklinker1:
Released in [`v0.19.17`](https://github.com/wxt-dev/wxt/releases/tag/wxt-v0.19.17)

PR #1219, opened 2024-11-27
fix: Correctly remove child elements with integrated UI remove

### LINKED ISSUE #1218: Integrated UI does not remove appended child elements
### Describe the bug

I noticed this when debugging the #1210  demo. Removing the wrapper does not GC the child elements. Therefore, when the wrapper is mounted again, the number of child elements will continue to increase.

**I will submit PR for fix.**

https://github.com/user-attachments/assets/f2b5c594-1e61-416d-8704-8cf4b7be8368

### Reproduction

[`c469da8` (#1210)](https://github.com/wxt-dev/wxt/pull/1210/commits/c469da8d61019b1cb47d108db939fc6850a3dcde)
The current demo of this commit will grow in elements.

### Steps to reproduce

_No response_

### System Info

```shell
System:
    OS: macOS 15.1.1
    CPU: (14) arm64 Apple M4 Pro
    Memory: 2.19 GB / 64.00 GB
    Shell: 3.7.1 - /opt/homebrew/bin/fish
  Binaries:
    Node: 22.11.0 - ~/.local/share/mise/installs/node/22.11.0/bin/node
    npm: 10.9.0 - ~/.local/share/mise/installs/node/22.11.0/bin/npm
    pnpm: 9.13.0 - ~/.local/share/mise/installs/node/22.11.0/bin/pnpm
    bun: 1.1.34 - ~/.local/share/mise/installs/bun/1.1.34/bin/bun
    Watchman: 2024.11.11.00 - /opt/homebrew/bin/watchman
  Browsers:
    Brave Browser: 130.1.71.123
    Chrome: 131.0.6778.86
    Edge: 131.0.2903.70
    Safari: 18.1.1
  npmPackages:
    wxt: workspace:* => 0.19.16
```

### Used Package Manager

pnpm

### Validations

1natsu172:
fix #1218

Empty the child elements when removing. ref: https://developer.mozilla.org/en-US/docs/Web/API/Element/replaceChildren

### DIFF
--- a/packages/wxt/src/client/content-scripts/ui/index.ts
+++ b/packages/wxt/src/client/content-scripts/ui/index.ts
@@ -34,6 +34,7 @@ export function createIntegratedUi<TMounted>(
   };
   const remove = () => {
     options.onRemove?.(mounted);
+    wrapper.replaceChildren();
     wrapper.remove();
     mounted = undefined;
   };


aklinker1:
@1natsu172 Do we also need to do the same for the other 2 functions? They just call `remove()` as well.

1natsu172:
@aklinker1 It doesn't happen on the iframe side, shadow-ui already has an implementation that removes the children (https://github.com/wxt-dev/wxt/blob/aab4244d842702872a42c1da18005fc8fa88f4cd/packages/wxt/src/client/content-scripts/ui/index.ts#L142) .

1natsu172:
I haven't debugged it yet, but it might be because iframe don't append `options.onMount`.

https://github.com/wxt-dev/wxt/blob/51c7ee5fa82caead63f3c6ff2d7ab069abfef7d6/packages/wxt-demo/src/entrypoints/iframe.content.ts#L5-L9

aklinker1 (approved):
Alright, cool, we'll get this merged!

Issue #1202, opened 2024-11-23
Allow enabling `dev.reloadCommand` when more than 4 shortcut keys are registered


aiktb:
### Feature Request

Currently, WXT auto disable `dev.reloadCommand` when the extension registers more than 4 shortcut keys and there is no way to force it to be enabled.

This feature is useful and should be allowed if developers are aware of the risk of possible shortcut key conflicts.

Possible API for `wxt.config.ts`:
Ignore warnings to enable this feature when `dev.reloadCommand` is `true`, or change it to object type with a `forceEnable` boolean key.

https://github.com/wxt-dev/wxt/blob/5a3790a7d3595b7b44cf1824c569b379485536dd/packages/wxt/src/core/utils/manifest.ts#L85-L88

aiktb:
> This won't work. If you're registering more than 4 commands in the wxt.config.ts. The extension's manifest itself won't load. You can view the screenshot
> 
> I'm trying to load an extension, pure vanilla extension config, not using this framework. Still it is throwing this error. [screenshot](https://img.enacton.com/ShareX/2024/1128/brave_IDF3z4LVIX.png)

https://stackoverflow.com/a/61057647/17635343
I've seen extensions with more than ten commands, and this restriction only applies to commands with a "suggested_key".

aiktb:
So this is actually a WXT bug?
WXT should calculate the number of keys in the following way:
```js
const suggestedKeyCount = Object.values(commands).filter(command => command.suggested_key).length
if(suggestedKeyCount >= 4){
  warnings.push([ 
     "Extension already has 4 registered suggested key commands, WXT's reload command is disabled", 
   ]); 
}
```

I'll open a PR.

nishu-murmu:
> So this is actually a WXT bug? WXT should calculate the number of keys in the following way:
> 
> ```js
> const suggestedKeyCount = Object.values(commands).filter(command => command.suggested_key).length
> if(suggestedKeyCount >= 4){
>   warnings.push([ 
>      "Extension already has 4 registered suggested key commands, WXT's reload command is disabled", 
>    ]); 
> }
> ```
> 
> I'll open a PR. (Tomorrow)

Yes, just debugged it.
This is correct.

aklinker1:
Released in https://github.com/wxt-dev/wxt/releases/tag/wxt-v0.19.18

PR #1226, opened 2024-11-29
fix: correct `dev.reloadCommand` restriction to consider only commands with suggested keys

aiktb:
Closed https://github.com/wxt-dev/wxt/issues/1202

This limitation only exists in Chrome. I tried to load 10 suggested keys in Firefox MV2 and no error occurred.


aklinker1 (approved):
Awesome, thanks for the PR!

PR #1229, opened 2024-12-01
docs: Changed "Chrome Extension" to "web extension"

takuyahara:
Generalized wording as WXT is not only for Chrome Extension.


docs/.vitepress/config.ts
```diff
@@ -18,7 +18,7 @@ import { version as storageVersion } from '../../packages/storage/package.json';
 const title = 'Next-gen Web Extension Framework';
 const titleSuffix = ' – WXT';
 const description =
-  "WXT provides the best developer experience, making it quick, easy, and fun to develop chrome extensions for all browsers. With built-in utilities for building, zipping, and publishing your extension, it's easy to get started.";
+  "WXT provides the best developer experience, making it quick, easy, and fun to develop browser extensions for all browsers. With built-in utilities for building, zipping, and publishing your extension, it's easy to get started.";
```
aklinker1, line 21:
```suggestion
  "WXT provides the best developer experience, making it quick, easy, and fun to develop browser extensions. With built-in utilities for building, zipping, and publishing your extension, it's easy to get started.";
```

aklinker1 (commented):
Should it be "browser extension" or "web extension"? 

Here's Mozilla's docs where it refers to them as "browser extensions", but the URL uses "web extension" and the API standard is called "web extension API" and even that page uses both...

https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/What_are_WebExtensions
https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions

This is something I've struggled with for a while, trying to figure out the right generic  term... Thoughts?

takuyahara:
I propose "browser extension" because it's more general and common.

* https://trends.google.com/trends/explore?q=%22browser%20extension%22,%22web%20extension%22,%22browser%20extensions%22,%22web%20extensions%22&hl=en

"Web extension API" seems like Firefox-specific term and Chrome calls "Chrome Extensions API".

* https://developer.chrome.com/docs/extensions/reference/api

Note that I intentionally kept the heading "Next-gen Web Extension Framework" in index page as is, because I guessed WXT stands for "Web eXTension" and then it shouldn't change for branding.

takuyahara:
P.S. I found "web extension" is widely used in docs, my PR might better to use it at this moment as it's not a good idea to change "Chrome Extension" to "web extension" and change "web extension" to something else entirely at the same time. The latter one can be done in the future.

aklinker1:
https://www.w3.org/community/webextensions/

The API standard is called the "Web Extension API". It's not specific to Firefox, it is, as far as I can tell, the general term accepted by both Chrome and Firefox, who are both apart of the community group.

Oddly enough, the group is described as "browser web extension".

> We will specify a model, permissions, and a common core of APIs for web browser extensions. By specifying the APIs, functionality, and permissions of WebExtensions, we can make it even easier for extension developers to enhance end user experience, while moving them towards APIs that improve performance and prevent abuse.

I would prefer to have all 3 variations on the homepage for SEO, then standardize to "web extension" everywhere else.

takuyahara:
I agree to have variations then should "web extension" be used in this PR?

aklinker1:
Yes

takuyahara:
Done

.gitpod.yml
Timeraa, line 1: Don't think you meant to commit this

takuyahara: True, thank you for pointing out 🙏

aklinker1 (approved):
Thanks! Will get this merged once checks pass.

PR #1238, opened 2024-12-02
fix: Exclude skipped entrypoints from Firefox sources zip

### LINKED ISSUE #574: filtered out excluded entryPoint files when pack into firefox source zip
### filtered out excluded entryPoint files when pack into firefox source zip

for example, When define content script with `exclude['firefox']`, I don't want this content script to show within the zip file

"N/A"

alternatives as I think of, maintain another git branch and remove those files shall not show within zip file at the end.

### Additional context

use wxt version 0.17.7

nishu-murmu:
@aklinker1 I've applied the logic as you mentioned, passing a boolean parameter to the `findEntrypoints` function.
Additionally while filtering `options.exclude` I had to use relative path, cause it wasn't working in `minimatch` method.

I've tested this on Windows.

This fixes #574

### DIFF
--- a/packages/wxt/src/core/utils/building/find-entrypoints.ts
+++ b/packages/wxt/src/core/utils/building/find-entrypoints.ts
@@ -156,24 +156,9 @@ export async function findEntrypoints(
   }
 
   if (allowAllEntryPoints) {
-    // wxt.config.zip.zipSources = true;
-    wxt.config.zip.excludeSources = entrypoints
-      .map((entry: Entrypoint) => {
-        if (entry.options?.include?.length && entry.options?.exclude?.length) {
-          wxt.logger.warn(
-            `The ${entry.name} entrypoint lists both include and exclude, but only one can be used per entrypoint. Entrypoint ignored.`,
-          );
-          return '';
-        }
-        if (entry.options.exclude?.includes(wxt.config.browser)) {
-          return entry.inputPath;
-        }
-        if (skippedEntrypointNames.includes(entry.name)) {
-          return entry.inputPath;
-        }
-        return '';
-      })
-      .filter(Boolean);
+    wxt.config.zip.excludeSources = entrypoints.map((entry: Entrypoint) => {
+      return entry.inputPath;
+    });
   }
   const targetEntrypoints = entrypoints.filter((entry) => {
     const { include, exclude } = entry.options;
--- a/packages/wxt/src/core/utils/building/find-entrypoints.ts
+++ b/packages/wxt/src/core/utils/building/find-entrypoints.ts
@@ -31,9 +31,12 @@ import { wxt } from '../../wxt';
 import { createExtensionEnvironment } from '../environments';
 
 /**
- * Return entrypoints and their configuration by looking through the project's files.
+ * Return entrypoints and their configuration by looking through the project's files. *//**
+ * @param {string} [allowAllEntryPoints] - If true, all entrypoints are returned.
  */
-export async function findEntrypoints(): Promise<Entrypoint[]> {
+export async function findEntrypoints(
+  allowAllEntryPoints?: boolean,
+): Promise<Entrypoint[]> {
   // Make sure required TSConfig file exists to load dependencies
   await fs.mkdir(wxt.config.wxtDir, { recursive: true });
   try {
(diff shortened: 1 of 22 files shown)


packages/wxt/src/core/utils/building/find-entrypoints.ts
```diff
@@ -151,6 +154,12 @@ export async function findEntrypoints(): Promise<Entrypoint[]> {
         .join('\n')}`,
     );
   }
+
+  if (allowAllEntryPoints) {
+    wxt.config.zip.excludeSources = entrypoints.map((entry: Entrypoint) => {
+      return entry.inputPath;
+    });
+  }
```
aklinker1, line 162: Here, two changes:

1. You should only be adding excluded entrypoints to the `excludeSources`, but you're adding all entrypoints. See the filter right below this.
2. Instead of overwriting the `excludeSources` entirely, you need to merge the existing array with the skipped entrypoints. Note that this function is called multiple times, so you should use a `Set` to do this, not `Array.push`, which is why I used the term "merge".

packages/wxt/src/core/utils/building/internal-build.ts
```diff
@@ -28,8 +28,11 @@ import { isCI } from 'ci-info';
  * 2. Executes the rebuild function with a blank previous output so everything is built (see
  *    `rebuild` for more details)
  * 3. Prints the summary
+ * @param {string} [allowAllEntryPoints] - If true, all entrypoints are returned.
  */
-export async function internalBuild(): Promise<BuildOutput> {
+export async function internalBuild(
+  allowAllEntryPoints?: boolean,
+): Promise<BuildOutput> {
```
aklinker1, line 35: Undo this whole file. Instead of modifying the existing build functions, you should just call `findEntrypoints` inside `zip.ts` before we actually generate the ZIP.

Probably before this line:
https://github.com/wxt-dev/wxt/blob/4ce178c592ec106955697a4c4bc32cc0f24da65e/packages/wxt/src/core/zip.ts#L58

aklinker1 (changes requested):
Thanks for the PR, but as it is now, this will not work. See comments below.

aklinker1:
Actually, after looking at `findEntrypoints` a bit more, we should change it to return all entrypoints all the time. However, we need to make sure `skipped` is set properly based on if the entrypoint is excluded. Then, inside `groupEntrypoints`, we should only add entrypoints that aren't skipped to the groups to build.

That way, you can just call `findEntrypoints` inside `zip.ts` and update `excludeSources` accordingly, also in that file. I don't want to go modifying config in random places, I'd rather put the custom logic for excluding sources from being zipped right next to the code where we zip sources.

If we put it in the `find-entrypoints.ts` file like it is now, it's going to be a confusing side-effect that's hard to find in the future.

aklinker1:
@nishu-murmu I created a PR into your fork with my suggested changes mentioned in the previous comment. I realized it probably didn't make much sense to you, so I did it myself. 

https://github.com/nishu-murmu/wxt/pull/1

nishu-murmu:
I've refactored the code.
Now it seems to be working.

aklinker1:
@nishu-murmu Could you add an e2e test for this in `packages/wxt/e2e/tests/zip.test.ts`?

See this test for reference: https://github.com/wxt-dev/wxt/blob/0d796b63c22e41f347ba8bfe0c88f73b40107522/packages/wxt/e2e/tests/zip.test.ts#L151-L182

nishu-murmu:
> @nishu-murmu ~I created a PR into your fork with my suggested changes mentioned in the previous comment. I realized it probably didn't make much sense to you, so I did it myself.~
> 
> ~[nishu-murmu#1](https://github.com/nishu-murmu/wxt/pull/1)~
> 
> _Edit: I decided to make these changes it's own PR into WXT, so once #1244 is merged, we can update this branch like so:_
> 
> ```diff
> // zip.ts
>   if (wxt.config.zip.zipSources) {
> +   const entrypoints = await findEntrypoints();
> +   const skippedEntrypoints = entrypoints.filter(entry => entry.skipped);
> +   const excludeSources = [...wxt.config.zip.excludeSources, ...skippedEntrypoints.map(entry => relative(wxt.config.zip.sourcesRoot, entry.inputPath))]
>     await wxt.hooks.callHook('zip:sources:start', wxt);
>     // ...
>     await zipDir(wxt.config.zip.sourcesRoot, sourcesZipPath, {
>       include: wxt.config.zip.includeSources,
> -     exclude: wxt.config.zip.excludeSources,
> +     exclude: excludeSources,
> ```
> 
> _That should be all the changes this branch need._

`const entrypoints = await findEntrypoints();`
This line is causing problem

Cause in the zip.ts file

`const output = await internalBuild();`
is being called which is internally calling `findEntrypoints()`

then 
` const entrypoints = await findEntrypoints();` in line number 58 is calling again.

This is causing:
this below error:
`AssertionError: Expected "entrypoints:resolved" to be called 1 time(s): expected "spy" to be called 1 times, but got 2 times`.

Let me see what I can do.

aklinker1:
@nishu-murmu I would recomend just updating the test to expect the hook to be called twice. That should be fine

nishu-murmu:
> @nishu-murmu I would recomend just updating the test to expect the hook to be called twice. That should be fine

Yeah, I got it.

aklinker1 (approved):
:tada: :shipit:

PR #1280, opened 2024-12-16
chore: Upgrade React to v19 in demo

### LINKED ISSUE #1268: Update React Template to v19
### Feature Request

Update the react template so it uses react version 19. 

This should be as easy as upgrading the UV to plug-in and the dependencies:

https://github.com/wxt-dev/wxt/blob/main/templates%2Freact%2Fpackage.json

> Also, should those be peer dependencies?

https://discord.com/channels/1212416027611365476/1224505556346212515/1316317864755724288

N/A

### Additional context

N/A

nishu-murmu:
This closes #1268


aklinker1 (approved):
Thanks, looks good! We can ignore the failed check, just because this PR is a fork.

aklinker1:
Do we also need to upgrade `@wxt-dev/module-react`?

nishu-murmu:
> Do we also need to upgrade `@wxt-dev/module-react`?

I thought as we or rather you are developing, we might need latest version for testing purposes, as I've often used wxt-demo for fixing bugs and all,
Also just to be in sync with the latest version. If you feel otherwise then I'll revert it back.

aklinker1:
I haven't looked into what's required for react 19, but if we're updating the template, then that's what people are going to use and we need to update the module as well.

nishu-murmu:
So this is correct right? But why ignore the test checks, was it breaking in the main?

aklinker1 (changes requested):

aklinker1 (approved):

PR #1291, opened 2024-12-18
Enable adjusting / passing the zip file artifact template from the config

Dari4sho:


packages/wxt/src/core/resolve-config.ts
```diff
@@ -289,7 +289,7 @@ function resolveZipConfig(
   return {
     name: undefined,
     sourcesTemplate: '{{name}}-{{version}}-sources.zip',
-    artifactTemplate: '{{name}}-{{version}}-{{browser}}.zip',
+    artifactTemplate: mergedConfig.artifactTemplate ?? '{{name}}-{{version}}-{{browser}}.zip',
```
aklinker1, line 292: Hmm, we spread the `mergedConfig` a few lines down. That should overwrite the `artifactTemplate` option. Does that not work for you?

Dari4sho: Aaahh, I'm blind, appreciating the hint, works perfectly fine ♥️

Issue #1302, opened 2024-12-24
anti virus triggering.


SamHoque:
### Describe the bug

When running the build command, I am getting an anti virus alert on the manifest.json file.

![image](https://github.com/user-attachments/assets/342cf9ca-ccf7-4a22-99c0-83b0f74308d3)

### Reproduction

run build

### System Info

```shell
System:
    OS: Windows 11 10.0.22631
    CPU: (24) x64 12th Gen Intel(R) Core(TM) i9-12900K
    Memory: 10.93 GB / 31.79 GB
  Binaries:
    Node: 18.20.2 - C:\Program Files\nodejs\node.EXE
    npm: 10.5.0 - C:\Program Files\nodejs\npm.CMD
    pnpm: 9.12.2 - C:\Program Files\nodejs\pnpm.CMD
    bun: 1.1.42 - ~\.bun\bin\bun.EXE
  Browsers:
    Edge: Chromium (124.0.2478.80)
    Internet Explorer: 11.0.22621.3527
  npmPackages:
    wxt: ^0.19.19 => 0.19.19
```

### Used Package Manager

bun

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/wxt-dev/wxt/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://wxt.dev/guide/installation.html).
- [X] Check that there isn't [already an issue](https://github.com/wxt-dev/wxt/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/wxt-dev/wxt/discussions) or join our [Discord Chat Server](https://discord.gg/ZFsZqGery9).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

aklinker1:
Link for reference: https://www.microsoft.com/en-us/wdsi/threats/malware-encyclopedia-description?Name=Trojan:JS/FakeBardExt.B

Woah, thanks for reporting, that's not good... Unfamiliar with this, and I don't have Windows, so could you do some testing for me?

Does it only happen for v0.19.19 of WXT? What about the latest, v0.19.22? What about a different package manager, does it happen with NPM?

SamHoque:
1. Same issue with Wxt v.0.19.22 
[...]

PR #1304, opened 2024-12-25
🌟 Implemented a way for users to define the output entrypoint name

SamHoque:
Additional changes:
- Added author field to manifest
- Implemented recursive finding of package.json

check #1302

### DIFF
--- a/packages/wxt/src/core/resolve-config.ts
+++ b/packages/wxt/src/core/resolve-config.ts
@@ -220,6 +220,7 @@ export async function resolveConfig(
     userModules,
     plugins: [],
     ...moduleOptions,
+    entrypointNames: mergedConfig.entrypointNames ?? {},
   };
 }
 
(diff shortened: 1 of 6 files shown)


aklinker1:
I don't want to add the ability to add hashes to filenames. This is because in my experience, changing filenames between submissions to the Chrome Web Store and Firefox Addon Store results in longer review times and more manual reviews. My guess is that the diff tools they use don't work well with renamed files, so these submissions get flagged for manual review.

As for your other two changes...

> - Added author field to manifest

Can you open a separate PR for this?

> - Implemented recursive finding of package.json

What is the point of this? What's your use-case for not having a package.json in your extension's root directory?

aiktb:
https://developer.chrome.com/docs/extensions/reference/manifest/author

Adding the Author field will break a lot of extensions, it must match the email of the Google account that submitted the extension.

Even if this doesn't break anything, I don't understand the point of it since the Author field doesn't seem to be displayed anywhere.

aklinker1:
> Does anyone know what this field does?

No idea. I don't see a reason to link it from the package.json either.

aklinker1:
Gonna close this as per the discussion above.

PR #1308, opened 2024-12-27
fix(web-ext): Correctly pass `browserConsole` option to open devtools in Firefox

psrpinto:
The browser console does not currently open on Firefox, because wxt is passing the wrong property to web-ext. The property is called `browserConsole` but wxt was passing `console`. 

See https://github.com/mozilla/web-ext/blob/master/src/cmd/run.js#L25


aklinker1 (approved):
Thanks!

aklinker1:
Released in [`v0.19.24`](https://github.com/wxt-dev/wxt/releases/tag/wxt-v0.19.24)

PR #1317, opened 2024-12-31
fix: missing browser in shadow-root file

1natsu172:
I saw the Discord announcement and tried it and found that shadow RootUi was broken. I found the `browser` import missing(`loadCss` failed), so I fix that.

With this fix both my project and wxt-demo mount fine.


aklinker1:
Thanks... But why didn't typescript catch this 🤔 gonna double check that no more browser imports are missing before merging

1natsu172:
> But why didn't typescript catch this 🤔 

That's because the `browser.runtime.getURL` part has `@ts-expect-error`.

> gonna double check

I checked this regex( `/@ts-expect-error[^\n]*\n[^\n]*\bbrowser\b/g` ) to see if there were any others, but please double check. If you can think of anything else besides `browser`, that too.

aklinker1 (approved):

PR #1332, opened 2025-01-04
feat: Custom types package based on `@types/chrome`

aklinker1:


docs/guide/resources/upgrading.md
```diff
@@ -67,7 +67,11 @@ function getMessageSenderUrl(sender: browser.runtime.MessageSender): string { //
 }
 ```
 
-`@types/chrome` are more up-to-date, contain less bugs, and don't have any auto-generated names. So even if you continue to use the polyfill, you will need to update your types to use these types.
+`@types/chrome` are more up-to-date and contain less bugs, but not all types do not have the same names. You will have to find the equivalent types in `@types/chrome` that you were using before.
```
aleclarson, line 70: > but not all types do not have the same names

Double negative is confusing :)

aklinker1:
Closed in favor of #1530

Issue #1307, opened 2024-12-25
Deno support


aklinker1:
### Feature Request

It is kinda possible to use WXT with [deno](https://deno.com/): [wxt+deno.zip](https://github.com/user-attachments/files/18246802/wxt%2Bdeno.zip)

However there are a few problems:

1. `package.json` isn't found (because we use a `deno.json`) so the name/version can't be extracted automatically
2. `wxt build` and `wxt zip` don't exit
3. TSConfig doesn't exist, so we can't extend `.wxt/tsconfig.json`... Creating a tsconfig.json separate from the `deno.json` breaks `deno run`, so the only option is to reference `.wxt/wxt.d.ts` from another declaration file. This means all the import path aliases won't be recognized. You could manually add them to `deno.json`, but 🤷 I didn't for this example

These things kinda get in the way of the existing template setup and `wxt init` command. Similar to PNPM, we'll need to modify the files after downloading them for it to work with deno.

#### Is your feature request related to a bug?

N/A

### What are the alternatives?

N/A

### Additional context

I've been using deno lately and it's very good. Want to make sure WXT supports it.

PR #1336, opened 2025-01-06
Added missing member used in `UserManifest` to define the manifest name.

Permik:
Hiya!

I was doing cursed things, like setting this project to run via Deno (successfully mind you!) and I came across a smol oversight in the typings, so I patched it up :)

>Added the missing manifest name member when defining the manifest name via the wxt.config.ts manifest option.

Happy to take input if you have any!

Br
Permik

### DIFF
--- a/packages/wxt/src/types.ts
+++ b/packages/wxt/src/types.ts
@@ -916,6 +916,9 @@ export type UserManifest = {
   web_accessible_resources?:
     | string[]
     | chrome.runtime.ManifestV3['web_accessible_resources'];
+} & {
+  // Defines the manifest name, alternatively derived from "<rootDir>/package.json"
+  name?: string
 };
 
 export type UserManifestFn = (


aklinker1:
Thanks for the PR, but the `name` property already has types... Are you not seeing a type for it?

Permik:
Yeah, the member is not defined/suggested for the UserManifest type that the `defineConfig` consumes in `wxt.config.ts`. To note, it is a available indirectly as a member of the `Function` prototype path, but at least with Deno's LSP, the `name` member of the Function Object is not actively suggested as a valid member for the object.
This _might_ just be a a Deno LSP shortcoming fwiw.
```
export default defineConfig({
  manifest: {
    ///
  },
});
```

aklinker1:
I think this is related to Deno. I'll track under #1307. Could you share a reproduction?

Permik:
I'll look into this later today that how could I formulate a reproduction of this issue :D
More of a UX/DX thing, so defining a programmatic repro can be a bit challenging, but I'll see what I can do!

aklinker1:
Ahh, so I actually have the same problem in my deno example from that issue. If I navigate to the definition of `UserManifest`, we see the problem:

Screenshot 2025-01-08 at 12 23 53 AM

If I run `deno install @types/chrome` and add a `import "@types/chrome";` to the top of my `wxt.config.ts` file, it works:

```ts
import { defineConfig } from "wxt";
import denoJson from "./deno.json" with { type: "json" };
import "@types/chrome";

export default defineConfig({
  manifest: {
    name: "Deno Example",
    version: denoJson.version,
  },
});
```

Screenshot 2025-01-08 at 12 28 05 AM

This likely effects more than just the manifest types. It will effect all extension APIs in the upcoming v0.20 release... Hmm will need to consider a fix for this.

aklinker1:
So I'm gonna close this issue, it's definitely a deno-only thing, and this PR doesn't address the root problem.

aklinker1:
I guess the fix is to simply configure the deno template correctly when the time comes. I don't think there will be anything specific to change in WXT's code.

PR #1337, opened 2025-01-06
docs: Add missing web_accessible_resources to `injectScript` example

itsNikolay:
According to the documentation for injectScript(), it requires web_accessible_resources to be defined in the manifest file
(here https://github.com/wxt-dev/wxt/blob/main/packages/wxt/src/client/inject-script.ts#L16).
 However, this is missing from the current documentation. I have added the missing details.

P.S. I also verified that the script does not run without defining web_accessible_resources.

### DIFF
--- a/docs/guide/essentials/content-scripts.md
+++ b/docs/guide/essentials/content-scripts.md
@@ -566,6 +566,20 @@ export default defineContentScript({
 });
 ```
 
+```json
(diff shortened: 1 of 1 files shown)


aklinker1 (approved):
Thanks!

aklinker1 (approved):

Issue #1339, opened 2025-01-08
bug: The dev server did not automatically open devtools


rxliuli:
### Describe the bug

```ts
import { defineConfig } from 'wxt'

// See https://wxt.dev/api/config.html
export default defineConfig({
  srcDir: 'src',
  extensionApi: 'chrome',
  modules: ['@wxt-dev/module-svelte'],
  runner: {
    startUrls: ['https://www.google.com'],
    openDevtools: true,
  },
})
```

https://github.com/user-attachments/assets/4614a75b-180d-4ef0-89b6-a6ee4cccb0a4

The following question seems to be relevant
https://github.com/wxt-dev/wxt/issues/366

### Reproduction

[shadcn-ui-wxt-error-demo.zip](https://github.com/user-attachments/files/18340663/shadcn-ui-wxt-error-demo.zip)

### Steps to reproduce

```sh
pnpm i && pnpm dev
```

### System Info

```shell
System:
    OS: macOS 15.2
    CPU: (12) arm64 Apple M2 Max
    Memory: 6.72 GB / 64.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.12.1 - /usr/local/bin/node
    Yarn: 1.22.22 - /usr/local/bin/yarn
    npm: 10.5.0 - /usr/local/bin/npm
    pnpm: 9.15.0 - /usr/local/bin/pnpm
    bun: 1.1.6 - ~/.bun/bin/bun
    Watchman: 2024.10.07.00 - /opt/homebrew/bin/watchman
  Browsers:
    Brave Browser: 131.1.73.97
    Chrome: 131.0.6778.205
    Safari: 18.2
```

### Used Package Manager

pnpm

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/wxt-dev/wxt/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://wxt.dev/guide/installation.html).
- [X] Check that there isn't [already an issue](https://github.com/wxt-dev/wxt/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/wxt-dev/wxt/discussions) or join our [Discord Chat Server](https://discord.gg/ZFsZqGery9).
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

aklinker1:
[...]

PR #1343, opened 2025-01-08
fix: support openDevtools for chromium

rxliuli:
close https://github.com/wxt-dev/wxt/issues/1339


aklinker1:
See https://github.com/wxt-dev/wxt/issues/1339#issuecomment-2576805956

PR #1364, opened 2025-01-22
Deduplicate `wxt:content-script-started`

### LINKED ISSUE #1359: Premature context invalidation for content scripts
### Describe the bug

I've noticed that `ContentScriptContext` would get invalidated prematurely and therefore restrict any further use of wxt APIs. Initially I thought I missed some configuration with wxt but the minimal reproduction I've described below indicates that there might be a bug with wxt itself.

### Reproduction

Either use the attached archive or follow the steps below:

1. Initialize a vanilla extension via `pnpm dlx wxt@latest init`
2. Use the following script as your content script.
3. This issue does not occur on all pages but try the documentation from Google: https://developers.google.com/docs/api/reference/rest
4. Sometimes this script will work be sometimes not. But the event listener without `ctx` will always work.

```ts
  matches: ["<all_urls>"],
  main(ctx) {
    console.log("Content script loaded", ctx.isValid);
    setInterval(
      () => console.log("Content script is alive", ctx.isValid),
      1000
    );
    window.addEventListener("keydown", (e) =>
      console.log("works without ctx", e.key)
    );
    ctx.addEventListener(window, "keydown", (e) =>
      console.log("doesn't work with ctx", e.key)
    );
  },
});
```

**Reproduction archive:** [test.zip](https://github.com/user-attachments/files/18484244/test.zip)

### Steps to reproduce

_No response_

### System Info

```shell
System:
    OS: Linux 6.11 Ubuntu 24.10 24.10 (Oracular Oriole)
    CPU: (32) x64 Intel(R) Core(TM) i9-14900K
    Memory: 49.19 GB / 61.56 GB
    Container: Yes
    Shell: 5.9 - /usr/bin/zsh
  Binaries:
    Node: 22.13.0 - ~/.nvm/versions/node/v22.13.0/bin/node
    npm: 10.9.2 - ~/.nvm/versions/node/v22.13.0/bin/npm
    pnpm: 9.15.3 - ~/.local/share/pnpm/pnpm
  Browsers:
    Chrome: 132.0.6834.83
```

### Used Package Manager

pnpm

### Validations

DenizUgur:
> This fix prevents premature context invalidation
> fixes #1359

### DIFF
--- a/packages/wxt/src/client/content-scripts/content-script-context.ts
+++ b/packages/wxt/src/client/content-scripts/content-script-context.ts
@@ -42,6 +42,7 @@ export class ContentScriptContext implements AbortController {
   private isTopFrame = window.self === window.top;
   private abortController: AbortController;
   private locationWatcher = createLocationWatcher(this);
+  private receivedTimestamps = new Set<number>();
 
   constructor(
     private readonly contentScriptName: string,
@@ -246,6 +247,10 @@ export class ContentScriptContext implements AbortController {
         event.data?.type === ContentScriptContext.SCRIPT_STARTED_MESSAGE_TYPE &&
         event.data?.contentScriptName === this.contentScriptName
       ) {
+        // Deduplicate messages
+        if (this.receivedTimestamps.has(event.timeStamp)) return;
+        this.receivedTimestamps.add(event.timeStamp);
+
         const wasFirst = isFirst;
         isFirst = false;
         if (wasFirst && options?.ignoreFirstEvent) return;


aklinker1:
> Some pages might send the same message again, either by design or a flaw.

Not by design, I'm curious what is sending the message a second time... Is the message coming from a different frame? The same frame? I don't really understand why we need to deduplicate handling the event when we shouldn't be sending the event a second time in the first place.

DenizUgur:
When I was debugging the library, I have noticed that wxt doesn't send the event twice but rather the page I've linked would sporadically dispatch the events again. With the uglified source code it's hard to say what their justification was.

Are you able to reproduce the issue?

aklinker1:
No, I was not. A separate issue with vite 6.0.9 prevented me from following the repro steps. Been trying to fix that.

But ok, the website sending the event more than one makes sense, that's for the additional detail. In the case, since we can't control when the website re-sends the events, we should previously use the ID approach you mentioned instead of a timestamp.

DenizUgur:
I've revised the PR to use UUID. Last commit [b099287](https://github.com/wxt-dev/wxt/pull/1364/commits/b09928767bdf4d48689f48b718ae8d6a8cfd3bb1) can be reverted if it's not necessary to validate the origin of the messages.

aklinker1 (changes requested):
Thanks, it looks good now. Last change - Can we just use `Math.random` for the ID instead of installing a new package? Or if you want to use a UUID, the builtin crypto API can generate one for you on modern browsers:

https://developer.mozilla.org/en-US/docs/Web/API/Crypto

```ts
const id = crypto?.randomUUID?.() ?? Math.rand()
```

But I think a random number would be fine. Then we don't need to check if the crypto API is available.

DenizUgur:
Sure, I didn't want to use `crypto.randomUUID` because it's only available in secure contexts. I agree though, random number would be enough.

packages/wxt/src/client/content-scripts/content-script-context.ts
```diff
@@ -228,24 +230,41 @@ export class ContentScriptContext implements AbortController {
...
     );
   }
 
+  verifyScriptStartedEvent(event: MessageEvent) {
+    const isScriptStartedEvent =
+      event.data?.type === ContentScriptContext.SCRIPT_STARTED_MESSAGE_TYPE;
+    const isSameContentScript =
+      event.data?.contentScriptName === this.contentScriptName;
+    const isNotDuplicate = !this.receivedMessageIDs.has(event.data?.messageId);
+    const hasValidId = this.sentMessageIDs.has(event.data?.messageId);
```
aklinker1, line 252: Wait, don't we only want to invalidate this context instance if it didn't send the message?

Your logic looks like it will only invalidate the code if this context was the one that sent the message.

aklinker1: Yeah, so the way this works is when a new context is created on a page, it sends out a message to any context already present, telling them they're no longer valid.

aklinker1: Basically makes sure only the newest context is valid. So a context needs to listen for messages from other instances of the class.

DenizUgur: Got it, I've removed that check now.

aklinker1 (approved):

aklinker1:
Nice, thanks. Running checks, will get this merged and deployed once they pass and I get back on

PR #1367, opened 2025-01-24
docs: Add missing bracket in example

rxliuli:
As shown in the picture, a bracket is missing, quickly submit a PR to fix it.

image


Timeraa (approved):
Changes look good, idk about the format/lint though

rxliuli:
> Changes look good, idk about the format/lint though

When submitting, lint automatically did some formatting. Do I need to disable automatic formatting and submit again? Speaking of which, if it's not needed, why is there this lint?

aklinker1 (approved):
It wasn't formatted before because of the syntax error (missing `}`). So now that it's valid JS, it was able to format it. That's how it's supposed to work.

PR #1368, opened 2025-01-24
feat: Add `storage.clear`

chengxilo:
Though it is easy to call browser.storage.clear() to clear all the items, I guess it is needed to support clear in wxt storage. So I add this method.

### DIFF
--- a/packages/storage/src/__tests__/index.test.ts
+++ b/packages/storage/src/__tests__/index.test.ts
@@ -491,16 +491,13 @@ describe('Storage Utils', () => {
 
       describe('clear', () => {
         it('should remove all items', async () => {
-          const item1 = storage.defineItem(`${storageArea}:one`);
-          const item2 = storage.defineItem(`${storageArea}:two`);
           await fakeBrowser.storage[storageArea].set({
             one: 1,
             two: 2,
           });
 
           await storage.clear(storageArea);
-          expect(await item1.getValue()).toBeNull();
-          expect(await item2.getValue()).toBeNull();
+          expect(await fakeBrowser.storage[storageArea].get()).toEqual({});
         });
       });
 
--- a/packages/storage/src/__tests__/index.test.ts
+++ b/packages/storage/src/__tests__/index.test.ts
@@ -489,6 +489,21 @@ describe('Storage Utils', () => {
         });
       });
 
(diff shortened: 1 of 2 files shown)


Timeraa:
I don't quite see why the module would need it if you have `browser.storage.<area>.clear()`

chengxilo:
> I don't quite see why the module would need it if you have `browser.storage.<area>.clear()` since wxt/storage just adds some niche features on top of it

Yes, I agree that current approach of using `browser.storage.<area>.clear()` is enough.

However, I believe it would be much better if all my storage operations were handled uniformly through wxt-storage. At the moment, I use wxt-storage and directly call browser.storage.<area>.clear() because the method is not wrapped. But I wish I can handle all my storage operation in wxt-storage.  In my opinion, consolidating all storage operations under a single abstraction layer would not only simplify the code but also provide a more consistent and pleasant development experience for developer using wxt.

But if wxt-storage is aimed to be a small package and only work for those complex method, I think clear can be ignored, as user can clean their storage easily with directly call and it should be enough.

packages/storage/src/__tests__/index.test.ts
```diff
@@ -489,6 +489,21 @@ describe('Storage Utils', () => {
...
+          const item1 = storage.defineItem(`${storageArea}:one`);
+          const item2 = storage.defineItem(`${storageArea}:two`);
+          await fakeBrowser.storage[storageArea].set({
+            one: 1,
+            two: 2,
+          });
+
+          await storage.clear(storageArea);
+          expect(await item1.getValue()).toBeNull();
+          expect(await item2.getValue()).toBeNull();
```
aklinker1, line 503: We can simplify this test a little, it doesn't need to rely on the storage item implementation. We can check if storage was cleared using just `fakeBrowser`.


```suggestion
          await fakeBrowser.storage[storageArea].set({
            one: 1,
            two: 2,
          });

          await storage.clear(storageArea);
          expect(await fakeBrowser.storage[storageArea].get()).toEqual({});
```

aklinker1 (changes requested):
Thanks for the PR, looks good. Some small changes and we can get this merged!

Timeraa (approved):

aklinker1 (approved):

PR #1384, opened 2025-01-31
chore: add a newline in issue templates config

okineadev:


aklinker1:
Prettier was OK with the previous formatting, I don't see what the point of this PR is

okineadev:
> Prettier was OK with the previous formatting, I don't see what the point of this PR is

Better readability, lumping everything together is not good

aklinker1:
I disagree, I think the `-` is enough of a separator to tell the sections apart.

okineadev:
We need to force the formatter to insert newlines.

Issue #759, opened 2024-06-20
Support custom hostname


aklinker1:
### Feature Request

https://discord.com/channels/1212416027611365476/1231241305116508272

Support passing custom hostname to dev server

#### Is your feature request related to a bug?

Can't run node from docker because the hostname needs to be `0.0.0.0` instead of `localhost`.

aklinker1:
Released in [`v0.18.13`](https://github.com/wxt-dev/wxt/releases/tag/wxt-v0.18.13)

Issue #1360, opened 2025-01-21
Popup HTML auto refresh does not work with Vite 6.0.9+


sapphi-red:
### Describe the bug

Due to a security issue (https://github.com/vitejs/vite/security/advisories/GHSA-vg6x-rcgg-rjx6), Vite 6.0.9+ now requires a token to connect to the WebSocket endpoint.
Because WXT connects to the WS server on its own, this connection no longer successes.
https://github.com/wxt-dev/wxt/blob/7c18d9c0b8c638204b913b13de25c08ced5d9eb5/packages/wxt/src/sandbox/dev-server-websocket.ts#L43-L46

The way to fix would be to pass `config.webSocketToken` as a query param (e.g. `?token=${config.webSocketToken}`) when connecting to WS server.
https://github.com/vitejs/vite/blob/9654348258eaa0883171533a2b74b4e2825f5fb6/packages/vite/src/node/config.ts#L610-L620
But ideally, it should not connect to the WebSocket server on its own and instead use `import.meta.hot`.

### Reproduction

`pnpm dlx wxt@latest init`

### Steps to reproduce

1. `pnpm dlx wxt@latest init`
2. `pnpm dev`
3. Open the extension popup.
4. Edit `entrypoints/popup/index.html`

### System Info

```shell
System:
    OS: Windows 11 10.0.22631
    CPU: (24) x64 AMD Ryzen 9 5900X 12-Core Processor
    Memory: 11.03 GB / 31.92 GB
  Binaries:
    Node: 22.13.0 - C:\Program Files\nodejs\node.EXE
    npm: 10.9.2 - C:\Program Files\nodejs\npm.CMD
    pnpm: 9.15.3 - C:\Program Files\nodejs\pnpm.CMD
    bun: 1.1.38 - ~\AppData\Local\Microsoft\WinGet\Links\bun.EXE
  Browsers:
    Edge: Chromium (131.0.2903.99)
    Internet Explorer: 11.0.22621.3527
```

### Used Package Manager

pnpm

### Validations

- [x] Read the [Contributing Guidelines](https://github.com/wxt-dev/wxt/blob/main/CONTRIBUTING.md).
- [x] Read the [docs](https://wxt.dev/guide/installation.html).
- [x] Check that there isn't [already an issue](https://github.com/wxt-dev/wxt/issues) that reports the same bug to avoid creating a duplicate.
[...]

PR #1406, opened 2025-02-07
feat: enable wxt usage inside of devcontainers

alecdwm:
### Overview

For users who want to run the dev server inside of a container, e.g. via orbstack or devcontainers, it's necessary to configure the dev server *bind address* separately to the dev server *origin* or websocket url.

Otherwise, the following error is thrown in the extension console and hot reloading is broken:

Insecure CSP value "http://0.0.0.0:3000" in directive 'script-src'.

The *bind address*, hereafter referred to as `host`, is always going to be either `localhost`/`127.0.0.1`/`::1` to lock it down to the local machine, or `0.0.0.0`/`::` to open it up to LAN traffic.

Whereas the `origin` could be `localhost`, `my-dev-computer.local`, `some-dev-container.orb.local`, or all sorts of different configurations depending on the environment.

The existing codebase conflates these two variables into the one `hostname`, which makes using wxt impossible in the above scenario.

[The full Discord discussion can be found here](<https://discord.com/channels/1212416027611365476/1231241305116508272/1260700302630326342>)

### Solution

The changes in this PR introduce the two new config parameters, `host` and `origin`, with an attempt to maintain backwards compatibility with existing user configurations (`hostname`) as well as some sane but overridable defaults.

- `hostname` has been replaced with `host` everywhere that the code is expecting a bind address.
- `hostname` has been replaced with `origin` everywhere that the code is expecting a URL which the frontend can use to contact the dev server.
- If provided, the `hostname` config option will set `host` and `origin` just as it would have before these changes, but a deprecation notice will be served on the command line.
- If `host` is not configured, it will default to `localhost`.
- If `origin` is not configured, it will default to `http://localhost:<dev-server-port>`
- If `origin` is configured to a plain hostname, the protocol and port will be automatically added e.g. `dev.local` -> `http://dev.local:<dev-server-port>`
- If the user is running the dev server behind a proxy, the protocol and port can be manually overridden e.g. `origin: "https://dev.local:9090"` for a proxy with TLS listening on port 9090

### Manual Testing

To test, run wxt inside of a docker container using e.g. docker desktop or orbstack. Make sure that the `.output` directory is bind-mounted to your host so that you can install the extension from the filesystem into Chrome.

Once running, add the extension to Chrome and open the background console. Note that the console is unable to connect to the dev server.

Restart wxt inside of the container with the CLI argument `--host 0.0.0.0` and reload the extension in Chrome. Note that the dev server is now contactable, there are no CSP errors, and that hot reloading works as expected!

### Related Issues

This PR is a revision of the changes in https://github.com/wxt-dev/wxt/issues/759 / https://github.com/wxt-dev/wxt/pull/807


packages/wxt/src/@types/globals.d.ts
```diff
@@ -1,6 +1,4 @@
-declare const __DEV_SERVER_PROTOCOL__: string;
-declare const __DEV_SERVER_HOSTNAME__: string;
-declare const __DEV_SERVER_PORT__: string;
+declare const __DEV_SERVER_ORIGIN__: string;
```
alecdwm, line 1: Instead of constructing a best-guess URL to use to contact the dev server, we can use `ORIGIN` here

packages/wxt/src/cli/commands.ts
```diff
@@ -37,13 +44,9 @@ cli
         debug: flags.debug,
         filterEntrypoints: getArrayFromFlags(flags, 'filterEntrypoint'),
         dev:
-          flags.port == null
+          Object.keys(serverOptions).length === 0
             ? undefined
```
alecdwm, line 48: There's a test which requires this to be `undefined` when neither `host` nor `port` are set

packages/wxt/src/core/builders/vite/plugins/devServerGlobals.ts
```diff
@@ -15,9 +15,9 @@ export function devServerGlobals(
 
       return {
         define: {
-          __DEV_SERVER_PROTOCOL__: JSON.stringify('ws:'),
-          __DEV_SERVER_HOSTNAME__: JSON.stringify(server.hostname),
-          __DEV_SERVER_PORT__: JSON.stringify(server.port),
+          __DEV_SERVER_ORIGIN__: JSON.stringify(
+            server.origin.replace(/^http(s):/, 'ws$1:'),
```
alecdwm, line 19: The regex replace converts `http:` to `ws:` and `https:` to `wss:`

packages/wxt/src/core/utils/manifest.ts
```diff
@@ -467,7 +467,7 @@ function discoverIcons(
 }
 
 function addDevModeCsp(manifest: Manifest.WebExtensionManifest): void {
-  const permission = `http://${wxt.server?.hostname ?? ''}/*`;
+  const permission = `${wxt.server?.origin.replace(/:[0-9]+$/, '') ?? `http://localhost:${wxt.server?.port}`}/*`;
```
alecdwm, line 470: The port has to be removed from `origin` in order to comply with existing CSP rules tests.  
I think changing the expected CSP is out of scope for this PR. If I encounter issues using the current rules with devcontainers I'll open another PR!

aklinker1 (commented):
Thanks for taking the time to write your PR! Looks good, I'll get this tested and merged soon.

> For some history, I was always confused why Vite chose the name "host" instead of "hostname", because that was the part of the URL the string was used in. But that's not true, it's the host they're binding the server to, in networking terms, not in "parts of the URL" terms. And thus "hostname" was chosen incorrectly in WXT 🤦 

Unfortunately, this PR is blocked by https://github.com/wxt-dev/wxt/issues/1360#issuecomment-2644442432, so it can't be merged right away - my goal is to get that fixed and merged within the next day. Essentially, I'm removing any code used to connect to the dev server manually, using `import.meta.hot` instead. This change is to comply with the security updates in [Vite 6.0.9](https://github.com/vitejs/vite/blob/v6.0.11/packages/vite/CHANGELOG.md#609-2025-01-20). But there will likely be some conflicts with this PR.

I'm gonna give that fix priority over this one since it's security related. Not sure how it's going to effect this PR's behavior...

In the meantime, could you also add a FAQ question around how to run a WXT project inside docker? That would be really good to document!

https://github.com/wxt-dev/wxt/blob/main/docs/guide/resources/faq.md

alecdwm:
Sounds great! I'll rebase and fix conflicts after you've had a chance to deal with https://github.com/wxt-dev/wxt/issues/1360#issuecomment-2644442432

Will also update the FAQ 😀

aklinker1:
> Unfortunately, this PR is blocked by https://github.com/wxt-dev/wxt/issues/1360#issuecomment-2644442432, so it can't be merged right away

@alecdwm Well, I wasn't able to get that change released quickly... So let's get your PR merged.

alecdwm:
@aklinker1 Hey hey! I've rebased onto latest `main`, fixed the checks and updated the faq 💪

steffkes:
i'd like to give this change a try (because containers!) .. but the pkg.pr.new hasn't built for the latest change and the previous one fails with a ominous error message of 

>  `ERROR` addDevModePermissions is not defined

which i'm not entirely getting? I've installed it with `npm install --save-dev https://pkg.pr.new/wxt@140`

steffkes:
btw I have no idea what exactly changed that finally led to this error message .. but I verified that this function is indeed not there after all:

wxt@0.19.26:
```
# grep -rn "addDevModePermissions" node_modules/wxt/
node_modules/wxt/dist/core/utils/manifest.mjs:72:  if (wxt.config.command === "serve") addDevModePermissions(manifest);
node_modules/wxt/dist/core/utils/manifest.mjs:377:function addDevModePermissions(manifest) {
```

wxt@0.20.0: 
```
# grep -rn "addDevModePermissions" node_modules/wxt/
node_modules/wxt/dist/core/utils/manifest.mjs:75:  if (wxt.config.command === "serve") addDevModePermissions(manifest);
node_modules/wxt/dist/core/utils/manifest.mjs:375:function addDevModePermissions(manifest) {
```

https://pkg.pr.new/wxt@1406:
```
# grep -rn "addDevModePermissions" node_modules/wxt/
node_modules/wxt/dist/core/utils/manifest.mjs:71:  if (wxt.config.command === "serve") addDevModePermissions(manifest);
```

alecdwm:
[aklinker1](https://github.com/aklinker1) fixed conflicts with `main`

packages/wxt/src/core/utils/manifest.ts
```diff
@@ -458,7 +458,7 @@ function discoverIcons(
   return icons.length > 0 ? Object.fromEntries(icons) : undefined;
 }
 
-function addDevModeCsp(manifest: Browser.runtime.Manifest): void {
+function addDevModeCsp(manifest: Manifest.WebExtensionManifest): void {
```
aklinker1, line 461: You can undo this change. The old code is correct in v0.20

```suggestion
function addDevModeCsp(manifest: Browser.runtime.Manifest): void {
```

aklinker1 (changes requested):
One small change and we're good to merge this!

steffkes:
That's great, thanks for working on it @alecdwm - with the latest changes (https://pkg.pr.new/wxt@3c03641) the function is restored and the development environment inside a docker container works as expected.

well, almost anyways :) but i think this should be worked on in a separate ticket. it's running `npm run dev:firefox:container` which is defined as `wxt -b firefox --host 0.0.0.0` in `package.json`

```
$ docker compose up app
[+] Running 1/1
 ✔ Container extension-wxt-app-1  Creat...                                 0.0s
Attaching to app-1
app-1  |
app-1  | > wxt-vue-starter@0.0.0 dev:firefox:container
app-1  | > wxt -b firefox --host 0.0.0.0
...
app-1  |   └─ .output/chrome-mv3-dev/wxt.svg                     1.07 kB
app-1  | Σ Total size: 32.98 kB
app-1  | ℹ Load ".output/chrome-mv3-dev" as an unpacked extension manually
app-1 exited with code 0
```

since the command does exit .. the entire container stops. i'm currently working around this limitation with:

```
$ docker compose run --service-ports app ash
/app # npm run dev:firefox:container

> wxt-vue-starter@0.0.0 dev:firefox:container
> wxt -b firefox --host 0.0.0.0

...
  └─ .output/firefox-mv2-dev/wxt.svg                     1.07 kB
Σ Total size: 32.71 kB
ℹ Load ".output/firefox-mv2-dev" as an unpacked extension manually
```

which keeps running, while the `--service-ports` passed to `docker compose run` exposes the service on the configure port (like `docker compose up` would do)

while i'm not sure at all where this behaviour originates, but i also never looked into yet

aklinker1:
Huh, so the dev command exits without pressing `ctrl+C`? That's odd... Yeah, we can look into it in another issue. I don't think that's blocking us from merging this code as-is.

alecdwm:
Should be good to go now @aklinker1!

aklinker1 (approved):

czy1996:
Really nice work, thanks!
A little more help wanted, the feature was not shipped with v0.20.0, how can I give this a try before next version is released ? Sorry about the stupid question but I'm new to frontend world...

aklinker1:
> Really nice work, thanks! A little more help wanted, the feature was not shipped with v0.20.0, how can I give this a try before next version is released ? Sorry about the stupid question but I'm new to frontend world...

See https://github.com/wxt-dev/wxt/pull/1406#issuecomment-2696147893 and https://github.com/wxt-dev/wxt/blob/main/CONTRIBUTING.md#install-unreleased-versions

packages/wxt/src/core/builders/vite/plugins/devServerGlobals.ts
Sokwva: after upgrade wxt from 0.20.0 to 0.20.6, an error print in background console.I want to know if this error is related to this change 😭
```
[wxt] Failed to setup web socket connection with dev server DOMException: Failed to construct 'WebSocket': The URL's scheme must be either 'ws' or 'wss'. 'http' is not allowed.
    at getDevServerWebSocket
```
OS: windows 10 21H2
browser: Edge 119.0.2151.72

alecdwm: Ah whoops! The regex is missing a `?`, so it doesn’t handle `http` like it is supposed to.
The correct regex should be: `^http(s)?:`
cc @aklinker1

aklinker1: @alecdwm @Sokwva Fixed by https://github.com/wxt-dev/wxt/pull/1707

PR #1416, opened 2025-02-11
fix: respect `inlineConfig.mode` when setting NODE_ENV

aleclarson:
### Overview

I think this fix is self-explanatory, but let me know if you need clarification.

### Manual Testing

1. Add an expression that uses `import.meta.env.DEV` (or a statement wrapped with an `if (import.meta.env.DEV)` condition)

2. Build with `pnpm build --mode=development`

3. Notice how, without this PR, `import.meta.env.DEV` unexpectedly resolves to false.


Timeraa (approved):

aklinker1:
Released in [v0.19.27](https://github.com/wxt-dev/wxt/releases/tag/wxt-v0.19.27)

PR #1417, opened 2025-02-11
fix: resolve wxt modules from the root

aleclarson:
### Overview

I have a local clone of WXT which I make fixes on and link into my web extension project for manual testing. My project is using a "WXT module" which is not being found by the current logic (before this PR). The reason is that WXT does a basic `import(…)` of each string in the `modules` array, which doesn’t work when the `wxt` package is symlinked into the project.

I thought about using [import-meta-resolve](https://www.npmjs.com/package/import-meta-resolve) for a more comprehensive solution. For simplicity, I went with the same approach you took with `resolveWxtModuleDir` which is just using `createRequire` from the `node:module` API.

### DIFF
--- a/packages/wxt/src/core/resolve-config.ts
+++ b/packages/wxt/src/core/resolve-config.ts
@@ -1,5 +1,5 @@
 import { loadConfig } from 'c12';
-import { createRequire } from 'node:module';
+import { resolve as esmResolve } from 'import-meta-resolve';
 import {
   InlineConfig,
   ResolvedConfig,
@@ -29,6 +29,7 @@ import { getEslintVersion } from './utils/eslint';
 import { safeStringToNumber } from './utils/number';
 import { loadEnv } from './utils/env';
 import { getPort } from 'get-port-please';
+import { fileURLToPath, pathToFileURL } from 'node:url';
 
 /**
  * Given an inline config, discover the config file if necessary, merge the results, resolve any
@@ -157,10 +158,7 @@ export async function resolveConfig(
   }
 
   const userModules = await resolveWxtUserModules(
-    path.resolve(
-      root,
-      inlineConfig.configFile ? path.dirname(inlineConfig.configFile) : '.',
-    ),
+    root,
     modulesDir,
     mergedConfig.modules,
   );
@@ -426,17 +424,21 @@ async function getUnimportEslintOptions(
  * Returns the path to `node_modules/wxt`.
  */
 function resolveWxtModuleDir() {
-  // TODO: Use this once we're fully running in ESM, see https://github.com/wxt-dev/wxt/issues/277
-  // const url = import.meta.resolve('wxt', import.meta.url);
-  // resolve() returns the "wxt/dist/index.mjs" file, not the package's root
-  // directory, which we want to return from this function.
-  // return path.resolve(fileURLToPath(url), '../..');
+  // TODO: Drop the __filename expression once we're fully running in ESM
+  // (see https://github.com/wxt-dev/wxt/issues/277)
+  const importer =
+    typeof __filename === 'string'
+      ? pathToFileURL(__filename).href
+      : import.meta.url;
 
-  const nodeRequire = globalThis.require ?? createRequire(import.meta.url);
+  // TODO: Switch to import.meta.resolve() once the parent argument is unflagged
+  // (e.g. --experimental-import-meta-resolve) and all Node.js versions we support
+  // have it.
+  const url = esmResolve('wxt', importer);
 
-  // resolve() returns the "wxt/dist/index.mjs" file, not the package's root
+  // esmResolve() returns the "wxt/dist/index.mjs" file, not the package's root
   // directory, which we want to return from this function.
-  return path.resolve(nodeRequire.resolve('wxt'), '../..');
+  return path.resolve(fileURLToPath(url), '../..');
 }
 
 async function isDirMissing(dir: string) {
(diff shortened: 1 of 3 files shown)


packages/wxt/src/core/resolve-config.ts
```diff
@@ -479,14 +481,18 @@ export async function mergeBuilderConfig(
 }
 
 export async function resolveWxtUserModules(
+  rootDir: string,
   modulesDir: string,
   modules: string[] = [],
 ): Promise<WxtModuleWithMetadata<any>[]> {
+  const nodeRequire = createRequire(path.join(rootDir, 'index.js'));
```
aleclarson, line 488: `index.js` is not required to exist, but we cannot use a directory with `createRequire` or else the directory itself won't be searched

packages/wxt/src/core/resolve-config.ts
```diff
@@ -156,6 +157,10 @@ export async function resolveConfig(
   }
 
   const userModules = await resolveWxtUserModules(
+    path.resolve(
+      root,
+      inlineConfig.configFile ? path.dirname(inlineConfig.configFile) : '.',
+    ),
```
aklinker1, line 163: Modules should always go in `<root>/modules`, where the config file is placed doesn't effect this.

```suggestion
    root,
```

packages/wxt/src/core/resolve-config.ts
```diff
@@ -1,4 +1,5 @@
 import { loadConfig } from 'c12';
+import { createRequire } from 'node:module';
```
aklinker1, line 2: I believe checks are failing because you moved this to a static import. It was causing problems as a static, that's why I used a dynamic import before.

packages/wxt/src/core/resolve-config.ts
```diff
@@ -479,14 +481,18 @@ export async function mergeBuilderConfig(
...
+  rootDir: string,
   modulesDir: string,
   modules: string[] = [],
 ): Promise<WxtModuleWithMetadata<any>[]> {
+  const nodeRequire = createRequire(path.join(rootDir, 'index.js'));
+
   // Resolve node_modules modules
   const npmModules = await Promise.all<WxtModuleWithMetadata<any>>(
     modules.map(async (moduleId) => {
+      const resolvedModuleId = nodeRequire.resolve(moduleId);
```
aklinker1, line 493: OK, so the main change here is that we need to resolve the ID from the root directory before importing it? That fixes the issue when WXT is symlinked?

packages/wxt/src/core/resolve-config.ts
aleclarson: The failed check is due to `.cjs` entry points being resolved by `nodeRequire.resolve`. The CJS entry point then uses `require("wxt")` which is not supported, because it resolves to an ES module.

aleclarson: My recent commit has fixed this.

packages/wxt/src/core/resolve-config.ts
```diff
@@ -426,17 +424,21 @@ async function getUnimportEslintOptions(
...
-
-  const nodeRequire = globalThis.require ?? createRequire(import.meta.url);
-
-  // resolve() returns the "wxt/dist/index.mjs" file, not the package's root
+  // TODO: Drop the __filename expression once we're fully running in ESM
+  // (see https://github.com/wxt-dev/wxt/issues/277)
+  const importer =
+    typeof __filename === 'string'
+      ? pathToFileURL(__filename).href
+      : import.meta.url;
```
aleclarson, line 432: Not 100% sure the `__filename` check is necessary, tbh. This was added defensively, in case the Node.js version doesn't have `import.meta.url`.

packages/wxt/src/core/resolve-config.ts
```diff
@@ -426,17 +424,21 @@ async function getUnimportEslintOptions(
...
+
+  // TODO: Switch to import.meta.resolve() once the parent argument is unflagged
+  // (e.g. --experimental-import-meta-resolve) and all Node.js versions we support
+  // have it.
+  const url = esmResolve('wxt', importer);
```
aleclarson, line 437: I could've stuck with `createRequire` here, since we're only using a parent directory of the resolved entry point, so it doesn't matter if a CJS module is returned, but I decided it was cleaner to use `esmResolve` since we have it in scope. LMK if you disagree.

packages/wxt/src/core/resolve-config.ts
aklinker1: We definitely need to keep it in. I don't want to switch to using `import.meta.url` when older versions of node 18 and node 20 don't support it yet. v0.20 will migrate us fully to ESM, at which point I'm happy removing the `__filename` check. But that needs to come with a breaking change, it can't be in a patch request.

packages/wxt/src/core/resolve-config.ts
```diff
@@ -1,5 +1,5 @@
 import { loadConfig } from 'c12';
-import { createRequire } from 'node:module';
+import { resolve as esmResolve } from 'import-meta-resolve';
```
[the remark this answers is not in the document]
aklinker1: We'll give it a go!

packages/wxt/src/core/resolve-config.ts
aklinker1: Or maybe I'm misremembering... It looks like I was using it before? But after checking for the global `require` function... Hmm, not sure. I always have to look things up whenever it comes to ESM/CJS interop

aklinker1 (approved):
Alright, let's give it a go!

aklinker1:

PR #1432, opened 2025-02-13
fix: Properly detect and reload changed files when using Vite's `?suffix` imports

### LINKED ISSUE #1062: CSS changes are not applied without `?url` (HMR)
### Describe the bug

I am manually injecting a CSS file located within `entrypoints/`

`entrypoints/content.ts`
```ts
import "./some.css";
// if the above line isn't included, then a change in that file will not trigger a reload
// if the above line is included, then a change in that file will trigger a reload, but not update the contents of the css file
// to actually update the css, you have to quit wxt entirely and start a new dev server

  matches: ['*://example.com/*'],
  cssInjectionMode: "manual", // this line doesn't seem to make a difference, unless the lines 12-17 are commented out, in which case it prevents styles from being injected
  main() {
    console.log('Hello content.');
    // if the below lines are commented out (and line 9, css injection mode), the issue is fixed
    // let link = document.createElement("link");
    // link.rel = "stylesheet";
    // link.href = browser.runtime.getURL("/assets/some.css" as any);
    // link.classList.add("schooltape");
    // document.head.appendChild(link);
  },
});

```

May be related to https://github.com/wxt-dev/wxt/issues/386

### Reproduction

1. clone https://github.com/42Willow/wxt-css-reload-bug-repro and cd
2. `pnpm install`
3. `pnpm run dev`
4. navigate to https://example.com
5. modify `entrypoints/some.css` - extension reloads but styles are not changed

It doesn't seem to have any effect on whether the css file is explicitly imported into the content script or not, only that the extension only reloads if there is a change **and** the css file has been imported by the content script. Is this intended behaviour? as the css file is located within `entrypoints/`

### Steps to reproduce

built files are located in `dist/` of the repo provided above (output of `pnpm dev`)

### System Info

```shell
System:
    OS: Linux 6.6 NixOS 24.11 (Vicuna) 24.11 (Vicuna)
    CPU: (8) x64 Intel(R) Core(TM) i5-10210U CPU @ 1.60GHz
    Memory: 9.29 GB / 15.26 GB
    Container: Yes
    Shell: 0.98.0 - /run/current-system/sw/bin/nu
  Binaries:
    Node: 20.17.0 - /etc/profiles/per-user/willow/bin/node
    npm: 10.8.2 - /etc/profiles/per-user/willow/bin/npm
    pnpm: 9.10.0 - /etc/profiles/per-user/willow/bin/pnpm
    bun: 1.1.29 - /etc/profiles/per-user/willow/bin/bun
  Browsers:
    Chromium: 129.0.6668.89
  npmPackages:
    wxt: ^0.19.11 => 0.19.11
```

### Used Package Manager

pnpm

### Validations

1natsu172:
### Overview

This PR for the #1062. The user modified file path has no suffix, and the moduleIds path written in the import syntax has a suffix. So it was unmatched and dev server no-changes.

I've changed the code style a bit. If you don't like it, please modify it to your liking.

### Manual Testing

1. cd wxt-demo && pnpm dev
2. edit & save `src/entrypoints/ui.content/manual-style.css`
3. check reload and apply new styles.

### Related Issue

This PR closes #1062

### DIFF
--- a/packages/wxt-demo/src/entrypoints/ui.content/index.ts
+++ b/packages/wxt-demo/src/entrypoints/ui.content/index.ts
@@ -1,11 +1,16 @@
 import 'uno.css';
 import './style.css';
+import manualStyle from './manual-style.css?inline';
 
 export default defineContentScript({
   matches: ['https://*.duckduckgo.com/*'],
   cssInjectionMode: 'ui',
 
   async main(ctx) {
+    const style = document.createElement('style');
+    style.textContent = manualStyle;
+    document.head.append(style);
+
     const ui = await createShadowRootUi(ctx, {
       name: 'demo-ui',
       position: 'inline',
--- a/packages/wxt-demo/src/entrypoints/ui.content/manual-style.css
+++ b/packages/wxt-demo/src/entrypoints/ui.content/manual-style.css
@@ -1,0 +1,4 @@
+body {
+  padding: 2rem;
+  background-color: blanchedalmond;
+}
--- a/packages/wxt/src/core/utils/building/__tests__/detect-dev-changes.test.ts
+++ b/packages/wxt/src/core/utils/building/__tests__/detect-dev-changes.test.ts
@@ -359,5 +359,63 @@ describe('Detect Dev Changes', () => {
 
       expect(actual).toEqual(expected);
     });
+
+    it('should detect changes to import files with `?suffix`', () => {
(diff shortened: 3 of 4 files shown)


aklinker1 (approved):
Nice, thanks for adding a test.

> I've changed the code style a bit. If you don't like it, please modify it to your liking.

Nope, no changes, looks much better!

PR #1449, opened 2025-02-18
fix: regression on variable expansion in dotenv files

stefanmaric:
### Overview

[PR #978](https://github.com/wxt-dev/wxt/pull/978), released in [v0.19.10](https://github.com/wxt-dev/wxt/releases/tag/wxt-v0.19.10),
broke variable expansion in dotenv files, a Vite built-in feature.

I encountered this while migrating from v0.18.2 to the latest.

Since already defined variables in process.env take presedence,
the call to dotenv-expand by vite won't have any effect. Calling
dotenv-expand ourselves to mimic what vite does restores the lost
functionality.

Vite has a slightly more complex env loading logic, so these changes
won't bring 100% parity, but should be good for most use-cases.

See: https://github.com/vitejs/vite/blob/642d528b7b403eb91c67ff809ffa0fb99a1ff56e/packages/vite/src/node/env.ts

### Manual Testing

I've tested this in production already by applying the very same
changes with `pnpm patch` in my project.

### Related Issue

None reported so far (I did not bother to create an issue first).

### DIFF
--- a/packages/wxt/package.json
+++ b/packages/wxt/package.json
@@ -98,6 +98,7 @@
     "consola": "^3.2.3",
     "defu": "^6.1.4",
     "dotenv": "^16.4.5",
+    "dotenv-expand": "^12.0.1",
     "esbuild": "^0.21.5",
     "fast-glob": "^3.3.2",
     "filesize": "^10.1.6",
--- a/packages/wxt/src/core/utils/env.ts
+++ b/packages/wxt/src/core/utils/env.ts
@@ -1,21 +1,24 @@
 import { config } from 'dotenv';
+import { expand } from 'dotenv-expand';
 import type { TargetBrowser } from '../../types';
 
 /**
  * Load environment files based on the current mode and browser.
  */
 export function loadEnv(mode: string, browser: TargetBrowser) {
-  return config({
-    // Files on top override files below
-    path: [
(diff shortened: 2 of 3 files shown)


stefanmaric:
> @aklinker1 wants to keep dependency usage down

If it helps put your mind at ease, this change technically doesn't introduce any new dependencies since `dotenv-expand` is already a transitive dependency via `vite`.

See: https://github.com/vitejs/vite/blob/60456a54fe90872dbd4bed332ecbd85bc88deb92/packages/vite/package.json#L118

aklinker1:
> If it helps put your mind at ease, this change technically doesn't introduce any new dependencies since `dotenv-expand` is already a transitive dependency via `vite`.

Yeah, adding this dependency makes sense. It'll at least be shared with Vite until there's a major version update and we get out of sync.

aklinker1:
Also, looks like there's a type error due to subdependency changes. Try running `pnpm i --fix-lockfile` to resolve it. Check with `pnpm check`.

If that doesn't work, I can fix it later.

stefanmaric:
> Also, looks like there's a type error due to subdependency changes. Try running `pnpm i --fix-lockfile` to resolve it. Check with `pnpm check`.

Indeed. I've pushed the package lock fix. 👍

aklinker1:

PR #1465, opened 2025-02-28
fix: restarting keyboardShortcuts listener on file changes.

nishu-murmu:
### Overview

I noticed that keyboard shortcuts weren't working after files were being saved or HMR was triggered.
Hope this fixes it.

### DIFF
--- a/packages/wxt/src/core/create-server.ts
+++ b/packages/wxt/src/core/create-server.ts
@@ -109,13 +109,15 @@ async function createServerInternal(): Promise<WxtDevServer> {
       await buildAndOpenBrowser();
 
       // Listen for file changes and reload different parts of the extension accordingly
-      const reloadOnChange = createFileReloader(server);
+      const reloadOnChange = () => {
+        createFileReloader(server);
+        keyboardShortcuts.start();
+        keyboardShortcuts.printHelp({
+          canReopenBrowser:
+            !wxt.config.runnerConfig.config.disabled && !!runner.canOpen?.(),
+        });
+      };
       server.watcher.on('all', reloadOnChange);
-      keyboardShortcuts.start();
-      keyboardShortcuts.printHelp({
-        canReopenBrowser:
-          !wxt.config.runnerConfig.config.disabled && !!runner.canOpen?.(),
-      });
     },
 
     async stop() {


Timeraa (commented):
Was noticing this too but wouldn't this potentially be a memory leak since we keep creating it?

nishu-murmu:
> Was noticing this too but wouldn't this potentially be a memory leak since we keep creating it?

Yeah, now that I think about it, it really can be. as such there are already existing issues for windows.
We can just close the listeners instead of the interface. That way only one interface will be used.

aleclarson (changes requested):

packages/wxt/src/core/create-server.ts
```diff
@@ -111,13 +111,15 @@ async function createServerInternal(): Promise<WxtDevServer> {
       await buildAndOpenBrowser();
 
       // Listen for file changes and reload different parts of the extension accordingly
-      const reloadOnChange = createFileReloader(server);
+      const reloadOnChange = () => {
+        createFileReloader(server);
```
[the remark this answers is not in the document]
nishu-murmu: Thanks for pointing it out. completely overlooked that 😓. I've returned that `createFileReloader` function.

packages/wxt/src/core/keyboard-shortcuts.ts
```diff
@@ -37,8 +40,9 @@ export function createKeyboardShortcuts(
     },
 
     stop() {
-      rl?.close();
-      rl = undefined;
+      if (rl) {
+        rl.removeListener('line', handleInput);
```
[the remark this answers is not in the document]
nishu-murmu: Added that.

packages/wxt/src/core/keyboard-shortcuts.ts
```diff
@@ -26,7 +26,10 @@ export function createKeyboardShortcuts(
 
   return {
     start() {
-      if (rl) return;
+      if (rl) {
+        rl.on('line', handleInput);
+        return;
+      }
 
       rl = readline.createInterface({
```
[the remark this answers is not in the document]
nishu-murmu: Added this as well.

packages/wxt/src/core/keyboard-shortcuts.ts
```diff
@@ -37,8 +35,9 @@ export function createKeyboardShortcuts(
     },
 
     stop() {
-      rl?.close();
-      rl = undefined;
+      if (rl) {
+        rl?.removeListener('line', handleInput);
```
aleclarson, line 39: don't need the `if` check since `?.` does that

packages/wxt/src/core/create-server.ts
```diff
@@ -111,13 +111,15 @@ async function createServerInternal(): Promise<WxtDevServer> {
...
+            !wxt.config.runnerConfig.config.disabled && !!runner.canOpen?.(),
+        });
+        return createFileReloader(server);
+      };
       server.watcher.on('all', reloadOnChange);
-      keyboardShortcuts.start();
-      keyboardShortcuts.printHelp({
-        canReopenBrowser:
-          !wxt.config.runnerConfig.config.disabled && !!runner.canOpen?.(),
-      });
```
aklinker1, line 120: One side-effect to this change is that the help message is printed after every file is saved, right? Should we leave the `printHelp` call outside the `reloadOnChange` where it was before so we only print the message when the server is started/restarted?

packages/wxt/src/core/create-server.ts
```diff
@@ -111,13 +111,15 @@ async function createServerInternal(): Promise<WxtDevServer> {
...
-      const reloadOnChange = createFileReloader(server);
+      const reloadOnChange = () => {
+        keyboardShortcuts.start();
+        keyboardShortcuts.printHelp({
+          canReopenBrowser:
+            !wxt.config.runnerConfig.config.disabled && !!runner.canOpen?.(),
+        });
+        return createFileReloader(server);
+      };
       server.watcher.on('all', reloadOnChange);
```
aklinker1, line 122: I think this line is still wrong...

```suggestion
      const reloadOnChange = createFileReloader(server);
      server.watcher.on('all', async (...args) => {
        await reloadOnChange(...args);
        
        // Restart keyboard shortcuts after file is changed - for some reason they stop working.
        keyboardShortcuts.start();
      });
      
      keyboardShortcuts.printHelp({
        canReopenBrowser:
          !wxt.config.runnerConfig.config.disabled && !!runner.canOpen?.(),
      });
```

aklinker1 (changes requested):
Do we know why the readline interface stops working when a file is saved? Or are we just making a change in hopes that it fixes the problem?

packages/wxt/src/core/create-server.ts
nishu-murmu: Yeah, we can keep it outside, users can see it once when they restart the server.
Vite is also doing the same.

packages/wxt/src/core/create-server.ts
nishu-murmu: I've updated the code as you recommended, I've passed the first two arguments as it was throwing type error.

aklinker1 (approved):
Let's merge this and give it a go.

Edit: hmm, for some reason checks aren't running. I'll get the going when I get home.

PendalF89:
Hi guys, after upgrading from 0.20.7 => 0.20.8 I encountered with this warnings in console:

```
✔ Opened browser in 618 ms                                                                                                                  16:37:31
ℹ Press o + enter to reopen the browser                                                                                                     16:37:31
(node:1604) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 line listeners added to [Interface]. MaxListeners is 10. Use emitter.setMaxListeners() to increase limit
    at genericNodeError (node:internal/errors:985:15)
    at wrappedFn (node:internal/errors:539:14)
    at _addListener (node:events:581:17)
    at Interface.addListener (node:events:599:10)
    at Object.start (file:///C:/Users/Boris/WebstormProjects/test/node_modules/wxt/dist/core/keyboard-shortcuts.mjs:18:10)
    at FSWatcher.<anonymous> (file:///C:/Users/Boris/WebstormProjects/test/node_modules/wxt/dist/core/create-server.mjs:78:27)
```

I think, it's related with this issue.

PR #1411, opened 2025-02-08
fix(security): Use `import.meta.hot` for communication with Vite 6.0.9+ (and 5.4.14+) dev server

aklinker1:
### Overview

WXT previously relied on connecting to Vite's dev server manually, instead of using `import.meta.hot`. Vite 6.0.9 introduced several changes that broke WXT's dev mode:

- Vite introduced a randomly generated auth token to ensure of your application can access it
   
   This broke WXT because WXT doesn't know what the token is. While it is possible to get the token from the dev server, it would take quite a bit of refactoring to route the token into the background script during development, so I'm not gonna do that. Instead, WXT now uses `import.meta.hot`.
   
   For MV2, WXT dynamically imports the `vite/client` package from the dev server. For MV3, we're updating dev mode to always set `background.type: "module"` in the manifest.
   
> [!WARNING]
> This is technically a breaking change to dev mode, however, it should have no effect. WXT will continue to bundle the background as IIFE if you don't specify `type: "module"` in your entrypoint.

- CORS: TODO

### Manual Testing

Install the new version from pkg.pr.new and spin up your extension in dev mode:

```sh
pnpm i https://pkg.pr.new/wxt@1411
pnpm dev
```

### Related Issue

- This PR closes #1360
- This PR closes #1362

Issue #1362, opened 2025-01-21
dev websocket connection failed


aianddeng:
### Describe the bug

<img width="1021" alt="Image" src="https://github.com/user-attachments/assets/38857cb3-4487-4b73-9a59-1bd27c04cbe7" />

After executing "run dev", the websocket reported an error, and the content script did not execute successfully, but no error message was reported.

### Reproduction

bunx wxt@latest init

The default template already has this issue.

### System Info

```shell
System:
    OS: macOS 15.3
    CPU: (8) arm64 Apple M1 Pro
    Memory: 515.16 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.9.0 - ~/Library/Caches/fnm_multishells/19945_1737477607928/bin/node
    Yarn: 1.22.22 - ~/Library/Caches/fnm_multishells/19945_1737477607928/bin/yarn
    npm: 10.8.3 - ~/Library/Caches/fnm_multishells/19945_1737477607928/bin/npm
    pnpm: 9.11.0 - ~/Library/Caches/fnm_multishells/19945_1737477607928/bin/pnpm
    bun: 1.1.43 - ~/.bun/bin/bun
  Browsers:
    Chrome: 132.0.6834.83
    Safari: 18.3
  npmPackages:
    wxt: ^0.19.13 => 0.19.24
```

### Used Package Manager

bun

### Validations

- [x] Read the [Contributing Guidelines](https://github.com/wxt-dev/wxt/blob/main/CONTRIBUTING.md).
- [x] Read the [docs](https://wxt.dev/guide/installation.html).
- [x] Check that there isn't [already an issue](https://github.com/wxt-dev/wxt/issues) that reports the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/wxt-dev/wxt/discussions) or join our [Discord Chat Server](https://discord.gg/ZFsZqGery9).
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

appsplash99:
Similar thing for me on chrome:

<img width="758" alt="Image" src="https://github.com/user-attachments/assets/7b8618f7-84f5-43ab-9a0c-5d4401407e65" />

ericlin11354:
Bump, I have this issue as well

aklinker1:
[...]

PR #1497, opened 2025-03-04
fix: Allow vite 6.0.9+

aklinker1:
### Overview

Until #1411 is ready, this allows projects to use newer versions of vite. IT DOES NOT RESOLVE THE SECURITY ISSUE, that's what #1411 is for. But I've ran into some snags there and haven't had much time to work on it, so it hasn't ben merged after a month. So I'm releasing this just to allow people to upgrade.

### Manual Testing

Run the demo extension in dev mode. This PR also upgrades the local version of Vite used during development to `6.2.0`, which would break without this change.

### Related Issue

Related to:
- https://github.com/wxt-dev/wxt/issues/1360
- https://github.com/wxt-dev/wxt/issues/1362

But doesn't resolve them. Just applies an unsafe, temporary workaround.


aklinker1:

PR #1499, opened 2025-03-04
ci: prevent action running on forks

btea:
### Overview
When I fork the repository and synchronize the upstream code, the related actions will fail to run. Therefore, I think the related actions should be prevented from running in the forked repository.

### Related Issue


aklinker1 (approved):
Yup, good change, thanks!

PR #1500, opened 2025-03-04
feat: adding property styles in root document

### LINKED ISSUE #1461: Automatically add styles required on the root document outside shadow root
### Feature Request

In `createShadowRootUi`, some styles, like `@property` styles, must be added outside the shadow root. See https://github.com/tailwindlabs/tailwindcss/issues/15005

WXT should automatically extract these styles from the CSS here and add them outside the shadow root instead.

https://github.com/wxt-dev/wxt/blob/357bbdaec9d610b5f3dd50fc928dc733d6d2701c/packages/wxt/src/client/content-scripts/ui/index.ts#L131-L135

https://github.com/wxt-dev/wxt/discussions/1457

Wait for Tailwind to do something about it?

nishu-murmu:
### Overview

Just extracted the property styles from entryCss text and append it in the root document.
Don't know if this follows standard and consistency. Please let me know.

### Manual Testing

Create components within shadowDOM using `createShadowRootUi` API.

### Related Issue

This PR closes #1461

### DIFF
--- a/packages/wxt/src/client/content-scripts/ui/index.ts
+++ b/packages/wxt/src/client/content-scripts/ui/index.ts
@@ -129,7 +129,15 @@ export async function createShadowRootUi<TMounted>(
 ): Promise<ShadowRootContentScriptUi<TMounted>> {
   const css = [options.css ?? ''];
   if (ctx.options?.cssInjectionMode === 'ui') {
-    const entryCss = await loadCss();
+    let entryCss = await loadCss();
+    let propertyCss = entryCss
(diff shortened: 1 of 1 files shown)


aklinker1 (changes requested):
This might work for tailwind, where the `@property` declarations are last, but not for any other project that uses it.

I'd prefer a generic solution.

mengxi-ream:
> This might work for tailwind, where the `@property` declarations are last, but not for any other project that uses it.
> 
> I'd prefer a generic solution.

UnoCSS declare `@property` in the front

PR #1501, opened 2025-03-04
chore: Upgrade to `buildc@2.0.0-alpha`

aklinker1:
### Overview

2x as fast as before at checking for cached builds and scheduling new ones.


aleclarson:
Out of curiosity, how does this compare to TypeScript's incremental build feature?

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#faster-subsequent-builds-with-the---incremental-flag

aklinker1:
> Out of curiosity, how does this compare to TypeScript's incremental build feature?
> 
> https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#faster-subsequent-builds-with-the---incremental-flag

Unrelated, since we don't use TS to build packages. WXT uses [`unbuild`](https://unjs.io/packages/unbuild) to transpile the TS source code to `.d.ts` for NPM.

If we did use TS, it might do something similar if you can configure TS references to all the packages correctly. But I've never used `tsc` to emit JS, just for type checking. So I'm not 100% sure that's correct.

This package computes the workspace dependency tree and builds package dependencies. It's more similar to turborepo/nx/lerna in that regard.

PR #1507, opened 2025-03-08
fix: add missing timer clear methods to ContentScriptContext

kikutadev:
### Overview

Added missing timer clear methods (`clearTimeout`, `clearInterval`, `cancelAnimationFrame`, `cancelIdleCallback`) to the ContentScriptContext class. The current class provides corresponding timer creation methods (`setTimeout`, `setInterval`, etc.) but lacks methods to clear them, causing TypeScript compilation errors.

### Manual Testing

Test with code using ContentScriptContext like:
```typescript
const timerId = ctx.setTimeout(() => {
  console.log('This should not run');
}, 1000);

ctx.clearTimeout(timerId); // This should work properly
```

### Related Issue

No related issue.

### DIFF
--- a/packages/wxt/src/client/content-scripts/content-script-context.ts
+++ b/packages/wxt/src/client/content-scripts/content-script-context.ts
@@ -1,6 +1,6 @@
-import { ContentScriptDefinition } from '../../types';
 import { browser } from 'wxt/browser';
 import { logger } from '../../sandbox/utils/logger';
+import { ContentScriptDefinition } from '../../types';
 import { WxtLocationChangeEvent, getUniqueEventName } from './custom-events';
 import { createLocationWatcher } from './location-watcher';
 
@@ -161,6 +161,38 @@ export class ContentScriptContext implements AbortController {
 
     this.onInvalidated(() => cancelIdleCallback(id));
     return id;
+  }
+
+  /**
+   * Clears a timeout set by ctx.setTimeout.
+   */
+  clearTimeout(id: number | null): void {
+    if (id !== null) {
+      clearTimeout(id);
+    }
+  }
+
+  /**
+   * Clears an interval set by ctx.setInterval.
+   */
+  clearInterval(id: number | null): void {
+    if (id !== null) {
+      clearInterval(id);
+    }
+  }
+
+  /**
+   * Cancels an animation frame request set by ctx.requestAnimationFrame.
+   */
+  cancelAnimationFrame(id: number): void {
+    cancelAnimationFrame(id);
+  }
+
+  /**
+   * Cancels an idle callback request set by ctx.requestIdleCallback.
+   */
+  cancelIdleCallback(id: number): void {
+    cancelIdleCallback(id);
   }
 
   /**


aklinker1:
Why not just call `clearTimeout`?

kikutadev:
@aklinker1 
Thank you for reviewing my PR. 

I understand that directly using global functions like clearTimeout would technically work, but I proposed these additions for the following reasons:

1. API Consistency: The context already provides methods like setTimeout and setInterval, so developers would naturally expect corresponding clear methods to be available as well.

2. Documentation Clarity: While the documentation shows examples of using ctx.setTimeout(), it doesn't mention how to clear these timers, which can be confusing for developers. (I personally was confused when I encountered type errors using ctx.clearTimeout(), and ended up creating a d.ts file to wrap wxt/client).

3. Type Safety: I was concerned that getting a timer ID from ctx.setTimeout() and then trying to use the global clearTimeout might cause type confusion or inconsistencies.

While I acknowledge that not adding these clear methods wouldn't necessarily cause problems, I believe making these additions would create a more intuitive and consistent developer experience. That said, I'm open to your thoughts on the best approach.

packages/wxt/src/client/content-scripts/content-script-context.ts
```diff
@@ -163,6 +163,38 @@ export class ContentScriptContext implements AbortController {
...
   }
 
+  /**
+   * Clears a timeout set by ctx.setTimeout.
+   */
+  clearTimeout(id: number | null): void {
+    if (id !== null) {
+      clearTimeout(id);
+    }
+  }
```
aklinker1, line 173: These should be `| undefined` if we decide to keep them. And we don't need an if-statement either.

```suggestion
  clearTimeout(id: number | undefined): void {
    clearTimeout(id);
  }
```

Same applies to `clearInterval`.

aklinker1 (changes requested):
Sorry for the delay in getting back to you, was focused on getting v0.20 out the door!

> API Consistency: The context already provides methods like setTimeout and setInterval, so developers would naturally expect corresponding clear methods to be available as well.

The purpose of the context object is to provides wrappers around async APIs that need to be canceled when the context is invalidated - the clear functions don't fit that purpose. I see where  you're coming from around consistencies, but I think this could be resolved by updating the docs to use more clear wording and include examples for how to stop them.

> Documentation Clarity: While the documentation shows examples of using ctx.setTimeout(), it doesn't mention how to clear these timers, which can be confusing for developers. (I personally was confused when I encountered type errors using ctx.clearTimeout(), and ended up creating a d.ts file to wrap wxt/client).

We can update the docs and JSDocs to mention how to clear the timers.

> Type Safety: I was concerned that getting a timer ID from ctx.setTimeout() and then trying to use the global clearTimeout might cause type confusion or inconsistencies.

This should be fixed by #1531, so this shouldn't be a concern anymore.

kikutadev:
Thank you for your feedback. I understand your reasoning about the context object's purpose and agree that documentation improvements would be better than adding these methods.

I'm happy to close this PR since the type safety concerns will be addressed in #1531 .

Sorry for not fully understanding the project's contribution guidelines. This was my first contribution attempt. Would you prefer I close this PR?

aklinker1:
I've updated the docs in 030f23d2e9044203ea824e70995535bb60d64978

PR #1496, opened 2025-03-03
chore(deps): Upgrade to Vite 6 and related dependencies

aklinker1:
Dependencies upgraded:

| Dependency | Version Change |
| --- | --- |
| `vite`                      | `5` &rarr; `6` (monorepo only, you can still use vite 5+ in your own projects) |
| `@vitejs/plugin-vue`        | `5.2.0` &rarr; `5.2.1` |
| `@vitest/coverage-v8`       | `2.1.4` &rarr; `3.0.7` |
| `c12`                       | `1.11.2` &rarr; `3.0.2` |
| `chokidar`                  | `3.6.0` &rarr; `4.0.3` |
| `esbuild`                   | `0.21.5` &rarr; `0.25.0` |
| `jiti`                      | `1.21.6` &rarr; `2.4.2` |
| `tsx`                       | `4.15.7` &rarr; `4.19.3` |
| `unbuild`                   | `2.0.0` &rarr; `3.5.0` |
| `unocss`                    | `0.64.0` &rarr; `66.0.0`  (monorepo only, you can still use v0.64+ in your own projects) |
| `vitepress`                 | `1.5.0` &rarr; `1.6.3` |
| `vitest`                    | `2.1.4` &rarr; `3.0.7` |
| `vitest-mock-extended`      | `2.0.2` &rarr; `3.0.1` |
| `vitest-plugin-random-seed` | `1.1.0` &rarr; `1.1.1` |

This replaces:

- https://github.com/wxt-dev/wxt/pull/1487
- https://github.com/wxt-dev/wxt/pull/1483
- https://github.com/wxt-dev/wxt/pull/1481
- https://github.com/wxt-dev/wxt/pull/1480
- https://github.com/wxt-dev/wxt/pull/1477
- https://github.com/wxt-dev/wxt/pull/1476
- https://github.com/wxt-dev/wxt/pull/1475
- https://github.com/wxt-dev/wxt/pull/1472

This closes:

- https://github.com/wxt-dev/wxt/security/dependabot/21
- https://github.com/wxt-dev/wxt/security/dependabot/22

PR #1508, opened 2025-03-08
chore(deps): Bump unjs ecosystem

lishaduck:
### Overview

Following https://antfu.me/posts/move-on-to-esm-only, the unjs ecosystem is now ESM only.
This means that we need to extend our range of supported package versions.

Notably, I didn't upgrade ohash, because v2 removes `murmur` in favor of `digest`. (Which is ironic, because that's why I started this PR).

Other notes:
- We're already ESM-only.
- #1501 will also help with this, as buildc is also pulling in old unjs versions (but only internally, of course).
- There are other outdated deps, but I only did unjs because they were partially updated in #1496.
- unjs now offers an alternative to `import-meta-resolve` called `exsolve`. I'd leave that for another PR.

### Manual Testing

Shouldn't be needed, these are all just ESM-only migrations.

### Related Issue

None

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -181,8 +181,8 @@ catalogs:
       specifier: ^3.0.0
       version: 3.0.0
     nypm:
-      specifier: ^0.6.0
-      version: 0.6.0
+      specifier: ^0.3.12
+      version: 0.3.12
     ohash:
       specifier: ^1.1.4
       version: 1.1.4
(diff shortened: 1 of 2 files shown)


lishaduck:
> Shouldn't be needed, these are all just ESM-only migrations.

I take that back, nypm now supports Deno, and the way we're using nypm makes that a breaking change.
I gave a shot at updating, but Deno's so different I don't know how we want to handle it.

aklinker1 (approved):
Thanks for the PR!

> I take that back, nypm now supports Deno, and the way we're using nypm makes that a breaking change.
I gave a shot at updating, but Deno's so different I don't know how we want to handle it.

Yeah, Deno is too different from the rest of the JS ecosystem, you can kinda use it in a project, but it's hard: https://github.com/wxt-dev/wxt/issues/1307

The NYPM stuff is definitely problematic for it :/

Timeraa (approved):

PR #1511, opened 2025-03-10
feat: remove logs and add onMigrationComplete callback

### LINKED ISSUE #1503: option to use old logger behavior
### Feature Request

Hi I saw the logger behavior changed from an utility that check the environment, if the environment is product no message logger to just javascript `console.*`.

It makes sense for `error` and `warning` but for debug/log I would rather not having them logged in production for example: https://github.com/wxt-dev/wxt/blob/main/packages/storage/src/index.ts#L406-L409, in production I don't want to have this logged as it might be used in content scripts, and some of my settings are secret which I don't want to take the risk to expose to the website that my content script runs on.

What do you think to revert the old log behaviors only for `debug` (perhaps `log` also ?). Of course logs in the dev server etc will remind the same.

It might not be a big issue, but for example sentry will overrid the `console.*` based on websites owners config.

"N/A"

I am doing a work around at the moment like

```ts
export const settingsStorage = (() => {
  const debug = console.debug;
  if (import.meta.env.PROD) {
    console.debug = () => {};
  }
  const store = storage.defineItem<Settings>(SETTINGS_STORAGE_ITEM_KEY, {
    version: 2,
    init: () => defaultSettings,
    migrations: {
      2: (setting: SettingsV1): SettingsV2 => {
        return defu(setting, defaultSettingsV2);
      },
    },
  });
  return store;
})();
```

RayGuo-ergou:
### Overview

As mentioned #1503, the logs in storage should be cleared, and just this morning, an idea came to my mind that we can introduce a new option `onMigrationComplete` that let user decide if they want to do anything when migration just completed.

I have not added anything in the document as I assume the types should be automatically generated. 

### Manual Testing

I have added a unit test case for the new option `onMigrationComplete`

### Related Issue

This PR closes #1503

### DIFF
--- a/packages/storage/src/__tests__/index.test.ts
+++ b/packages/storage/src/__tests__/index.test.ts
@@ -705,6 +705,30 @@ describe('Storage Utils', () => {
         expect(migrateToV3).toBeCalledWith(4);
       });
 
+      it('should call onMigrationComplete callback function if defined', async () => {
+        await fakeBrowser.storage.local.set({
+          count: 2,
+          count$: { v: 1 },
+        });
+        const migrateToV2 = vi.fn((oldCount) => oldCount * 2);
+        const migrateToV3 = vi.fn((oldCount) => oldCount * 3);
+        const onMigrationComplete = vi.fn((count, _v) => count);
+
+        storage.defineItem<number, { v: number }>(`local:count`, {
+          defaultValue: 0,
+          version: 3,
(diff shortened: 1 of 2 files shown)


aklinker1 (changes requested):
Can you split this into 2 PRs?

RayGuo-ergou:
sure thing!

PR #1513, opened 2025-03-11
feat(storage): Add `debug` option to enable migration logs

RayGuo-ergou:
### Overview

Part of https://github.com/wxt-dev/wxt/pull/1511, to remove console log/debug from the storage.

### Related Issue

This PR closes https://github.com/wxt-dev/wxt/issues/1503

### DIFF
--- a/packages/storage/src/__tests__/index.test.ts
+++ b/packages/storage/src/__tests__/index.test.ts
@@ -883,6 +883,39 @@ describe('Storage Utils', () => {
           { migratedValue: expect.any(Number) },
         );
       });
+      it('should not print migration logs if debug option is undefined or false', async () => {
+        await fakeBrowser.storage.local.set({
+          count: 2,
+          count$: { v: 1 },
+        });
+        await fakeBrowser.storage.local.set({
+          count2: 2,
+          count2$: { v: 1 },
+        });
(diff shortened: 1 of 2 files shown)


Timeraa (commented):
Personally think that the logs should still be sent in dev mode since that makes debugging easier

RayGuo-ergou:
Hi thanks for your feedback, but I do think the logging should be a opt-in behavior, and that's the reason for #1514, if the user wants to logging the messages they can still do but the correct approach is that the user cannot disable the logging if don't want to, for instance I have some secret that I don't want to be logged even in dev.

And without this logging, things still very clear what is going on, because:

1. Error will be thrown if anything goes wrong.
2. There's a metadata tells the user if the migration success

aklinker1:
Should we add an option then instead of removing them entirely?

RayGuo-ergou:
good idea. a debug flag seems easy simple and just work.

RayGuo-ergou:
i have changed to use `debug` flag and default to `false` and i make it as a `defineItemOption` level, thinking in the future can add more debug logging for `defineItem` not only limited to migration.

Timeraa (approved):

aklinker1 (approved):
Thanks!

RayGuo-ergou:
@aklinker1 conflict resolved

PR #1514, opened 2025-03-11
feat(storage): Add `onMigrationComplete` callback

### LINKED ISSUE #1503: option to use old logger behavior

RayGuo-ergou:
### Overview

Following https://github.com/wxt-dev/wxt/pull/1513 part of https://github.com/wxt-dev/wxt/pull/1511.

This PR introduces a new option `onMigrationComplete` to the storage `defineItem` that has two parameters, the migrated item and the migrated version.

Note: there might be conflicts and I will resolve when one is merged.

### Manual Testing

I have added a unit test case for the new option `onMigrationComplete`

### Related Issue

This PR closes #1503

### DIFF
--- a/packages/storage/src/index.ts
+++ b/packages/storage/src/index.ts
@@ -872,7 +872,7 @@ export interface WxtStorageItemOptions<T> {
   /**
    * A callback function that runs on migration complete.
    */
-  onMigrationComplete?: (storage: T, targetVersion: number) => void;
+  onMigrationComplete?: (migratedValue: T, targetVersion: number) => void;
 }
 
 export type StorageAreaChanges = {
--- a/packages/storage/src/index.ts
+++ b/packages/storage/src/index.ts
@@ -410,9 +410,7 @@ function createStorage(): WxtStorage {
           { key: driverKey, value: migratedValue },
           { key: driverMetaKey, value: { ...meta, v: targetVersion } },
         ]);
-        if (onMigrationComplete !== undefined) {
-          onMigrationComplete(migratedValue, targetVersion);
-        }
+        onMigrationComplete?.(migratedValue, targetVersion);
         console.debug(
           `[@wxt-dev/storage] Storage migration completed for ${key} v${targetVersion}`,
           { migratedValue },
--- a/packages/storage/src/index.ts
+++ b/packages/storage/src/index.ts
@@ -359,7 +359,11 @@ function createStorage(): WxtStorage {
     defineItem: (key, opts?: WxtStorageItemOptions<any>) => {
       const { driver, driverKey } = resolveKey(key);
 
-      const { version: targetVersion = 1, migrations = {} } = opts ?? {};
+      const {
+        version: targetVersion = 1,
+        migrations = {},
+        onMigrationComplete,
+      } = opts ?? {};
       if (targetVersion < 1) {
         throw Error(
           'Storage item version cannot be less than 1. Initial versions should be set to 1, not 0.',
@@ -406,6 +410,9 @@ function createStorage(): WxtStorage {
           { key: driverKey, value: migratedValue },
           { key: driverMetaKey, value: { ...meta, v: targetVersion } },
         ]);
+        if (onMigrationComplete !== undefined) {
+          onMigrationComplete(migratedValue, targetVersion);
+        }
         console.debug(
           `[@wxt-dev/storage] Storage migration completed for ${key} v${targetVersion}`,
           { migratedValue },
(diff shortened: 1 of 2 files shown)


Timeraa (changes requested):
Since we use `hookable` in many places inside WXT, I think it would be good to use it there too to match the other usages in WXT.

RayGuo-ergou:
Just had a quick look at the code base and I saw one hookable instance for `wxt` but for this storage module I don't think it should add under the same score(`wxt`). So I created another hookable instance only for `defineItems`

aklinker1 (approved):
Looks good, thanks for splitting this up! Using hookable sounds good to me too.

aklinker1:
Ehhh, after thinking about it more, @Timeraa I'm not so sure I want to use `hookable` here. It's not like WXT's module system, this is really just one callback defined in options. I think I would rather not bloat the package with another dependency for this case. We aren't using `hookable` in client code, so this would increase the client bundle size.

Timeraa:
> Ehhh, after thinking about it more, @Timeraa I'm not so sure I want to use `hookable` here. It's not like WXT's module system where multiple consumers are hooking into lots of different events multiple times, this is really just one callback defined in options. I think I would rather not bloat the package with another dependency for this case. We aren't using `hookable` in client code right now, so this would increase the final bundle size.

Yeah, was just a thought I had, though if we add more features it would probs be good, like hooks for when migration is about to start etc. yk?

Timeraa (approved):

aklinker1 (changes requested):
@RayGuo-ergou Could you revert the `hookable` change? I think we've decided we don't want to use it for this right now.

Timeraa:
> > Yeah, was just a thought I had, though if we add more features it would probs be good, like hooks for when migration is about to start etc. yk?
> 
> Yeah, but we could just add an option for those, it wouldn't be much different.

Yeah true in this context it doesn't make much sense

Timeraa (approved):

packages/storage/src/index.ts
```diff
@@ -406,6 +410,9 @@ function createStorage(): WxtStorage {
           { key: driverKey, value: migratedValue },
           { key: driverMetaKey, value: { ...meta, v: targetVersion } },
         ]);
+        if (onMigrationComplete !== undefined) {
+          onMigrationComplete(migratedValue, targetVersion);
+        }
         console.debug(
```
aklinker1, line 416:
```suggestion
        onMigrationComplete?.(migratedValue, targetVersion);

        console.debug(
```

packages/storage/src/index.ts
```diff
@@ -862,6 +869,10 @@ export interface WxtStorageItemOptions<T> {
    * A map of version numbers to the functions used to migrate the data to that version.
    */
   migrations?: Record<number, (oldValue: any) => any>;
+  /**
+   * A callback function that runs on migration complete.
+   */
+  onMigrationComplete?: (storage: T, targetVersion: number) => void;
```
aklinker1, line 875:
```suggestion
  onMigrationComplete?: (migratedValue: T, targetVersion: number) => void;
```

aklinker1 (approved):
Awesome, thanks. A couple small changes then I'll get this merged!

packages/storage/src/index.ts
```diff
@@ -406,6 +410,9 @@ function createStorage(): WxtStorage {
           { key: driverKey, value: migratedValue },
           { key: driverMetaKey, value: { ...meta, v: targetVersion } },
         ]);
+        if (onMigrationComplete !== undefined) {
+          onMigrationComplete(migratedValue, targetVersion);
+        }
```
aklinker1, line 415:
```suggestion
        onMigrationComplete?.(migratedValue, targetVersion);
```

aklinker1 (changes requested):
Woops, meant to request changes

aklinker1 (approved):

Issue #1209, opened 2024-11-26
public should be relative to the rootDir instead of srcDir


2wheeh:
### Feature Request
Currently, `publicDir` is relative to `srcDir`, meaning when `srcDir` is set to `'src'`, public files must be placed in `src/public`. This feels unintuitive since public assets typically live at the root level in most web projects.

Having public assets in ``src/`` creates cognitive overhead since:

- Most developers expect public files to be at the root level. 
- It breaks convention with other popular frameworks/tools.
- Static assets conceptually don't belong with source code.

#### Is your feature request related to a bug?

N/A

### What are the alternatives?

Make `publicDir` relative to `rootDir` instead of `srcDir`, allowing the public directory to be placed at root level (e.g., `<rootDir>/<publicDir>`) while source files are organized in `src/`.

### Additional context

This aligns with conventions from other popular web frameworks (Vue, React, Nextjs etc.) where public assets are typically placed at the root level for better project organization and clarity.

2wheeh:
and the comments (jsDoc), it is `"${config.root}/public"` instead of `"${config.srcDir}/public"`.

2wheeh:
```
## v0.4.0 &rarr; v0.5.0
### Revert Move Public Directory
Change default `publicDir` to from `<rootDir>/public` to `<srcDir>/public`.
## v0.3.0 &rarr; v0.4.0
### Update Default Path Aliases
Use relative path aliases inside `.wxt/tsconfig.json`.
## v0.2.0 &rarr; v0.3.0
### Move Public Directory
Change default `publicDir` to from `<srcDir>/public` to `<rootDir>/public`.
```

I found this updated and reverted... 🫤

aklinker1:
WXT is inspired by Nuxt, and follows its conventions when it makes sense. Nuxt 3's `public` directory was relative to the `srcDir`. That is why WXT's has also always been relative to its `srcDir`. The 0.4/0.5 changes were me thinking I had read their docs wrong, but then realizing that no, I was right `public` was inside `srcDir`, so I reverted the change.

Nuxt 4 plans on changing the `public` directory to be relative to the `root` directory. However in their case, it's to extract common code out of the new `app` and `server` directories, to which public files belong to both. Read more about their decision here:

https://github.com/nuxt/nuxt/issues/26444

[...]

PR #1519, opened 2025-03-17
Fix comment for default publicDir location

namukang:
### Overview

The documentation for `publicDir` says that the default value is `${config.root}/public` but it should be `${config.srcDir}/public`

### DIFF
--- a/packages/wxt/src/types.ts
+++ b/packages/wxt/src/types.ts
@@ -29,7 +29,7 @@ export interface InlineConfig {
   /**
    * Directory containing files that will be copied to the output directory as-is.
    *
-   * @default "${config.root}/public"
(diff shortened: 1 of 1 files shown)


2wheeh:
FYI, It's updating to rootDir indeed https://github.com/wxt-dev/wxt/issues/1209#issuecomment-2501456738

namukang:
> FYI, It's updating to rootDir indeed [#1209 (comment)](https://github.com/wxt-dev/wxt/issues/1209#issuecomment-2501456738)

Oh great! I was a bit puzzled by why it was in the src dir rather than the root so this context is nice to have. Glad to hear that it’ll be changing to the root dir.

I’ll close this out since this will be reverted anyway with the upcoming change.

PR #1529, opened 2025-03-21
feat: tolerate module resolution errors on startup

aleclarson:
### Overview

Avoid crashing the dev server if a broken import exists.

Screenshot 2025-03-21 at 13 53 12

**Unrelated to this PR:** You can see the suggestion to install `vite-tsconfig-paths`, but as I understand it, WXT doesn't recommend using that package; preferring that the user sets `alias` in `wxt.config.ts` instead.

### Manual Testing

1. Introduce a broken `import` statement
2. Start the dev server


packages/wxt/src/core/create-server.ts
```diff
@@ -162,16 +167,30 @@ async function createServerInternal(): Promise<WxtDevServer> {
...
+      if (err instanceof SyntaxError && isBabelSyntaxError(err)) {
+        logBabelSyntaxError(err);
+        wxt.logger.info('Waiting for syntax error to be fixed...');
+        watchFile = err.id;
+      } else if (err instanceof Error && isModuleNotFoundError(err)) {
+        const details = parseModuleNotFoundError(err);
+        if (details && details.importer.startsWith('/@fs/')) {
+          wxt.logger.error(err.message);
+          wxt.logger.info('Waiting for import specifier to be fixed...');
+          watchFile = details.importer.slice(5);
```
aleclarson, line 181: This isn't *perfect*, since the broken import may not be the fault of the importer. For example, it could be a configuration issue with `tsconfig.json` or something.

But it's not a big deal, IMO, since the developer can restart the `dev` process or save the importer's file if necessary.

aleclarson: Before this can be merged, we should figure out how we can allow for saving the `wxt.config.ts` file to recreate the dev server as expected. Currently, it's just ignored, which is not a good DX.

aleclarson:
Do not merge yet. Looking into import errors coming from Vite's load fallback...

aleclarson:
Still not ready. Handling the `Rollup failed to resolve import` error now!

packages/wxt/src/core/utils/building/detect-dev-changes.ts
aklinker1, line 1: I like these changes, but we need to update the tests in `detect-dev-changes.test.ts`. It's a very well tested function with 100% coverage. So we need to add coverage for the two new cases: no build output and error files.

packages/wxt/src/core/utils/error.ts
```diff
@@ -0,0 +1,148 @@
...
+  path: string;
+}
+
+export function isViteLoadFallbackError(
+  error: VitePluginError,
+): error is ViteLoadFallbackError {
+  return error.plugin === 'vite:load-fallback' && error.hook === 'load';
+}
+
+export function resolveErrorFiles(err: Error): string[] {
```
aklinker1, line 88: And we should definintely add tests for this function. Will be nice to have example errors we can refer to to understand what exactly what is happening better.

aklinker1 (changes requested):
I like the refactor in this PR! Only requested change is to update/add tests.

aleclarson:
@aklinker1 I updated existing tests for `detectDevChanges`. I wasn't sure what the test for the new `errorFiles` argument should look like, so I created a stub for you to fill out.

> And we should definintely add tests for this function (`resolveErrorFiles`). Will be nice to have example errors we can refer to to understand what exactly what is happening better.

Do you mean unit tests with dummy errors resembling real ones, or are you asking for integration tests that reproduce the errors using Babel, Rollup, Vite, etc?

aklinker1:
> Do you mean unit tests with dummy errors resembling real ones, or are you asking for integration tests that reproduce the errors using Babel, Rollup, Vite, etc?

I was just taking about unit tests, but integration tests would probably be better for this.

aklinker1:
I know you've got other old PRs as well, I'll try and get to all of them this week

PatrykKuniczak:

PatrykKuniczak:
@aleclarson If you're eager to reopen and go on, on any PR, which I closed, feel free 😄

Issue #1459, opened 2025-02-24
Node types are pulled accidentally by importing vite types


minht11:
### Describe the bug

See the same issue in different package https://github.com/Anber/wyw-in-js/issues/33

I was investigating why node types are included in my project and came to cause being `.wxt/types/imports.d.ts` file, which imports for example:
` const ContentScriptContext: typeof import('wxt/client')['ContentScriptContext']`

And that file uses:

`import { ContentScriptDefinition } from '../../types'`

Which in turn imports:
`import type * as vite from 'vite';`

If we look at vite, it uses `<reference types="node" />` which means any file importing it also gets the same types.

<img width="867" alt="Image" src="https://github.com/user-attachments/assets/df4f542f-580f-467a-9ade-a4ccf581f10d" />

The fix would be split vite types from client ones, so two are not mixed.

### Reproduction

n/a

### Steps to reproduce

Try using `const timeout: number = setTimeout(() => {})`

### System Info

```shell
"wxt": "^0.19.27"
```

### Used Package Manager

pnpm

### Validations

- [x] Read the [Contributing Guidelines](https://github.com/wxt-dev/wxt/blob/main/CONTRIBUTING.md).
- [x] Read the [docs](https://wxt.dev/guide/installation.html).
- [x] Check that there isn't [already an issue](https://github.com/wxt-dev/wxt/issues) that reports the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/wxt-dev/wxt/discussions) or join our [Discord Chat Server](https://discord.gg/ZFsZqGery9).
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

aklinker1:
This was intended, but I can see why it would cause problems. Will try to fix post-v1.0!

garretwilson:
I just ran into this in a background script, which has no `window` object but still supports `setTimeout()` and the like. The following puts the error checker into a frenzy, because apparently it thinks this is the Node `setTimeout()` method, returning a `Timeout`:

```typescript
let timeoutId: number | null;
timeoutId = setTimeout(() => foo.bar(), 60*1000);
```

The work around is `as any as number`, like the old Java cast `(NoReallyMyOtherType)(Object)foo` (like the even old older C/C++ reference/dereference trick—so long ago I've forgotten how it works):

```typescript
[...]

PR #1085, opened 2024-10-19
staging: v0.20.0

aklinker1:
This PR will be rebased onto main once v0.20.0 is ready to be released. It contains all the breaking changes queued up for that release. No ETA for when this version will be released.

The goal is for this to be the last round of breaking changes before v1.0 is released.

- [x] #1084 closes #868
- [x] #1086 closes #1010
- [x] #1087
- [x] #1180 
- [x] #1181
- [x] #1182 closes #297
- [x] ~~#1214~~ Vite 6 works just fine, I don't see a reason to make these changes in v0.20, should be able to do them as non-breaking
- [x] #1221
- [x] #1216 closes #1209
- [x] #1261
- [x] #1258
- [x] #1271
- [x] #1269 closes #1195
- [x] #1270
- [x] #1311
- [x] #1310
- [x] #1530 closes #1213
- [x] #1530 closes #1116
- [x] Cleanup commits to non-WXT packages to prevent breaking changes to other packages.
- [x] ~~Remove `@types/chrome` from templates and add to changelog~~ https://github.com/wxt-dev/wxt/pull/1085/commits/80cc362ee46d8f22efc389256d452454de2fe94a
- [x] Revert version change

After merge:

- [ ] Release `wxt`
- [ ] Release `@wxt-dev/analytics`
- [ ] Cleanup changelogs
- [ ] Leave comments on all issues closed
- [ ] Discord announcement

PR #1531, opened 2025-03-22
fix: avoid leaking `@types/node` into client code

aleclarson:
### Overview

The `defineWxtPlugin` function is referenced by the generated `.wxt/types/imports.d.ts` module, which is used by client code. Therefore, we need to avoid importing from `wxt/src/types.ts`, as it depends on Node.js-specific packages (like `vite`). This inadvertently introduces Node.js globals into the extension's client environment, which is undesirable (e.g. it breaks the return type of `setTimeout`).

#### Edit 1

Updated other "offending" imports. Split out the option-related types (into `option-types.ts`) and the client-related types (into `client-types.ts`).

### Manual Testing

Without this change, you should see `setTimeout` return a `NodeJS.Timeout` object.

After this change, `setTimeout` will return a number as expected.

### Related Issue

https://github.com/wxt-dev/wxt/issues/1459

### DIFF
--- a/packages/wxt/src/client-types.ts
+++ b/packages/wxt/src/client-types.ts
@@ -1,0 +1,55 @@
+import type {
+  BackgroundEntrypointOptions,
+  BaseEntrypointOptions,
+  IsolatedWorldContentScriptEntrypointOptions,
+  MainWorldContentScriptEntrypointOptions,
+} from './option-types';
+import type { ContentScriptContext } from './utils/content-script-context';
+
+export type WxtPlugin = () => void;
+
+export interface IsolatedWorldContentScriptDefinition
+  extends IsolatedWorldContentScriptEntrypointOptions {
+  /**
+   * Main function executed when the content script is loaded.
+   *
+   * When running a content script with `browser.scripting.executeScript`,
+   * values returned from this function will be returned in the `executeScript`
+   * result as well. Otherwise returning a value does nothing.
+   */
+  main(ctx: ContentScriptContext): any | Promise<any>;
+}
+
+export interface MainWorldContentScriptDefinition
+  extends MainWorldContentScriptEntrypointOptions {
+  /**
+   * Main function executed when the content script is loaded.
+   *
+   * When running a content script with `browser.scripting.executeScript`,
+   * values returned from this function will be returned in the `executeScript`
+   * result as well. Otherwise returning a value does nothing.
+   */
+  main(): any | Promise<any>;
+}
+
+export type ContentScriptDefinition =
+  | IsolatedWorldContentScriptDefinition
+  | MainWorldContentScriptDefinition;
+
+export interface BackgroundDefinition extends BackgroundEntrypointOptions {
+  /**
+   * Main function executed when the background script is started. Cannot be async.
+   */
+  main(): void;
+}
+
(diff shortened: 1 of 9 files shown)


aleclarson:
I can retarget this to #1085 if you prefer. LMK!

Timeraa:
> I can retarget this to #1085 if you prefer. LMK!

I think that would be better as @aklinker1 should be merging 0.20.0 very soon as he rebased everything from main into it and we've been testing it

Timeraa:
0.20.0 is released and out so make sure to rebase your changes on main!

aleclarson:
Rebased ✅

aklinker1:
I still see `NodeJS.Timeout` in the demo extension after building WXT:

Screenshot 2025-04-09 at 10 33 34 AM

Isn't this... impossible to do? For example, you have to import `import { defineConfig } from 'wxt'` in the `wxt.config.ts` file. Doesn't that mean node types will always be referenced and included in your project?

aleclarson:
> Isn't this... impossible to do? For example, you have to import `import { defineConfig } from 'wxt'` in the `wxt.config.ts` file. Doesn't that mean node types will always be referenced and included in your project?

You can definitely avoid it. The key is to use multiple `tsconfig.json` files. Here's how I do it:

- `/tsconfig.json` used by default (e.g. `*.config.ts` modules and maybe some `/scripts/*.ts` modules)
  - any default `compilerOptions` I use
  - `types: ["node"]`
- `/src/tsconfig.json` used by `/src/**/*` modules
  - extends `../tsconfig.json`
  - `include: ["./"]` (to prevent config/script modules from being pulled in)
  - `types: []` (to prevent `@types/node` from being pulled in; you can add others as necessary, like `vite/client`)

This is off the top of my noggin, so could be missing a detail. I assume there isn't a demo with this setup already?

aleclarson:
I would love to get this merged. Let me know what the next step is.

aklinker1:
Sorry, been busy and it's too easy to just go down the list of recent notifications and forget about these older ones. But I'm catching up!

You are definitely right about using multiple TSConfig files. Nuxt supports projects using one tsconfig file or multiple by generating 4 different tsconfig files in the `.nuxt` dir:

- `tsconfig.json` - extend this to use a simple, single tsconfig file setup
- `tsconfig.base.json` - extend this to split the app/server dirs into separate TS projects. It has references to the 2 other config file below that configure their respective directories.
- `tsconfig.app.json`
- `tsconfig.server.json`

> Names might be wrong, I'm typing this up by memory.

I'd like to support both approaches as well.

WXT generates the first one today, but we could generate the other 3. So I think the full solution here is, in addition to the changes already in this PR, update the `generate-wxt-dir.ts` file to generate the 3 other config files.

I'd recommend looking at a real nuxt project if you've never seen their setup before. A basic starter will have them:

```sh
pnpm create nuxt@latest <project-name>
pnpm nuxt prepare
ls .nuxt | grep tsconfig
```

The new tsconfig file generation can come in a separate PR - I need to re-review your changes here since it's been a while, but we'll get this merged soon! But it's past midnight here and I have some early meetings in the morning... So I'll re-review it tomorrow after work

PatrykKuniczak:
@aleclarson Are you eager to go on with that?

PatrykKuniczak:

PR #1550, opened 2025-03-31
docs: Fix knowledge file generation

aklinker1:
### Overview

The blog with the new RSS feed was overwriting the `buildEnd` callback setup by the knowledge plugin, so the knowledge files were never generated.

I also added required config for the new blog pages to be added to the knowledge files correctly.

### Related Issue

https://discord.com/channels/1212416027611365476/1296842665249472614/1356324259344941076

image

### DIFF
--- a/docs/.vitepress/config.ts
+++ b/docs/.vitepress/config.ts
@@ -1,4 +1,4 @@
-import { defineConfig } from 'vitepress';
+import { DefaultTheme, defineConfig } from 'vitepress';
 import typedocSidebar from '../api/reference/typedoc-sidebar.json';
 import {
   menuGroup,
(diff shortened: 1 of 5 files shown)


docs/.vitepress/config.ts
```diff
@@ -118,6 +128,9 @@ export default defineConfig({
       md.use(footnote);
       md.use(groupIconMdPlugin);
     },
+    languageAlias: {
+      mjs: 'js',
+    },
```
aklinker1, line 133: There was also a warning during the build process that this resolves.

packages/wxt/typedoc.json
```diff
@@ -15,7 +15,6 @@
     "src/utils/define-wxt-plugin.ts",
     "src/utils/match-patterns.ts",
     "src/utils/storage.ts",
-    "src/browser.ts",
```
aklinker1, line 18: Including this in the API docs was causing the entire `@types/chrome` JSDoc to be included, which was taking a long time an inflating the knowledge files. So I've removed this page from the API reference for now.

PR #1558, opened 2025-04-02
fix: add JSDoc type annotation to auto-imports for ESlint

nostrorom:
### Overview

Add type annotation for auto-imports (ESLint 9). If typechecked, `eslint.config.js` may pop an error:

```js
      The types of 'languageOptions.sourceType' are incompatible between these types.
        Type 'string' is not assignable to type 'SourceType | undefined'.ts(2322)
```

### Manual Testing

On a project running checks on JS, IDE should not show squiggly lines

### Related Issue

No related issue


aklinker1:
Looks like a test snapshot needs updated. Run `pnpm test` to see the changes. If they're OK, type `u` to update them.

PR #1566, opened 2025-04-08
feat(runner): Create new `@wxt-dev/runner` package

aklinker1:
### Overview

This package's goal is to replace `web-ext` with a lighter-weight alternative. Additionally, Chrome will soon be dropping `--load-extesion` flag, meaning we need another way to load extensions into browsers programatically.

This should:

- Reduce WXT's install size from 80MB to 55MB
- Add support for E2E testing on Firefox
- Provide presets for persisting data between launches (user-level, project-level, or no persistence)

### Manual Testing

```sh
cd packages/runner
pnpm dev
pnpm dev firefox-nightly
```

### Related Issue

N/A


.gitignore
```diff
@@ -22,3 +22,4 @@ stats.html
 .tool-versions
 .cache
 *-stats.txt
+data
```
aklinker1, line 25:
```suggestion
```

.npmrc
```diff
@@ -0,0 +1 @@
+use-node-version=22.14.0
```
aklinker1, line 1: For `WebSocket` to be available globally, we need to use Node 22...

Timeraa (approved):

aklinker1:
Alright, this is done and working on Linux+Mac. Will need someone to test on Windows. Also need people to share where their browser binaries are located. Will post a message on discord asking for help.

PR #1573, opened 2025-04-12
feat: ignore elements with a vite-ignore or wxt-ignore attribute

Tensai75:
### Overview

This PR adds the ability for the user to use ignore attributes, either `vite-ignore` or `wxt-ignore` in html elements to make wxt ignore those elements and not process the src or href attributes.
If `wxt-ignore` is used as attribute, the attribute is removed. The `vite-ignore` attribute does not need to be removed as it is removed later by Vite.

### Manual Testing

Add the attribute `vite-ignore` or `wxt-ignore` to a `link` element in an entrypoint html file. The `href` attribute should not be processed in the resulting html file and the `vite-ignore` or `wxt-ignore` should be removed.

### Related Issue

N/A.
Manually ignoring elements may be beneficial in some specific edge cases like in the problem which was discussed here: #1548

### DIFF
--- a/packages/browser/package.json
+++ b/packages/browser/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@wxt-dev/browser",
   "description": "Provides a cross-browser API for using extension APIs and types based on @types/chrome",
-  "version": "0.0.313",
+  "version": "0.0.315",
   "type": "module",
   "main": "src/index.mjs",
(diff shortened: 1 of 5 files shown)


Tensai75:
Will add new separate PRs for the wxt-ignore attribute and the `{{packageVersion}}` as template variable.

PR #1574, opened 2025-04-13
feat: type-safe `import.meta.env.BROWSER` with new `targetBrowers` config

7sDream:
### Overview

Currently, `import.meta.env.BROWSER` is typed as a string, which, while technically correct from a type system perspective, can be limiting in practical development. An extension might only support a specific set of browsers, such as "chrome" | "firefox".

Instead of a string, defining BROWSER as an enum-like type could offer several advantages. TypeScript's controw flow analyze makes this approach particularly useful. For instance, a switch statement on import.meta.env.BROWSER would no longer require a default case. The absence of a default case is beneficial because it prevents potential oversights when adding support for new browsers like "edge" in the future. With an enum-like type, the type checker would enforce the addition of a new case in all relevant locations.

### Manual Testing

add `targetBrowsers: ["firefox", "chrome"]` in `wxt.config.js` config file, then run `wxt prepare`(e.g. `pnpm run -F wxt wxt prepare <project-root>`), you should notice the type of `import.meta.env.BROWSER` is`"firefox" | "chrome"` instead of `string`.

A simple e2e is added, `pnpm -F wxt test -t targetBrowers` to run it.

### Unresolved concerns

- [x] Should we stop the prepare/build/dev process if current browser setting is not in the targetBrowsers list? <br>Yes!

### Related Issue

N/A, If you want more context, join the [discord thread](https://discord.com/channels/1212416027611365476/1359491654565757149):

### DIFF
--- a/docs/guide/essentials/config/environment-variables.md
+++ b/docs/guide/essentials/config/environment-variables.md
@@ -35,7 +35,7 @@ WXT provides some custom environment variables based on the current command:
 | Usage                              | Type      | Description                                           |
 | ---------------------------------- | --------- | ----------------------------------------------------- |
 | `import.meta.env.MANIFEST_VERSION` | `2 │ 3`   | The target manifest version                           |
-| `import.meta.env.BROWSER`          | `string`  | The target browser.                                   |
+| `import.meta.env.BROWSER`          | `string`  | The target browser                                    |
 | `import.meta.env.CHROME`           | `boolean` | Equivalent to `import.meta.env.BROWSER === "chrome"`  |
 | `import.meta.env.FIREFOX`          | `boolean` | Equivalent to `import.meta.env.BROWSER === "firefox"` |
 | `import.meta.env.SAFARI`           | `boolean` | Equivalent to `import.meta.env.BROWSER === "safari"`  |
--- a/docs/guide/essentials/config/environment-variables.md
+++ b/docs/guide/essentials/config/environment-variables.md
@@ -42,7 +42,7 @@ WXT provides some custom environment variables based on the current command:
 | `import.meta.env.EDGE`             | `boolean` | Equivalent to `import.meta.env.BROWSER === "edge"`    |
 | `import.meta.env.OPERA`            | `boolean` | Equivalent to `import.meta.env.BROWSER === "opera"`   |
 
-The `BROWSER` can be a safer literal string union type of all browser names that you want to support, see [`targetBrowsers`](/api/reference/wxt/interfaces/InlineConfig#targetBrowers) option.
+The `BROWSER` can be a safer literal string union type of all browser names that you want to support, see [`targetBrowsers`](/api/reference/wxt/interfaces/InlineConfig#targetBrowsers) option.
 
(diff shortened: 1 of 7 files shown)


docs/guide/essentials/config/environment-variables.md
```diff
@@ -42,6 +42,8 @@ WXT provides some custom environment variables based on the current command:
 | `import.meta.env.EDGE`             | `boolean` | Equivalent to `import.meta.env.BROWSER === "edge"`    |
 | `import.meta.env.OPERA`            | `boolean` | Equivalent to `import.meta.env.BROWSER === "opera"`   |
 
+The `BROWSER` can be a safer literal string union type of all browser names that you want to support, see [`targetBrowsers`](/api/reference/wxt/interfaces/InlineConfig#targetbrowsers) option.
```
aklinker1, line 45: "safer literal string union type" is a mouthful lol. Let's go with something a little easier to understand:

```suggestion
You can set the [`targetBrowsers`](/api/reference/wxt/interfaces/InlineConfig#targetbrowsers) option to make the `BROWSER` variable a more specific type, like `"chrome" | "firefox"`.
```

packages/wxt/src/core/resolve-config.ts
```diff
@@ -70,6 +70,12 @@ export async function resolveConfig(
   if (debug) logger.level = LogLevels.debug;
 
   const browser = mergedConfig.browser ?? 'chrome';
+  const targetBrowsers = mergedConfig.targetBrowsers ?? [];
+  if (targetBrowsers.length > 0 && targetBrowsers.indexOf(browser) === -1) {
```
aklinker1, line 74:
```suggestion
  if (targetBrowsers.length > 0 && !targetBrowsers.includes(browser)) {
```

packages/wxt/src/core/utils/globals.ts
```diff
@@ -12,7 +12,12 @@ export function getGlobals(
     {
       name: 'BROWSER',
       value: config.browser,
-      type: `string`,
+      type:
+        config.targetBrowsers.length === 0
+          ? 'string'
+          : config.targetBrowsers
+              .map((browser) => JSON.stringify(browser))
```
aklinker1, line 19: Let's be a little bit more careful here. If someone where to enter an object instead of a string, I would prefer to show `"[Object object]"` over `{ ... }` as the type. Basically, we need to ensure the type will always be a string.

```suggestion
              .map((browser) => `"${browser}"`)
```

aklinker1 (commented):
This looks good to me!

packages/wxt/src/types.ts
```diff
@@ -108,6 +108,12 @@ export interface InlineConfig {
    * "chrome"
    */
   browser?: TargetBrowser;
+  /**
+   * Target browsers to support. If not empty, `import.meta.env.BROWSER` will be narrowed to a string literal type containing only the specified browser names.
```
aklinker1, line 112:
```suggestion
   * Target browsers to support. When set, `import.meta.env.BROWSER` will be narrowed to a string literal type containing only the specified browser names.
```

7sDream:
Revised as suggested!

aklinker1 (approved):

PR #1582, opened 2025-04-16
chore: update docs to point to webExt instead of runner; update config in demo

nostrorom:
### Overview

[Browser startup docs](https://wxt.dev/guide/essentials/config/browser-startup.html) mentions (and points to) `runner` prop, deprecated for `webExt`


Timeraa (approved):

aklinker1 (approved):
Thanks, good catch!

PR #1586, opened 2025-04-17
updating entrypoints.md

techlism:
### Overview
In the example  browser.action.onClick is not a valid method but browser.action.onClicked is what that's available 

### Manual Testing

I have just updated entrypoint.md file and things work as intended

### Related Issue

I know this is a bit unprofessional to raise directly a PR, I wouldn't coming from a conventional route, so reject it at your own perusal


aklinker1 (approved):
Thanks! I wrote some of these old examples from memory... Not surprised I messed them up lol

PR #1597, opened 2025-04-20
fix: adding missing "" to PublicPath and browser.runtime.getUrl.

nishu-murmu:
### Overview

for public path `""` was missing
basically `browser.runtime.getUrl("")` command is throwing type error.
I guess something was updated in this function `getPathsDeclarationEntry` which broke that.
when running `pnpm compile` it also throws error.

### Manual Testing
1. add `browser.runtime.getUrl("")` in background.ts file (you'll see the type error).
2. additionally try running this command `pnpm compile`


aklinker1 (commented):
Should it go first? I purposefully put `"/"` first so IDEs show it at the top. `""` should probably go above it?

nishu-murmu:
> Should it go first? I purposefully put `"/"` first so IDEs show it at the top. `""` should probably go above it?

@aklinker1 Updated.

Timeraa (approved):

aklinker1 (approved):

PR #1601, opened 2025-04-22
fix: fix typescript error on `defineItem` fallback

iyume:
### Overview

Currently, the second signature of `WxtStorage.defineItem` method is incorrect that it doesn't respect the fallback value.

https://github.com/wxt-dev/wxt/blob/4c35798dbaaf96983d1148544716568beae41732/packages/storage/src/index.ts#L748-L754

The following examples show the bug clearly.

```typescript
export const storageTodoItems = storage.defineItem<StorageTodoItem[]>(
  'local:StorageTodoItems',
  { fallback: [] },
)  // Type is StorageTodoItem[]

export const storageTodoItems = storage.defineItem<StorageTodoItem[]>(
  'local:StorageTodoItems',
)  // Type is StorageTodoItem[] | null

export const storageTodoItems = storage.defineItem<StorageTodoItem[]>(
  'local:StorageTodoItems', {}
)  // Type is StorageTodoItem[]. Wrong!!!
```

### Manual Testing

```typescript
export const storageTodoItems = storage.defineItem<StorageTodoItem[]>(
  'local:StorageTodoItems',
)  // Type is StorageTodoItem[] | null

export const storageTodoItems = storage.defineItem<StorageTodoItem[]>(
  'local:StorageTodoItems',
  {},
)  // Type is StorageTodoItem[] | null

export const storageTodoItems = storage.defineItem<StorageTodoItem[]>(
  'local:StorageTodoItems',
  { fallback: [] },
)  // Type is StorageTodoItem[]
```

### Related Issue

### DIFF
--- a/packages/storage/src/__tests__/index.test.ts
+++ b/packages/storage/src/__tests__/index.test.ts
@@ -1281,6 +1281,19 @@ describe('Storage Utils', () => {
       it('should define a nullable value when options are not passed', () => {
         const item = storage.defineItem<number>(`local:test`);
         expectTypeOf(item).toEqualTypeOf<WxtStorageItem<number | null, {}>>();
+
+        const item2 = storage.defineItem<number>(`local:test`, {});
+        expectTypeOf(item2).toEqualTypeOf<WxtStorageItem<number | null, {}>>();
+
+        const item3 = storage.defineItem<number>(`local:test`, {
+          fallback: undefined,
+        });
+        expectTypeOf(item3).toEqualTypeOf<WxtStorageItem<number | null, {}>>();
+
+        const item4 = storage.defineItem<number>(`local:test`, {
+          defaultValue: undefined,
+        });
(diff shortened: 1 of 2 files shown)


aklinker1 (changes requested):
Thanks for the PR, looks good!

Could you add type tests to [`packages/storage/src/__tests__/index.test.ts`](https://github.com/wxt-dev/wxt/blob/main/packages/storage/src/__tests__/index.test.ts) instead of just listing them in the PR description? That way we can ensure they don't break in the future.

Something like this should be good enough:

```ts
it("should consider the fallback when typing the value", () => {
  expectTypeOf(...).toBe<...>()
});
```

Don't worry about running vitest's type-test mode, just run `pnpm check` to run the typescript compiler.

iyume:
@aklinker1 Thanks for the review! Tests are added for this case and passing with `pnpm check`.

aklinker1 (approved):

PR #1602, opened 2025-04-22
fix: move the `public` folder outside of `src` for the default Svelte template

0x2E:
### Overview

The `public` folder should be outside the `src`; otherwise, it won't be processed during the build.

### Manual Testing

N/A

### Related Issue

N/A


aklinker1 (approved):
Thanks!

PR #1604, opened 2025-04-22
feat: add `{{packageVersion}}` as template variable

Tensai75:
### Overview

This PR adds {{packageVersion}} as an additional template variable to artifactTemplate and sourcesTemplate.
Currently only {{version}} can be used. However, when building for Firefox, a package version such as 1.0.0-beta.1 is reduced to 1.0.0 because Firefox does not allow suffixes in the version in the manifest file. Therefore, the version numbers in the file names of the zip files for Firefox differ from the version number in the file name for Chrome, e.g.:
```
myextension-1.0.0-beta.1-chrome.zip
myextension-1.0.0-firefox.zip
myextension-1.0.0-sources.zip
```

With this PR and when using {{packageVersion}} instead of {{version}}, the version in the file names is the same:
```
myextension-1.0.0-beta.1-chrome.zip
myextension-1.0.0-beta.1-firefox.zip
myextension-1.0.0-beta.1-sources.zip
```

### Manual Testing

Change your package version to 1.0.0-beta.1 .
Set a custom artifactTemplate and sourcesTemplate in your wxt.config.ts:
```
artifactTemplate: "{{name}}-{{packageVersion}}-{{browser}}.zip",
sourcesTemplate: "{{name}}-{{packageVersion}}-sources.zip" ,
```
Run `wxt zip -b firefox`.
The resulting zip files should contain the package version `1.0.0-beta.1` in the file name.

### Related Issue

N/A
I think this PR might be beneficial for people like me who want to see the package version number in the filename.

### DIFF
--- a/packages/wxt/src/core/zip.ts
+++ b/packages/wxt/src/core/zip.ts
@@ -31,6 +31,7 @@ export async function zip(config?: InlineConfig): Promise<string[]> {
     safeFilename(
       (await getPackageJson())?.name || path.basename(process.cwd()),
     );
+  const packageVersion = (await getPackageJson())?.version;
   const applyTemplate = (template: string): string =>
     template
       .replaceAll('{{name}}', projectName)
@@ -39,6 +40,7 @@ export async function zip(config?: InlineConfig): Promise<string[]> {
         '{{version}}',
         output.manifest.version_name ?? output.manifest.version,
       )
+      .replaceAll('{{packageVersion}}', packageVersion)
       .replaceAll('{{mode}}', wxt.config.mode)
       .replaceAll('{{manifestVersion}}', `mv${wxt.config.manifestVersion}`);
 


packages/wxt/src/core/zip.ts
```diff
@@ -39,6 +40,7 @@ export async function zip(config?: InlineConfig): Promise<string[]> {
         '{{version}}',
         output.manifest.version_name ?? output.manifest.version,
```
aklinker1, line 41: As you can see here, the `{{version}}` variable will already use the `manifest.version_name` if it exists... And as long as you haven't manually set the manifest's `version` property, it will be automatically pulled from the `package.json`:

https://wxt.dev/guide/essentials/config/manifest.html#version-and-version-name

Tensai75: No, it doesn't.
I use custom artifactTemplate and sourcesTemplate:
```
zip: {
    excludeSources: ['icons/**', 'screenshots/**', '*.yml'],
    artifactTemplate: '{{name}}-v{{version}}-{{browser}}.zip',
    sourcesTemplate: '{{name}}-v{{version}}-sources.zip',
  },
```

Package version is `1.0.0-beta.1` and I don't set the manifest's version property manually. But the resulting zip files are named:
```
myextension-v1.0.0-beta.1-chrome.zip
myextension-v1.0.0-firefox.zip
myextension-v1.0.0-sources.zip
```
`-beta.1` is not included in the firefox and sources zip!

aklinker1: Hmm, I understand the problem now, yeah, I'm OK with this change. Two changes:

1. We call `getPackageJson` in the line above your change. Call it once and extract it into a variable. 
2. Update the JSDoc:
   https://github.com/wxt-dev/wxt/blob/f5619f6be1b36ebbf6018bb54660893d440878da/packages/wxt/src/types.ts#L143-L156
   https://github.com/wxt-dev/wxt/blob/f5619f6be1b36ebbf6018bb54660893d440878da/packages/wxt/src/types.ts#L167-L180

Tensai75: If you are OK with a breaking change, maybe it would be more consisten to to have 3 template variables for the version:

- `{{version}}` --> strictly use `manifest.version`
- `{{versionName}}` --> use `manifest.version_name` if available or fallback to `manifest.version` _(current behaviour of `{{version}}`)_
- `{{packageVersion}}` --> always use `package.version`

The behaviour of `{{version}}` would change but this feels more consitsten.

What do you say?

Tensai75: I have updated the PR according to your suggestions. I hope it is OK.
I will create a separate PR for the breaking change later.

aklinker1:
Looks like the formatting check failed. Run `pnpm -w check` locally to fix it.

Tensai75:
It should pass now. Sorry...

aklinker1:
> It should pass now. Sorry...

Haha, you're good!

PR #1605, opened 2025-04-22
Adds mode to the zip filenames

sleekslush:
### Related Issue

### DIFF
--- a/packages/wxt/src/core/resolve-config.ts
+++ b/packages/wxt/src/core/resolve-config.ts
@@ -301,8 +301,8 @@ function resolveZipConfig(
   const downloadedPackagesDir = path.resolve(root, '.wxt/local_modules');
   return {
     name: undefined,
-    sourcesTemplate: '{{name}}-{{version}}-sources.zip',
-    artifactTemplate: '{{name}}-{{version}}-{{browser}}.zip',
+    sourcesTemplate: '{{name}}-{{version}}-{{browser}}-{{mode}}-sources.zip',
+    artifactTemplate: '{{name}}-{{version}}-{{browser}}-{{mode}}.zip',
     sourcesRoot: root,
     includeSources: [],
     compressionLevel: 9,
--- a/packages/wxt/e2e/tests/zip.test.ts
+++ b/packages/wxt/e2e/tests/zip.test.ts
@@ -91,7 +91,7 @@ describe('Zipping', () => {
       mode: 'development',
       zip: {
         artifactTemplate: '{{name}}-{{version}}-{{browser}}-{{mode}}.zip',
-        sourcesTemplate: '{{name}}-{{version}}-{{mode}}-sources.zip',
+        sourcesTemplate: '{{name}}-{{version}}-{{browser}}-{{mode}}-sources.zip',
       },
     });
 


Tensai75:
I vote against this PR. `mode` should not be included in the default template for the zip file names.
There is aways the possibility to add custom artifactTemplate and sourcesTemplate to the `wxt.config.ts` if someone wants this.

sleekslush:
> I vote against this PR. `mode` should not be included in the default template for the zip file names. There is aways the possibility to add custom artifactTemplate and sourcesTemplate to the `wxt.config.ts` if someone wants this.

This was done to keep parity with the default behavior of the output directory. There's a quick chat about it here too https://discord.com/channels/1212416027611365476/1224507127863967784/1359972453778067738

Tensai75:
Well, then please let add my two cents to this discussion:
I strongly believe that default templates/settings should reflect the simplest use case that the majority of users would use/prefer. If more than 50% of users change the default templates/settings, they are poorly chosen.
And I would further argue that the majority of users only target one manifest mode per browser. In such a use case, most users would probably consider the mode to be irrelevant information that would only unnecessarily lengthen the name of the zip file.
I would even go so far as to suggest changing the default setting for the build folders to `{{browser}}` only, rather than changing the names of the zip files.

Anyway, this is just my opinion. But in any case, this PR is a breaking change that should be labelled as such.

sleekslush:
> I would even go so far as to suggest changing the default setting for the build folders to `{{browser}}` only, rather than changing the names of the zip files.

That's the way it was before 0.20.0. The change is here https://github.com/wxt-dev/wxt/pull/1086/files#diff-ff0465c3a486d3ba187204149a25fc8f632c44d65da356dc04c0f2b268a71506

My entire goal is to simply keep them consistent, hence the change.

> 
> Anyway, this is just my opinion. But in any case, this PR is a breaking change that should be labelled as such.

Sure.

packages/wxt/src/core/resolve-config.ts
```diff
@@ -301,8 +301,8 @@ function resolveZipConfig(
   const downloadedPackagesDir = path.resolve(root, '.wxt/local_modules');
   return {
     name: undefined,
-    sourcesTemplate: '{{name}}-{{version}}-sources.zip',
-    artifactTemplate: '{{name}}-{{version}}-{{browser}}.zip',
+    sourcesTemplate: '{{name}}-{{version}}-{{browser}}-{{mode}}-sources.zip',
+    artifactTemplate: '{{name}}-{{version}}-{{browser}}-{{mode}}.zip',
```
Tensai75, line 305: I have to admit I misunderstood the PR at first and thought it was the `{{manifestVersion}}` that should be added to the filenames, but it's the production mode.
Ok, then I agree with this PR in principle, because I too like consistency. But the `outDirTemplate` uses `{{modeSuffix}}` (‘-dev’ for development, ‘’ for production): 
https://github.com/wxt-dev/wxt/blob/d9f10c62ad3abd10a3d9c599d9c668607fe0956b/packages/wxt/src/core/resolve-config.ts#L112-L115
So we should use `{{modeSuffix}}` in `artifactTemplate` and `sourcesTemplate` as well and not add `{{browser}}` to `sourcesTemplate` (the source is exactly the same for all browsers):
```
sourcesTemplate: '{{name}}-{{version}}-sources{{modeSuffix}}.zip',
artifactTemplate: '{{name}}-{{version}}-{{browser}}{{modeSuffix}}.zip',
```
This would for development result in:
```
extensionname-1.0.0-chrome-dev.zip
extensionname-1.0.0-firefox-dev.zip
extensionname-1.0.0-sources-dev.zip
```
and for production in:

```
extensionname-1.0.0-chrome.zip
extensionname-1.0.0-firefox.zip
extensionname-1.0.0-sources.zip
```

sleekslush: I'm already handling this all in my config, so I'll just close this PR. Looks like you have another PR related to these templates, so I'll let you handle whatever you think will work best in the current state of things.

aklinker1:
> ```
> sourcesTemplate: '{{name}}-{{version}}-sources{{modeSuffix}}.zip'
> artifactTemplate: '{{name}}-{{version}}-{{browser}}{{modeSuffix}}.zip'
> ```

Yes, we should do this in the next major version.

PR #1635, opened 2025-05-01
feat: Add `@font-face` to be processed by `splitShadowRootCss`

seaders:
### Overview

Discussed with Aaron on the discord, when using `createShadowRootUi` for `defineContentScript`, font faces must be added outside that root, like is already done with `@property`, this PR does that.

### Manual Testing

There was already 2 tests checking `@property` are handled correctly, and added to the document css, not shadow, and I've amended them to handle and check that `@font-face` are too.


aklinker1 (commented):
Code looks good, away from my computer for a few days, will test when I can!

Issue #1589, opened 2025-04-17
feat: Places that allow passing enum values should also allow passing strings


rxliuli:
### Feature Request

For example, the code below uses two enumeration values of browser.declarativeNetRequest. Unfortunately, they do not exist in Safari. Should the type be modified to support string constants?

```ts
browser.declarativeNetRequest.updateDynamicRules({
  addRules: [
    {
      id: 1,
      action: {
        type: browser.declarativeNetRequest.RuleActionType.MODIFY_HEADERS,
        responseHeaders: [],
      },
      condition: {
        resourceTypes: [
          browser.declarativeNetRequest.ResourceType.XMLHTTPREQUEST,
        ],
      },
    },
  ],
})
```

<img width="796" alt="Image" src="https://github.com/user-attachments/assets/eb131fde-bcda-4684-bd92-4d5c584f94ae" />

#### Is your feature request related to a bug?

N/A

### What are the alternatives?

Currently, I am using type casting, which can solve the problem, but it doesn't seem to be a good method.

```ts
browser.declarativeNetRequest.updateDynamicRules({
  addRules: [
    {
      id: 1,
      action: {
        type: 'modifyHeaders' as Browser.declarativeNetRequest.RuleActionType,
        responseHeaders: [],
      },
      condition: {
        resourceTypes: [
          'xmlhttprequest' as Browser.declarativeNetRequest.ResourceType,
        ],
      },
    },
  ],
})
```

### Additional context

Of course, everything is fine in Chrome.

<img width="545" alt="Image" src="https://github.com/user-attachments/assets/17af7aed-dd33-45a1-a669-7505329c6a07" />

aklinker1:
This is an upstream issue with `@types/chrome`

rxliuli:
> This is an upstream issue with `@types/chrome`

But this is indeed correct in Chrome, and I'm not sure if upstream `@types/chrome` would be willing to accept this modification, but I will give it a try.

aklinker1:
`@wxt-dev/browser` will auto-update when new versions of `@types/chrome` are released. Looks like 0.0.326 has the change and is released.

PR #1646, opened 2025-05-02
fix: Allow passing strings instead of enums

aklinker1:
### Overview

Addresses concerns mentioned here:
- https://github.com/wxt-dev/wxt/issues/1589
- https://github.com/wxt-dev/wxt/discussions/1541#discussioncomment-12967903

> [!WARNING]
> Not sure if I want to actually merge this PR or not... The regex approach isn't very clean, and it makes mistakes. I think it would make more sense to update `@types/chrome` directly to allow using string litterals in specific cases, [like they already do](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/562ab89f5b7b6ec777941f9578b7c81eff10c5af/types/chrome/index.d.ts#L683-L684).

### Manual Testing

See new type test


packages/browser/src/gen/index.d.ts
```diff
@@ -13209,7 +13209,7 @@ export namespace Browser {
             /**
              * A list of request types. Requests that cannot match any of the types will be filtered out.
              */
-            types?: ResourceType[] | undefined;
+            types?: (ResourceType | `${ResourceType}`)[] | undefined;
```
aklinker1, line 13212: Downside to this approach. Here, `ResourceType` is actually a string literal union. There's another enum further down that is named `ResourceType`, so this type was also modified.

aklinker1:
`@types/chrome` fixed this in https://github.com/wxt-dev/wxt/issues/1589#issuecomment-2921653426

PR #1653, opened 2025-05-03
fix: Allow custom API URL in Google Analytics 4 provider options

honwhy:
### Overview

see dicussion: Help adding Umami and/or GA4 #1596

avoid leaking google analytics api_serect in extension by forwarding request to self-hosted server

### Related Issue

N/A

### DIFF
--- a/packages/analytics/modules/analytics/providers/google-analytics-4.ts
+++ b/packages/analytics/modules/analytics/providers/google-analytics-4.ts
@@ -4,6 +4,7 @@ import type { BaseAnalyticsEvent } from '../types';
 const DEFAULT_ENGAGEMENT_TIME_IN_MSEC = 100;
 
 export interface GoogleAnalytics4ProviderOptions {
+  apiUrl: string | undefined;
   apiSecret: string;
   measurementId: string;
 }
@@ -18,7 +19,7 @@ export const googleAnalytics4 =
       ): Promise<void> => {
         const url = new URL(
           config?.debug ? '/debug/mp/collect' : '/mp/collect',
-          'https://www.google-analytics.com',
+          options?.apiUrl ? options.apiUrl : 'https://www.google-analytics.com',
         );
         if (options.apiSecret)
           url.searchParams.set('api_secret', options.apiSecret);


packages/analytics/modules/analytics/providers/google-analytics-4.ts
```diff
@@ -4,6 +4,7 @@ import type { BaseAnalyticsEvent } from '../types';
 const DEFAULT_ENGAGEMENT_TIME_IN_MSEC = 100;
 
 export interface GoogleAnalytics4ProviderOptions {
+  apiUrl: string | undefined;
```
PatrykKuniczak:
```suggestion
  apiUrl?: string;
```

packages/analytics/modules/analytics/providers/google-analytics-4.ts
```diff
@@ -18,7 +19,7 @@ export const googleAnalytics4 =
       ): Promise<void> => {
         const url = new URL(
           config?.debug ? '/debug/mp/collect' : '/mp/collect',
-          'https://www.google-analytics.com',
+          options?.apiUrl ? options.apiUrl : 'https://www.google-analytics.com',
```
PatrykKuniczak:
```suggestion
          options.apiUrl ?? 'https://www.google-analytics.com',
```

packages/analytics/modules/analytics/providers/google-analytics-4.ts
PatrykKuniczak, line 1: You can simplify it a little, but anyway, we need to know if @aklinker1 want that change.

I'm also wondering about this package, when i was onboarding, my idea is to add types for that values:
```ts
     googleAnalytics4({
        apiSecret: import.meta.env.WXT_GA_API_SECRET,
        measurementId: import.meta.env.WXT_GA_MEASUREMENT_ID,
      }),
```
```ts
interface ImportMeta {
  env: {
    WXT_GA_API_SECRET: string;
    WXT_GA_MEASUREMENT_ID: string;
  };
}
```

To prevent users, from use hardcoded strings here(they can forget to change it to .env value or don't remove before commit, OF COURSE github have security bot, but anyway this could be extra layer of protection), because those tokens shouldn't be expose of course.

And this types can gives possibility, to have auto completion hint from IDE.
Let's give me feedback, if those changes is accepted and could be helpful

aklinker1:
> avoid leaking google analytics api_serect in extension by forwarding request to self-hosted server

Smart!

packages/analytics/modules/analytics/providers/google-analytics-4.ts
aklinker1: The secret will be included in the final bundle regardless of if it's hard coded or in your `.env` file - vite just replaces `import.meta.env` references with their values, it doesn't magically hide them or obfusticate anything.

As it is now, devs can name the env vars anything they want, I don't think WXT needs to enforce a specific name. Usually projects add their own env var types that are specific to their projects so they get type hints as well.

honwhy:
> Changes look good @honwhy, we can merge after you make the two recommended changes from @PatrykKuniczak.

Recommendation received. Updated and committed.

packages/analytics/modules/analytics/providers/google-analytics-4.ts
PatrykKuniczak: Yeah, in the final bundle or course, yeah yeah

Rest, make sense, my change isn't necessary under that circumstances.

PatrykKuniczak:
Anyway, we need approval of @aklinker1 but this is on the home stretch

PatrykKuniczak:
@aklinker1 I'll need to a little more on next few months, i shouldn't have `maintainer` role for now, even if i'll have it, i won't merge sth.

Maybe i can merge it, because you've accepted it, but it's better to PRs be merged by you 😸

Issue #1652, opened 2025-05-03
Add support for Firefox-specific APIs


XFox111:
### Feature Request

Add types for browser-specific APIs and make them separate. For example:

```ts
// This API is cross-browser compatible, so it can be accessed via `browser`
await browser.tabs.create({ url: "https://github.com" });

// This is Chrome-specific API, which is not accessible on Firefox
await chrome.sidePanel.open({ windowId: 123 });

// This is Firefox-specific API, which is not accessible on Chrome
await firefox.sidebarAction.open();
```

Ideally, if possible, it would be also a good idea to setup some linting for that, in case an extension tries to access browser-specific API without first checking which browser it is currently running on.

#### Is your feature request related to a bug?

#1651 

### What are the alternatives?

Bring back `webextension-polyfill`. That way we can use `browser` for cross-compatible/Firefox-specific APIs, and `chrome` for Chrome-specific ones.

### Additional context

Chrome and Firefox have quite a big set of incompatible APIs, and it would be nice to have access to them as well

XFox111:
> Ideally, if possible, it would be also a good idea to setup some linting for that, in case an extension tries to access browser-specific API without first checking which browser it is currently running on.

This could be achieved by setting `firefox` and `chrome` to `null` based on the browser it was compiled for. That way, instead of doing this:

```ts
if (import.meta.env.FIREFOX)
    firefox.sidebarAction.open();
else
    chrome.sidePanel.open({...});
```

Which isn't type-safe, we could do this:

```ts
if (firefox)
    firefox.sidebarAction.open();
else if (chrome)
    chrome.sidePanel.open({...});
```

Which doesn't look too good either, but at least it provides type safety

honwhy:
should wxt do dynamic dispatch ?
```
browser.sidePanel.open()
```

call `chrome.sidePanel.open` when in chrome, call `browser.sidebarAction.open` when in firefox

aklinker1:
So what are the main concerns here, and what are we trying to solve?

- Type definitions for Firefox/MV2 APIs
- Type-safety for accessing those APIs
- ...?

See https://github.com/wxt-dev/wxt/pull/1661#issuecomment-2949782376 for context. I'd like to get an understanding of what people want so we can build the correct solution.

[...]

PR #1661, opened 2025-05-06
feat: Add firefox and chrome variable in `@wxt-dev/browser`

SimYunSup:
### Overview

Add `chrome` and `firefox` variable for specific logic. I'm not sure if this implementation is correct, so I'm going to submit it as a Draft PR.

### Related Issue

Related #1652


aklinker1:
I'm not sure if I like this change.

- What about safari?
- Breaking the standards - projects migrating to WXT may not know which variable to use, and projects migrating away from WXT now have more work to do.
   - Counterpoints:
      - They could just install the `@wxt-dev/browser` package and continue to use it
      - WXT already uses `browser`, which most projects don't use.
- I would rather have a single variable include all the types, rather than exporting multiple variables each with their own types.
- Why add a variable for `chrome` when the `browser` variable is already typed using `@types/chrome`?
- What happens to this logic when Chrome [releases the `browser` global for compatibility](https://groups.google.com/a/chromium.org/g/chromium-extensions/c/gK1Sd57p4go)? `if (firefox) ... else if (chrome) ...` would not work anymore since it would always be firefox.

Let's continue the discussion on https://github.com/wxt-dev/wxt/issues/1652 before doing any more work here. I'm not convinced this is the correct solution yet.

aklinker1:
Gonna close this given the conversation here: https://github.com/wxt-dev/wxt/issues/1652#issuecomment-2949819368

Doesn't look like we'll be adding the `chrome` and `firefox` varaibles. Let's open a new PR for whatever is decided.

SimYunSup:
@aklinker1 This implemention is changed for `automatically generating TypeScript definitions based on data from @types/chrome and @types/firefox-webext-browser` as I commented https://github.com/wxt-dev/wxt/issues/1652#issuecomment-2952155548 . Could I create new PR from this branch?

PR #1673, opened 2025-05-15
docs: fix svelte app unmount code

hyunbinseo:
### Overview

> Cannot find name 'app'. Did you mean 'App'? `ts(2552)`

https://svelte.dev/docs/svelte/imperative-component-api#unmount

```ts
import { mount, unmount } from 'svelte';
import App from './App.svelte';

const app = mount(App, { target: document.body });

// later
unmount(app, { outro: true });
```

### DIFF
--- a/docs/guide/essentials/content-scripts.md
+++ b/docs/guide/essentials/content-scripts.md
@@ -205,13 +205,12 @@ export default defineContentScript({
   matches: ['<all_urls>'],
 
   main(ctx) {
-    let app: Record<string, any>;
     const ui = createIntegratedUi(ctx, {
       position: 'inline',
       anchor: 'body',
       onMount: (container) => {
         // Create the Svelte app inside the UI container
-        app = mount(App, { target: container });
+        return mount(App, { target: container });
       },
       onRemove: (app) => {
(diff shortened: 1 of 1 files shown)


docs/guide/essentials/content-scripts.md
```diff
@@ -205,14 +205,13 @@ export default defineContentScript({
...
+    let app: Record<string, any>;
     const ui = createIntegratedUi(ctx, {
       position: 'inline',
       anchor: 'body',
       onMount: (container) => {
         // Create the Svelte app inside the UI container
-        mount(App, {
-          target: container,
-        });
+        app = mount(App, { target: container });
```
SimYunSup, line 214:
```suggestion
        const app = mount(App, { target: container });
        return app;
```

https://github.com/wxt-dev/wxt/blob/4f772fc0f7bdecc597a5470639cb78bd93682cac/packages/module-react/entrypoints/content/index.tsx#L25-L34

React module is following `return appInstance;` in `onMount` and `unmount` in `onRemove`. So I think it must changed like above.

aklinker1 (approved):
Thanks!

PR #1707, opened 2025-06-04
fix: Support http server websocket inside containers

aklinker1:
### Overview

This addresses the regression mentioned here: https://github.com/wxt-dev/wxt/pull/1406#discussion_r2121977297

### DIFF
--- a/packages/wxt/src/core/builders/vite/plugins/devServerGlobals.ts
+++ b/packages/wxt/src/core/builders/vite/plugins/devServerGlobals.ts
@@ -16,7 +16,7 @@ export function devServerGlobals(
       return {
         define: {
           __DEV_SERVER_ORIGIN__: JSON.stringify(
-            server.origin.replace(/^http(s):/, 'ws$1:'),
+            server.origin.replace(/^https?:/, 'ws$1:'),
           ),
         },
       };


packages/wxt/src/core/builders/vite/plugins/devServerGlobals.ts
```diff
@@ -16,7 +16,7 @@ export function devServerGlobals(
       return {
         define: {
           __DEV_SERVER_ORIGIN__: JSON.stringify(
-            server.origin.replace(/^http(s):/, 'ws$1:'),
+            server.origin.replace(/^https?:/, 'ws$1:'),
```
alecdwm, line 19: Ah, it still needs the `(s)` @aklinker1:

without `(s)`:


with `(s)`:


from [regex101.com](https://regex101.com/)

aklinker1: Shoot, this is a replace. My bad, I was in a rush

aklinker1: Thanks for catching so quickly, glad I didn't release.

aklinker1: Fixed in https://github.com/wxt-dev/wxt/pull/1712

Screenshot 2025-06-05 at 12 22 13 AM

alecdwm: No worries, thank you! 😀

PR #1386, opened 2025-02-01
chore: setup markdownlint and fix markdown formatting

okineadev:
### Overview

- Fixed Markdownlint violations, simplified formatting, minimized HTML usage, and improved readability
	<details>
	<summary>Copilot summary</summary>
	
	This pull request includes several changes to improve the formatting and consistency of markdown files across the repository. The most important changes include updating email and URL formatting, modifying the `README.md` layout, and updating image links.
	
	Formatting improvements:
	
	* [`CODE_OF_CONDUCT.md`](diffhunk://#diff-ffdbe3a1e7ee93cacfc080b6c635ccf3a8f6b0f00f2fb884f78c6b5f9dac8fd2L63-R63): Updated email and URL formatting to use angle brackets for consistency. [[1]](diffhunk://#diff-ffdbe3a1e7ee93cacfc080b6c635ccf3a8f6b0f00f2fb884f78c6b5f9dac8fd2L63-R63) [[2]](diffhunk://#diff-ffdbe3a1e7ee93cacfc080b6c635ccf3a8f6b0f00f2fb884f78c6b5f9dac8fd2L119-R128)
	* [`CONTRIBUTING.md`](diffhunk://#diff-eca12c0a30e25b4b46522ebf89465a03ba72a03f540796c979137931d8f92055L151-R151): Updated URL formatting to use angle brackets for consistency.
	
	Layout and link updates:
	
	* [`README.md`](diffhunk://#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5L1-R34): Modified the layout to use markdownlint rules, updated image links, and improved readability by using markdown syntax for links and images. [[1]](diffhunk://#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5L1-R34) [[2]](diffhunk://#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5L82-R86)
	* [`SECURITY.md`](diffhunk://#diff-f6ed156e4bf5c791680662464b94ea5d753f219ee816b385f67870e2c0d7d4c7L5-R5): Updated image link formatting to use markdown syntax for consistency.
	
	</details>

PR #1709, opened 2025-06-04
chore: Add `markdownlint` to repo

aklinker1:
### Overview

This replaces #1386. I can't make changes to the PR and there's too much I want to change after looking at it again. Going to break it into two PRs:

1. Add and configure markdownlint. Add it to CI, but ignore any failures
2. Fix all the problems and require no issues to pass CI.

This PR implements the first step. Will create a second after this is merged to do the second.

### Manual Testing

```sh
pnpm experimental:check:md
```

### DIFF
--- a/.markdownlint.json
+++ b/.markdownlint.json
@@ -1,0 +1,6 @@
+{
+  "$schema": "https://raw.githubusercontent.com/DavidAnson/markdownlint/refs/heads/main/schema/markdownlint-config-schema.json",
+  "line-length": false,
+  "no-inline-html": false,
+  "first-line-heading": false
+}


.markdownlint.json
```diff
@@ -0,0 +1,6 @@
+{
+  "$schema": "https://raw.githubusercontent.com/DavidAnson/markdownlint/refs/heads/main/schema/markdownlint-config-schema.json",
+  "line-length": false,
+  "no-inline-html": false,
+  "first-line-heading": false
```
aklinker1, line 5: I'd like to get rid of this config file completely and just use the defaults, but for now:

- `--fix` doesn't appear to fix line-lengths
- Vitepress conflicts with first-line-heading
- I can fix no-inline-html (see https://github.com/wxt-dev/wxt/pull/1386/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5)

But I can do that in separate PRs.

PR #1732, opened 2025-06-12
chore: add `dev/build` script command

btea:
### Overview

https://pnpm.io/pnpm-cli#-c-path---dir-path

Add relevant script commands directly in the package.json in the root directory for quick and easy execution without having to manually enter the specified folder.
### Related Issue

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -40,19 +40,16 @@ Here are some helpful commands:
 
 ```sh
 # Build WXT package
-cd packages/wxt
 pnpm build
 ```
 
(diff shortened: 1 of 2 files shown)


aklinker1 (approved):

aklinker1:
Actually, no, I don't want a `build` script in the root of the monorepo that only builds a single package, even if it's the main one. The other changes are fine. But why not just use `pnpm --filter wxt build` or `pnpm --filter wxt-demo build`?

btea:
Yeah, the effect should look the same.

PR #1742, opened 2025-06-14
Take browser port configuration into account

nimatrueway:
### Related Issue


fbenites:
Works for me (testing with selenium requires a different port which is not set without this PR, I used the line for chromium and it works, else I get a  ERROR  connect ECONNREFUSED 127.0.0.1:38899                                                                                                                               5:23:06 PM

    at TCPConnectWrap.afterConnect [as oncomplete] (node:net:1634:16))

aklinker1:
Closing in favor of #1818

PR #1751, opened 2025-06-16
fix: rootDir, not srdDir for public/_locales

mjfaga:
### Overview

Fixing docs to call use the `rootDir`, not the `srcDir` when configuring the `public/_locales` directory.

### Manual Testing

N/A

### Related Issue

N/A


aklinker1 (approved):
Woops, thanks!

PR #1761, opened 2025-06-19
feat: Remove script element immediately in injectScript

ion1:
### Overview

I am trying to minimize any chance of my extension leaving something in the DOM or the window object etc.

A script which fails to load would be left in the DOM. Rather than adding an onerror handler, it seems that the script element can just be removed immediately after being added.

Everything seems to work: the script is executed; onload, onerror handlers do the right thing; document.currentScript invoked by the script returns the detached script element.

### Manual Testing

Just using injectScript normally should suffice. I am working on an extension which uses injectScript. It seems to work fine with this change.

### Related Issue

N/A

Issue #1755, opened 2025-06-17
injectScript passing data to the script


ion1:
### Feature Request

It would be nice if something like this was possible:
```ts
await injectScript("/script.js", {
  parameter: [1, 2, 3],
});
```
```ts
// entrypoints/script.ts
export default defineUnlistedScript({
  main<number[]>(ctx) {
    console.info(ctx.parameter); // [1, 2, 3]
  },
);
```
where `injectScript` would do something like this:
```ts
if (parameter !== undefined)
  script.dataset["wxtParameter"] = JSON.stringify(parameter);
```
and the build result from `entrypoints/script.ts` would include something like this:
```ts
const parameterString = document.currentScript?.dataset?.["wxtParameter"];
const parameter = parameterString !== undefined ? JSON.parse(parameterString) : undefined;
return await defintion.main({ parameter });
```

Additionally, if #1754 is implemented, this could type-check the parameter against the type of the function.
```ts
// entrypoints/content.ts
import script from "./script";

export default defineContentScript({
  matches: ["https://example.net/*"],
  async main() {
    await injectScript(script, {
      parameter: [1, 2, 3], // type-checked
    });
  },
});
```

Thanks!

#### Is your feature request related to a bug?

N/A

### What are the alternatives?

Leaving things as they are.

### Additional context

N/A

PR #1762, opened 2025-06-19
feat: Add manipulateScript option to injectScript

ion1:
> [!NOTE]
> This PR builds on top of #1761 at the moment. In case it is not accepted, I will rebase this on top of main.

### Overview

`manipulateScript` enables the the manipulation of the script element just before it is added to the DOM.

It can be useful for e.g. passing data to the script via the `dataset` property (which can be accessed by the script via
`document.currentScript`), as shown in the example:

```ts
// entrypoints/example.content.ts
  matches: ['*://*/*'],
  async main() {
    await injectScript('/example-main-world.js', {
      manipulateScript(script) {
        script.dataset['greeting'] = 'Hello there';
      },
    });
  },
});
```

```ts
// entrypoints/example-main-world.ts
  console.log(document.currentScript?.dataset['greeting']);
});
```

### Manual Testing

It can be tested as shown in the example above.

### Related Issue

#1755 proposes a higher level (and potentially type-checked) API for passing data to the script, but `manipulateScript` can still be useful for other things such as changing `script.async`/`script.defer`.

### DIFF
--- a/docs/guide/essentials/content-scripts.md
+++ b/docs/guide/essentials/content-scripts.md
@@ -591,6 +591,29 @@ For MV3, `injectScript` is synchronous and the injected script will be evaluated
 However for MV2, `injectScript` has to `fetch` the script's text content and create an inline `<script>` block. This means for MV2, your script is injected asynchronously and it will not be evaluated at the same time as your content script's `run_at`.
 :::
 
+The `script` element can be manipulated just before it is added to the DOM by using the `manipulateScript` option. This can be used to e.g. pass data to the script as shown in the example:
+
+```ts
+// entrypoints/example.content.ts
+export default defineContentScript({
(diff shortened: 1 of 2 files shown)


aklinker1 (changes requested):
Looks good, but can we call it `modify` instead of `manipulate`?

ion1:
> Looks good, but can we call it `modify` instead of `manipulate`?

Sure, I'll update my patch series in a bit.

ion1:
> Looks good, but can we call it `modify` instead of `manipulate`?

Updated.

ion1:
I had forgotten to rename one `manipulateScript` as `modifyScript` in the documentation. Fixed, and tested that the code in the documentation works.

aklinker1 (approved):
Getting all your PRs merged...

PR #1764, opened 2025-06-19
refactor: Use `script.text` instead of `innerHTML` in `injectScript`

ion1:
### Overview

`script.innerHTML` seems to act like `script.text` and not do anything unsafe with special characters, but it seems more appropriate to use `text`.

### Manual Testing

Any `injectScript` invocation in MV2 mode will hit the code path. I have tested it in Firefox.

### Related Issue

N/A


aklinker1:
Hey, thanks for your PRs! I'm back from a break and I am going through all the PRs oldest to newest, prioritizing bugs. So it might be a few weeks before I get to these 🙏

aklinker1 (approved):
OK, tested and this seems to work. Probably safer than `innerHTML`.

Issue #1754, opened 2025-06-17
Passing an UnlistedScriptDefinition to injectScript


ion1:
### Feature Request

It would be nice if the build system substituted the path of the script in the `injectScript` call as follows.
```ts
// main-injected.ts
export default defineUnlistedScript(/* ... */);
```
```ts
// isolated.content.ts
import mainInjected from "./main-injected";

export default defineContentScript({
  matches: ["https://example.net/*"],
  async main() {
    await injectScript(mainInjected);
  },
});
```

Thanks!

#### Is your feature request related to a bug?

N/A

### What are the alternatives?

Keeping things as-is.

### Additional context

N/A

PR #1765, opened 2025-06-20
[Advice needed] feat: Have injectScript return the return value of the script

ion1:
### Overview

A sketch for having `injectScript` return the result value from the script. The code _works_, but it currently has changes to virtual/unlisted-script-entrypoint.ts that only work with `injectScript` while being broken with the `browser.scripting` API.

I would like to get feedback on how you would like that to be resolved. Some options I can think of:
- Have the entrypoint detect whether it is being run from `injectScript` or `executeScript`. I do not know how to achieve that reliably. `document.currentScript` can be overridden by the web page using `Object.defineProperty`, so its value might not be a reliable way to determine that.
- Build separate entrypoints depending on whether the script is injected using `injectScript` or `executeScript`.

### Manual Testing

Use `injectScript` on a script which either returns a value or throws an exception; see that `injectScript` returns the value or rethrows the exception.

### Related Issue

#1754 proposes having a way to pass an `UnlistedScriptDefinition` rather than the filename to `injectScript`. If something like that was implemented, the type of the return value of `injectScript` could be based on the type of `main` within the `UnlistedScriptDefinition`.

### DIFF
--- a/packages/wxt/src/utils/inject-script.ts
+++ b/packages/wxt/src/utils/inject-script.ts
@@ -1,5 +1,6 @@
 /** @module wxt/utils/inject-script */
 import { browser } from 'wxt/browser';
+import { waitForScriptResultEvent } from './internal/script-result';
 
 export type ScriptPublicPath = Extract<
   // @ts-expect-error: PublicPath is generated per-project
@@ -16,7 +17,8 @@ export type ScriptPublicPath = Extract<
  * Make sure to add the injected script to your manifest's
  * `web_accessible_resources`.
  *
- * @returns A result object containing the created script element.
+ * @returns A result object containing the created script element and the return
+ * value of the script.
  */
 export async function injectScript(
   path: ScriptPublicPath,
@@ -39,6 +41,7 @@ export async function injectScript(
   // For MV2: Inline scripts execute synchronously when appended
   // For MV3: We need to wait for the load event
   const loadedPromise = isManifestV2 ? undefined : makeLoadedPromise(script);
+  const resultPromise = waitForScriptResultEvent(script);
 
   await options?.modifyScript?.(script);
 
@@ -49,9 +52,11 @@ export async function injectScript(
   }
 
   await loadedPromise;
+  const result = await resultPromise;
 
   return {
     script,
+    result,
   };
 }
 
@@ -100,4 +105,8 @@ export interface InjectScriptResult {
    * for them via `document.currentScript`.
    */
   script: HTMLScriptElement;
+  /**
+   * The return value of the script.
+   */
+  result: unknown;
 }
--- a/packages/wxt/src/utils/internal/script-result.ts
+++ b/packages/wxt/src/utils/internal/script-result.ts
@@ -1,0 +1,161 @@
+// The `MAIN` world script cannot use `browser.runtime.id`, but the uniqueness
+// of the event name is not an issue as the event is only dispatched on a
+// specific `script` element which is detached from the document.
+const SCRIPT_RESULT_EVENT_NAME = 'wxt:script-result';
+
+/**
+ * Dispatch a `CustomEvent` for a successful script result which carries a
+ * result value.
+ */
+export function dispatchScriptSuccessEvent(
+  target: EventTarget,
+  value: unknown,
+): boolean {
+  return dispatchScriptResultEvent(target, scriptSuccess(value));
+}
+
+/**
+ * Dispatch a `CustomEvent` for a failed script result which carries an `Error`
+ * value.
+ */
+export function dispatchScriptErrorEvent(
+  target: EventTarget,
+  error: Error,
+): boolean {
+  return dispatchScriptResultEvent(target, scriptError(error));
+}
+
+function dispatchScriptResultEvent(
+  target: EventTarget,
+  result: ScriptResult,
+): boolean {
+  return target.dispatchEvent(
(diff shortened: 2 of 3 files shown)


PatrykKuniczak:
@ion1 Are you have a plan to go on with that?

If yes, let's tell us, what type of advice you need, what info we can provide to let you through.

PatrykKuniczak:
@ion1 One more poke :)

ion1:
Hi,

Sorry I didn't manage to respond more quickly.

The issue is that code executed via `injectScript` needs to output its result value via communication through the DOM while code executed via `browser.scripting` needs to simply return it.

I'm not sure what would be the best way to handle the difference. Separate entry points? Some kind of reliable detection within a single entry point? I would appreciate your thoughts on this.

The draft just makes it work for `injectScript` while neglecting `browser.scripting`.

I am still interested in this feature, but unfortunately I cannot make any promises as to when I will have the energy to work on it next.

PatrykKuniczak:
@ion1 Thanks, i'll take a look on that, maybe today and try to give you my feedback.
But anyway we need approval for selected solution from @aklinker1

ion1:
Rebased on main.

packages/wxt/src/utils/internal/script-result.ts
```diff
@@ -0,0 +1,161 @@
...
+    result != null &&
+    typeof result === 'object' &&
+    'type' in result &&
+    typeof result.type === 'string' &&
+    ((result.type === 'success' && 'value' in result) ||
+      (result.type === 'error' &&
+        'error' in result &&
+        isDeconstructedError(result.error)))
+  );
+}
```
PatrykKuniczak, line 112:
```suggestion
  return (
    ((result.?type === 'success' && 'value' in result) ||
      (result.?type === 'error' &&
        'error' in result &&
        isDeconstructedError(result.error)))
  );
}
```
i hope this works the same way

packages/wxt/src/utils/inject-script.ts
```diff
@@ -100,4 +105,8 @@ export interface InjectScriptResult {
    * for them via `document.currentScript`.
    */
   script: HTMLScriptElement;
+  /**
+   * The return value of the script.
+   */
+  result: unknown;
```
PatrykKuniczak, line 111: should be typed
` A result object containing the created script element and the return value of the script.`

That's mean, you know at least that will be script, and you can pass e.g generic type param here, from that script return type

packages/wxt/src/virtual/unlisted-script-entrypoint.ts
```diff
@@ -1,42 +1,40 @@
...
-    throw err;
-  }
-  let result;
+// TODO: This only works with injectScript, not the scripting API.
+
+(async () => {
   try {
-    result = definition.main();
+    const script = document.currentScript;
+    if (script === null) {
```
PatrykKuniczak, line 14:
```suggestion
    if (document.?currentScript) {
```

packages/wxt/src/utils/internal/script-result.ts
PatrykKuniczak, line 1: This should be split to at least 3 files.
1 for dispatch helpers, 1 for errors and 1 "main"

PatrykKuniczak (changes requested):
Overall, before those changes, you need advice from @aklinker1
Because i don't have that much knowledge about project and don't want to make that big decision.

If you'll have suggestion/approval from aklinker and make necessary changes(if needed), i'll review it more precisely 😄

packages/wxt/src/utils/internal/script-result.ts
aklinker1: 🤷 one file is fine with me

aklinker1 (commented):
Alright, read through it. Makes sense, message passing is really the only way to do this.

Instead of stuffing this into `injectScript`, maybe we build a new util on top of it. If people want to use it, they can. If they don't, it's tree-stripped out.

```ts
import { evalScript } from 'eval-script/parent'

const res = await evalScript("/unlisted.js", ...args)
console.log(res) // "something"
```

```ts
import { defineEvalScript } from 'eval-script/child'

import defineEvalScript((...args) => {
  return "something";
})
```

That seems like the best approach to me, but the bigger question is if I want to be responsible for maintaining this... I don't think I do. Given how long it took for me to respond to this PR lol. Gotta be picky.

If there's interest, I might be willing to make it an official WXT package, like `@wxt-dev/storage`. But at least in my use-cases, it just doesn't seem like a util most extensions would need. But I could be wrong.

So how about this... Create an issue for this featur. If it gets, let's say, 50 👍 or other positive reactions I'll be down to add the package for it, maybe even add it to the core `wxt`.

> If you're willing to own the util and fix bugs or implement other features, that would help me out a ton too.

packages/wxt/src/utils/internal/script-result.ts
PatrykKuniczak: @aklinker1 OOOOOO right.

If this massive ifs, will be shortify this can be fine also for me.

PatrykKuniczak:
@ion1 Let's open https://github.com/wxt-dev/wxt/discussions and link this PR, it can be easier to gain 50 👍

PatrykKuniczak:
@ion1 Closing, if you'll open [discussion](https://github.com/wxt-dev/wxt/discussions) and recived 50 votes, feel free to link it there and reopen PR 😄

Issue #1752, opened 2025-06-17
Reduce users' fingerprintability


ion1:
### Feature Request

There are some things websites could use to get some bits of information to identify users with an extension which runs a content script on the site:
- The script uses `postMessage` to stop an old version of itself. A website could listen for `message` events and detect differences between users.
  
  `postMessage` could be substituted with a custom `window` event whose name includes a unique string (such as the ID of the extension). Then the website would at least need to check for the specific extension separately rather than just passively listening for `message` events.
- Content scripts injected to the `MAIN` world add a property to the `window` object. A website could e.g. hash the keys of `window` and get different values for users with an otherwise identical browser depending on whether they have an extension installed.
  
  `defineUnlistedScript` could perhaps wrap its result in an IIFE. `injectScript` does not seem to return the script's return value anyway, so another way to communicate is needed if that is desired.
- A `MAIN` world content script which does nothing else to make the web page aware of its existence but throws an uncaught exception will cause `console.error` to be called, which a site could detect.
  
  This one probably doesn't warrant changes in the code as a developer concerned about that can just wrap their `main` in a try-catch.

#### Is your feature request related to a bug?

N/A

### What are the alternatives?

Leaving everything as-is.

### Additional context

N/A

PR #1778, opened 2025-06-30
Content script context event

sleekslush:
### Related Issue

### DIFF
--- a/packages/wxt/src/utils/content-script-context.ts
+++ b/packages/wxt/src/utils/content-script-context.ts
@@ -39,27 +39,14 @@ import { createLocationWatcher } from './internal/location-watcher';
  * ```
  */
 export class ContentScriptContext implements AbortController {
-  private static SCRIPT_STARTED_MESSAGE_TYPE = getUniqueEventName(
-    'wxt:content-script-started',
-  );
-
-  private isTopFrame = window.self === window.top;
   private abortController: AbortController;
   private locationWatcher = createLocationWatcher(this);
-  private receivedMessageIds = new Set<string>();
 
   constructor(
     private readonly contentScriptName: string,
     public readonly options?: Omit<ContentScriptDefinition, 'main'>,
   ) {
     this.abortController = new AbortController();
-
-    if (this.isTopFrame) {
-      this.listenForNewerScripts({ ignoreFirstEvent: true });
-      this.stopOldScripts();
-    } else {
-      this.listenForNewerScripts();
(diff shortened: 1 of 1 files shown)


Timeraa:
Oops did not want to assign copilot to this, could you add more context to this (haha funny coincidence) to better understand this?

sleekslush:
> Oops did not want to assign copilot to this, could you add more context to this (haha funny coincidence) to better understand this?

Yea I don't care if this PR gets selected, but the problem that we're facing is this.

Every content script injection will cause this class to initiate a `window.postMessage` on each page load that it's whitelisted to. Our extension injects on `<all_urls>` for a variety of reasons. This normally isn't a huge problem, but we have encountered some customer complaints where the JS on the page is listening for window messages, checks the origin correctly, but then doesn't inspect the shape of the message itself before using it. This causes the page to basically break in a variety of cases and we have no way to disable this behavior from wxt, even when we aren't using it.

This PR is a different approach, maybe not the correct one, but basically defers removal of these event handlers to the next time the events fire. This means it's not immediate, but there's no window messaging involved and it solves the former issues I described.

If this isn't something we want to change or do, that's totally fair, but then at a minimum, we should have some way in wxt to prevent this `window.postMessage` behavior from happening if we aren't using the context object.

ion1:
> Every content script injection will cause this class to initiate a `window.postMessage` on each page load that it's whitelisted to. Our extension injects on `<all_urls>` for a variety of reasons. This normally isn't a huge problem, but we have encountered some customer complaints where the JS on the page is listening for window messages, checks the origin correctly, but then doesn't inspect the shape of the message itself before using it. This causes the page to basically break in a variety of cases and we have no way to disable this behavior from wxt, even when we aren't using it.

Using `postMessage` also makes it [easier to fingerprint the users of the extension](https://github.com/wxt-dev/wxt/issues/1752).

One alternative would be using a custom event based on the extension ID on the window object. Websites would at least need to check for every extension by their ID separately to fingerprint users.

PatrykKuniczak:
@sleekslush Are you want to go on with that?

PatrykKuniczak:
@aklinker1 Are you even want that change, if not close as stale.
Else let's told us here, maybe somebody will want to implement that

sleekslush:
The problem described in this PR is still an issue , but we currently patch it locally to disable this behavior completely. I don't have time to address this properly for WXT at the moment, so I'm going to close this.

PR #1787, opened 2025-07-02
Feat/Use filter for rolldown plugin #1716

### LINKED ISSUE #1716: [Rolldown] Use `filter` pattern for Vite hooks
### Feature Request

In some plugins WXT uses, we use `resolveId`, `load`, and `transform` to modify specfic files. We need to use [the `filter` feature](https://rolldown.rs/guide/plugin-development#plugin-hook-filters) to take advantage of rust.

The plugins that need updated are in the [`packages/wxt/src/core/builders/vite/plugins` directory](https://github.com/wxt-dev/wxt/tree/main/packages/wxt/src/core/builders/vite/plugins).

There are not unit tests for each of these plugins, but they are covered by E2E tests. So if `pnpm test` passes after your changes, nothing is broken.

Apart of #1677

Leave it as is and not take advantage of the native rust code.

PatrykKuniczak:
### Overview

I've adapt `filters` instead of making if statements and handling it by JS, now it should be handled by rust and make WXT more amazing ❤️ 

### Manual Testing

Let's run `pnpm test` and check if everything works fine :)

### Related Issue

This PR closes #1716

### DIFF
--- a/packages/wxt/src/core/builders/vite/plugins/devHtmlPrerender.ts
+++ b/packages/wxt/src/core/builders/vite/plugins/devHtmlPrerender.ts
@@ -40,32 +40,32 @@ export function devHtmlPrerender(
       },
       // Convert scripts like src="./main.tsx" -> src="http://localhost:3000/entrypoints/popup/main.tsx"
       // before the paths are replaced with their bundled path
-      transform(code, id) {
-        if (
-          config.command !== 'serve' ||
-          server == null ||
-          !id.endsWith('.html')
-        )
-          return;
+      transform: {
+        filter: {
+          id: /\.html$/,
+        },
+        handler(code, id) {
+          if (config.command !== 'serve' || server == null) return;
 
-        const { document } = parseHTML(code);
+          const { document } = parseHTML(code);
 
-        const _pointToDevServer = (querySelector: string, attr: string) =>
-          pointToDevServer(config, server, id, document, querySelector, attr);
-        _pointToDevServer('script[type=module]', 'src');
-        _pointToDevServer('link[rel=stylesheet]', 'href');
+          const _pointToDevServer = (querySelector: string, attr: string) =>
+            pointToDevServer(config, server, id, document, querySelector, attr);
+          _pointToDevServer('script[type=module]', 'src');
+          _pointToDevServer('link[rel=stylesheet]', 'href');
 
-        // Add a script to add page reloading
-        const reloader = document.createElement('script');
-        reloader.src = htmlReloadId;
-        reloader.type = 'module';
-        document.head.appendChild(reloader);
+          // Add a script to add page reloading
+          const reloader = document.createElement('script');
+          reloader.src = htmlReloadId;
+          reloader.type = 'module';
+          document.head.appendChild(reloader);
 
-        const newHtml = document.toString();
-        config.logger.debug('transform ' + id);
-        config.logger.debug('Old HTML:\n' + code);
-        config.logger.debug('New HTML:\n' + newHtml);
-        return newHtml;
+          const newHtml = document.toString();
+          config.logger.debug('transform ' + id);
+          config.logger.debug('Old HTML:\n' + code);
+          config.logger.debug('New HTML:\n' + newHtml);
+          return newHtml;
+        },
       },
 
       // Pass the HTML through the dev server to add dev-mode specific code
@@ -113,29 +113,35 @@ export function devHtmlPrerender(
     {
       name: 'wxt:virtualize-inline-scripts',
       apply: 'serve',
-      resolveId(id) {
-        // Resolve inline scripts
-        if (id.startsWith(virtualInlineScript)) {
-          return '\0' + id;
-        }
+      resolveId: {
+        filter: {
+          id: {
+            include: [new RegExp(`^${virtualInlineScript}`), /^\/chunks\//],
+          },
+        },
+        handler(id) {
+          // Resolve inline scripts
+          if (id.startsWith(virtualInlineScript)) {
+            return '\0' + id;
+          }
 
-        // Ignore chunks during HTML file pre-rendering
-        if (id.startsWith('/chunks/')) {
-          return '\0noop';
-        }
+          // Ignore chunks during HTML file pre-rendering
+          if (id.startsWith('/chunks/')) {
+            return '\0noop';
+          }
+        },
       },
-      load(id) {
-        // Resolve virtualized inline scripts
-        if (id.startsWith(resolvedVirtualInlineScript)) {
-          // id="virtual:wxt-inline-script?<key>"
+      load: {
+        filter: {
+          id: {
+            include: [new RegExp(`^${resolvedVirtualInlineScript}`)],
+            exclude: '\0noop',
+          },
+        },
+        handler(id) {
           const key = id.substring(id.indexOf('?') + 1);
           return inlineScriptContents[key];
-        }
-
-        // Ignore chunks during HTML file pre-rendering
-        if (id === '\0noop') {
-          return '';
-        }
+        },
       },
     },
   ];
--- a/packages/wxt/src/core/builders/vite/plugins/download.ts
+++ b/packages/wxt/src/core/builders/vite/plugins/download.ts
@@ -12,15 +12,22 @@ import { fetchCached } from '../../../utils/network';
 export function download(config: ResolvedConfig): Plugin {
   return {
     name: 'wxt:download',
-    resolveId(id) {
-      if (id.startsWith('url:')) return '\0' + id;
+    resolveId: {
+      filter: {
+        id: /^url:/,
+      },
+      handler(id) {
+        return `\0${id}`;
+      },
     },
-    async load(id) {
-      if (!id.startsWith('\0url:')) return;
-
-      // Load file from network or cache
-      const url = id.replace('\0url:', '');
-      return await fetchCached(url, config);
+    load: {
+      filter: {
+        id: /^\x00url:/,
+      },
+      async handler(id) {
+        const url = id.replace('\0url:', '');
+        return await fetchCached(url, config);
+      },
     },
   };
 }
--- a/packages/wxt/src/core/builders/vite/plugins/extensionApiMock.ts
+++ b/packages/wxt/src/core/builders/vite/plugins/extensionApiMock.ts
@@ -32,11 +32,21 @@ export function extensionApiMock(config: ResolvedConfig): vite.PluginOption {
         },
       };
     },
-    resolveId(id) {
-      if (id.endsWith(virtualSetupModule)) return resolvedVirtualSetupModule;
+    resolveId: {
+      filter: {
+        id: new RegExp(`${virtualSetupModule}$`),
+      },
+      handler() {
+        return resolvedVirtualSetupModule;
+      },
     },
-    load(id) {
-      if (id === resolvedVirtualSetupModule) return setupTemplate;
+    load: {
+      filter: {
+        id: new RegExp(`^${resolvedVirtualSetupModule}$`),
+      },
+      handler() {
+        return setupTemplate;
+      },
     },
   };
 }
--- a/packages/wxt/src/core/builders/vite/plugins/noopBackground.ts
+++ b/packages/wxt/src/core/builders/vite/plugins/noopBackground.ts
@@ -10,13 +10,21 @@ export function noopBackground(): Plugin {
   const resolvedVirtualModuleId = '\0' + virtualModuleId;
   return {
     name: 'wxt:noop-background',
-    resolveId(id) {
-      if (id === virtualModuleId) return resolvedVirtualModuleId;
+    resolveId: {
+      filter: {
+        id: new RegExp(`^${virtualModuleId}$`),
+      },
+      handler() {
+        return resolvedVirtualModuleId;
+      },
     },
-    load(id) {
-      if (id === resolvedVirtualModuleId) {
+    load: {
+      filter: {
+        id: new RegExp(`^${resolvedVirtualModuleId}$`),
+      },
+      handler() {
         return `import { defineBackground } from 'wxt/utils/define-background';\nexport default defineBackground(() => void 0)`;
-      }
+      },
     },
   };
 }
--- a/packages/wxt/src/core/builders/vite/plugins/removeEntrypointMainFunction.ts
+++ b/packages/wxt/src/core/builders/vite/plugins/removeEntrypointMainFunction.ts
@@ -16,14 +16,15 @@ export function removeEntrypointMainFunction(
     name: 'wxt:remove-entrypoint-main-function',
     transform: {
       order: 'pre',
-      handler(code, id) {
-        if (id === absPath) {
-          const newCode = removeMainFunctionCode(code);
-          config.logger.debug('vite-node transformed entrypoint', path);
-          config.logger.debug(`Original:\n---\n${code}\n---`);
-          config.logger.debug(`Transformed:\n---\n${newCode.code}\n---`);
-          return newCode;
-        }
+      filter: {
+        id: new RegExp(`^${absPath}$`),
+      },
+      handler(code) {
+        const newCode = removeMainFunctionCode(code);
+        config.logger.debug('vite-node transformed entrypoint', path);
+        config.logger.debug(`Original:\n---\n${code}\n---`);
+        config.logger.debug(`Transformed:\n---\n${newCode.code}\n---`);
+        return newCode;
       },
     },
   };
--- a/packages/wxt/src/core/builders/vite/plugins/resolveAppConfig.ts
+++ b/packages/wxt/src/core/builders/vite/plugins/resolveAppConfig.ts
@@ -22,15 +22,25 @@ export function resolveAppConfig(config: ResolvedConfig): vite.Plugin {
         },
       };
     },
-    async resolveId(id) {
-      if (id !== virtualModuleId) return;
-
-      return (await exists(appConfigFile))
-        ? appConfigFile
-        : resolvedVirtualModuleId;
+    resolveId: {
+      filter: {
+        id: {
+          exclude: new RegExp(`^${virtualModuleId}$`),
+        },
+      },
+      async handler() {
+        return (await exists(appConfigFile))
+          ? appConfigFile
+          : resolvedVirtualModuleId;
+      },
     },
-    load(id) {
-      if (id === resolvedVirtualModuleId) return `export default {}`;
+    load: {
+      filter: {
+        id: new RegExp(`^${resolvedVirtualModuleId}$`),
+      },
+      handler() {
+        return `export default {}`;
+      },
     },
   };
 }
--- a/packages/wxt/src/core/builders/vite/plugins/wxtPluginLoader.ts
+++ b/packages/wxt/src/core/builders/vite/plugins/wxtPluginLoader.ts
@@ -14,33 +14,50 @@ export function wxtPluginLoader(config: ResolvedConfig): vite.Plugin {
 
   return {
     name: 'wxt:plugin-loader',
-    resolveId(id) {
-      if (id === virtualModuleId) return resolvedVirtualModuleId;
-      if (id === virtualHtmlModuleId) return resolvedVirtualHtmlModuleId;
+    resolveId: {
+      filter: {
+        id: [
+          new RegExp(`^${virtualModuleId}$`),
+          new RegExp(`^${virtualHtmlModuleId}$`),
+        ],
+      },
+      handler(id) {
+        if (id === virtualModuleId) {
+          return resolvedVirtualModuleId;
+        } else {
+          return resolvedVirtualHtmlModuleId;
+        }
+      },
     },
-    load(id) {
-      if (id === resolvedVirtualModuleId) {
-        // Import and init all plugins
-        const imports = config.plugins
-          .map(
-            (plugin, i) =>
-              `import initPlugin${i} from '${normalizePath(plugin)}';`,
-          )
-          .join('\n');
-        const initCalls = config.plugins
-          .map((_, i) => `  initPlugin${i}();`)
-          .join('\n');
-        return `${imports}\n\nexport function initPlugins() {\n${initCalls}\n}`;
-      }
-      if (id === resolvedVirtualHtmlModuleId) {
-        return `import { initPlugins } from '${virtualModuleId}';
-
-try {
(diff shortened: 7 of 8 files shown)


packages/wxt/src/core/builders/vite/plugins/devHtmlPrerender.ts
```diff
@@ -113,29 +113,35 @@ export function devHtmlPrerender(
...
-        }
-
-        // Ignore chunks during HTML file pre-rendering
-        if (id.startsWith('/chunks/')) {
-          return '\0noop';
-        }
+      resolveId: {
+        filter: {
+          id: {
+            include: [new RegExp(`^${virtualInlineScript}`), /^\/chunks\//],
```
ion1: Since `virtualInlineScript` doesn't seem to be a regexp, you might want to use [RegExp.escape](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/escape).

PatrykKuniczak: I'll take a look on that today 🙂

PatrykKuniczak: @ion1 
image

As you can see, it doesn't work, because of error.

I see this feature is new(2025) and it shouldn't work with our configuration.

If you have any idea share it :)

PatrykKuniczak: But without that, it works, tests pass.

ion1: Many of the strings in the PR are constants for which a constant RegExp could be used as well. But not all of them are.

In particular, `absPath` within `removeEntrypointMainFunction` seems to be based on user-controlled path names.

The plugin API also seems to support picomatch patterns, but picomatch doesn't seem to come with an escape function for its own syntax.

Perhaps `absPath` could be used as a picomatch string rather than a regexp with the assumption that the paths are less likely to contain picomatch special characters than regexp ones (such as dots).

PatrykKuniczak: Ok, but we need to change it, if it works?

ion1: > Ok, but we need to change it, if it works?

I'm just worried about interpreting arbitrary filenames as regexps when the purpose is just to do a string match. In all fairness, the filenames come from a trusted user so it's not too scary.

Additionally, I'm a nobody. The opinion of the project authors is what matters.

> But without that, it works, tests pass.

For what it's worth, the test suite is probably not testing adversarial inputs assuming things are being interpreted as regexps when they weren't before.

PatrykKuniczak: > > Ok, but we need to change it, if it works?
> 
> I'm just worried about interpreting arbitrary filenames as regexps when the purpose is just to do a string match. In all fairness, the filenames come from a trusted user so it's not too scary.
> 
> Additionally, I'm a nobody. The opinion of the project authors is what matters.
> 
> > But without that, it works, tests pass.
> 
> For what it's worth, the test suite is probably not testing adversarial inputs assuming things are being interpreted as regexps when they weren't before.

If you have solution, feel free to share it to me here, otherwise let's wait for owner review 😁

PatrykKuniczak:
Small mistake of me, i've done `id` checking the other way round, let's run CI one more time, should work that time :)

packages/wxt/src/core/builders/vite/plugins/devHtmlPrerender.ts
```diff
@@ -113,29 +113,30 @@ export function devHtmlPrerender(
...
       },
-      load(id) {
-        // Resolve virtualized inline scripts
-        if (id.startsWith(resolvedVirtualInlineScript)) {
-          // id="virtual:wxt-inline-script?<key>"
+      load: {
+        filter: {
+          id: {
+            include: [new RegExp(`^${resolvedVirtualInlineScript}`)],
+            exclude: '\0noop',
```
sm17p: Your include set has no union with exclude set, so it's safe to remove this exclude filter

packages/wxt/src/core/builders/vite/plugins/devHtmlPrerender.ts
```diff
@@ -113,29 +113,30 @@ export function devHtmlPrerender(
...
-          return '\0' + id;
-        }
-
-        // Ignore chunks during HTML file pre-rendering
-        if (id.startsWith('/chunks/')) {
-          return '\0noop';
-        }
+      resolveId: {
+        filter: {
+          id: [new RegExp(`^${virtualInlineScript}`), /^\/chunks\//],
```
sm17p: You can probably remove `/^\/chunks\//` as you only care about returning `'\0' + id` if the filter matches

packages/wxt/src/core/builders/vite/plugins/download.ts
```diff
@@ -12,15 +12,22 @@ import { fetchCached } from '../../../utils/network';
...
     },
-    async load(id) {
-      if (!id.startsWith('\0url:')) return;
-
-      // Load file from network or cache
-      const url = id.replace('\0url:', '');
-      return await fetchCached(url, config);
+    load: {
+      filter: {
+        id: /^\x00url:/,
```
sm17p, line 27: This pattern doesn't match '\0url:'
```suggestion
        id: /^\\0url:/,
```

packages/wxt/src/core/builders/vite/plugins/wxtPluginLoader.ts
```diff
@@ -63,7 +80,7 @@ try {
           const newHead = document.createElement('head');
           document.documentElement.prepend(newHead);
         }
-        document.head.prepend(script);
+        document.head?.prepend(script);
```
sm17p, line 86: nitpick `document.head` exists at this point. Don't have to perform optional chaining

packages/wxt/src/core/builders/vite/plugins/devHtmlPrerender.ts
sm17p: > I'm just worried about interpreting arbitrary filenames as regexps when the purpose is just to do a string match. In all fairness, the filenames come from a trusted user so it's not too scary.
> 
+1 on the concern 

> The plugin API also seems to support picomatch patterns, but picomatch doesn't seem to come with an escape function for its own syntax.

> Perhaps absPath could be used as a picomatch string rather than a regexp with the assumption that the paths are less likely to contain picomatch special characters than regexp ones (such as dots).

but in this PR all the RegEx are one-to-one match of the previous checks. Maybe, we should slide it off this PR? This probably belongs in another issue as it's a concern pertaining to the current implementation. The changes in this PR don't diverge from original behaviour.

packages/wxt/src/core/builders/vite/plugins/devHtmlPrerender.ts
PatrykKuniczak: Oh yeah, that's true, i'll remove it. Thanks

packages/wxt/src/core/builders/vite/plugins/devHtmlPrerender.ts
PatrykKuniczak: That's also true, good job on your review :)

packages/wxt/src/core/builders/vite/plugins/download.ts
PatrykKuniczak: I have strange IDE error with that, but it looks fine right now, and i've change it to \\0 as you mentioned, i want to do it before :)

packages/wxt/src/core/builders/vite/plugins/wxtPluginLoader.ts
PatrykKuniczak: I've wondering why check looks like `document.head == null` and there's a problem i've changed it to `===` and now optional chaining isn't necessary, i want to do it before, but i wasn't sure why somebody want to check also `null` as string, i think it's mistake, which was resolved now :)

PatrykKuniczak:
@sm17p Thanks for your review, now code is a way simpler <3

packages/wxt/src/core/builders/vite/plugins/wxtPluginLoader.ts
sm17p: Uh that's probably because, of some JS magic. `a == null` is true, if a is either null or undefined so maybe that's the reason. This change might break it so better consult with the owner before this

packages/wxt/src/core/builders/vite/plugins/download.ts
PatrykKuniczak: Wait, wait, i can't use this, because of:
image


I need to use `/^\x00url:/` because only that works well

packages/wxt/src/core/builders/vite/plugins/wxtPluginLoader.ts
PatrykKuniczak: > Uh that's probably because, of some JS magic. `a == null` is true, if a is either null or undefined so maybe that's the reason. This change might break it so better consult with the owner before this
> 
> 

Tests passed, but overall owner need to take a look on these changes :)

packages/wxt/src/core/builders/vite/plugins/download.ts
sm17p: I reviewed `/^\x00url:/` on regexr again, and here are some insights

1. `\x00` is a hexadecimal for ASCII character null (char code '0'). Now, the first charachter that you want to match is actually a '\\' which is char code "92"
2. The second charachter that you want match with is zero or char code "48" or in regex hex terms `\x30`
3.  Either of `^\\\x30url:` or `^\\0url:` does the trick, they mean the same where first characther is '\\' and second is '0'

If the test cases are passing then there either must be something wrong in them or the setup behaviour is divregent from actual implementation, as `/^\x00url:/` shouldn't match with `\0url:` string

It's something that needs to be looked into IMO

PatrykKuniczak: @sm17p Both `\0` and `\x00` are `null`.
`The RegExp \0 Metacharacter in JavaScript is used to find the NULL character.`
image


Maybe i'm wrong, let's wait for owner/team review :)

sm17p: Yes, that's a good idea. Thanks for this!

packages/wxt/src/core/builders/vite/plugins/devHtmlPrerender.ts
```diff
@@ -113,29 +113,22 @@ export function devHtmlPrerender(
...
+        },
+        handler(id) {
           const key = id.substring(id.indexOf('?') + 1);
           return inlineScriptContents[key];
-        }
-
-        // Ignore chunks during HTML file pre-rendering
-        if (id === '\0noop') {
-          return '';
-        }
```
aklinker1: This needs to stay somehow. It dramatically speeds up dev startup speed by not bundling any of the JS code before writing the HTML files to disk.

packages/wxt/src/core/builders/vite/plugins/devHtmlPrerender.ts
```diff
@@ -113,29 +113,22 @@ export function devHtmlPrerender(
...
+          id: [new RegExp(`^${virtualInlineScript}`)],
+        },
+        handler(id) {
           return '\0' + id;
-        }
-
-        // Ignore chunks during HTML file pre-rendering
-        if (id.startsWith('/chunks/')) {
-          return '\0noop';
-        }
```
aklinker1: Same

packages/wxt/src/core/builders/vite/plugins/download.ts
aklinker1: As long as tests pass, that's fine. When I tried this originally, `\0` was throwing errors when building the regex string. If this works, that's fine 👍

packages/wxt/src/core/builders/vite/plugins/extensionApiMock.ts
```diff
@@ -32,11 +32,21 @@ export function extensionApiMock(config: ResolvedConfig): vite.PluginOption {
...
+      },
+      handler() {
+        return resolvedVirtualSetupModule;
+      },
     },
-    load(id) {
-      if (id === resolvedVirtualSetupModule) return setupTemplate;
+    load: {
+      filter: {
+        id: new RegExp(`^${resolvedVirtualSetupModule}$`),
```
aklinker1, line 46:
```suggestion
        id: resolvedVirtualSetupModule,
```

Is this the same or different? Does rolldown do a partial match for regular strings or a full match?

packages/wxt/src/core/builders/vite/plugins/noopBackground.ts
```diff
@@ -10,13 +10,21 @@ export function noopBackground(): Plugin {
   const resolvedVirtualModuleId = '\0' + virtualModuleId;
   return {
     name: 'wxt:noop-background',
-    resolveId(id) {
-      if (id === virtualModuleId) return resolvedVirtualModuleId;
+    resolveId: {
+      filter: {
+        id: new RegExp(`^${virtualModuleId}$`),
```
aklinker1, line 16: Same here. Does this work? No need to use regex if it's not needed.


```suggestion
        id: virtualModuleId,
```

packages/wxt/src/core/builders/vite/plugins/noopBackground.ts
```diff
@@ -10,13 +10,21 @@ export function noopBackground(): Plugin {
...
+      },
+      handler() {
+        return resolvedVirtualModuleId;
+      },
     },
-    load(id) {
-      if (id === resolvedVirtualModuleId) {
+    load: {
+      filter: {
+        id: new RegExp(`^${resolvedVirtualModuleId}$`),
```
aklinker1, line 24:
```suggestion
        id: resolvedVirtualModuleId,
```

packages/wxt/src/core/builders/vite/plugins/removeEntrypointMainFunction.ts
```diff
@@ -16,14 +16,15 @@ export function removeEntrypointMainFunction(
...
-      handler(code, id) {
-        if (id === absPath) {
-          const newCode = removeMainFunctionCode(code);
-          config.logger.debug('vite-node transformed entrypoint', path);
-          config.logger.debug(`Original:\n---\n${code}\n---`);
-          config.logger.debug(`Transformed:\n---\n${newCode.code}\n---`);
-          return newCode;
-        }
+      filter: {
+        id: new RegExp(`^${absPath}$`),
```
aklinker1, line 21:
```suggestion
        id: absPath,
```

packages/wxt/src/core/builders/vite/plugins/resolveAppConfig.ts
```diff
@@ -22,15 +22,23 @@ export function resolveAppConfig(config: ResolvedConfig): vite.Plugin {
...
     },
-    async resolveId(id) {
-      if (id !== virtualModuleId) return;
-
-      return (await exists(appConfigFile))
-        ? appConfigFile
-        : resolvedVirtualModuleId;
+    resolveId: {
+      filter: {
+        id: new RegExp(`^${virtualModuleId}$`),
```
aklinker1, line 28:
```suggestion
        id: virtualModuleId,
```

packages/wxt/src/core/builders/vite/plugins/resolveAppConfig.ts
```diff
@@ -22,15 +22,23 @@ export function resolveAppConfig(config: ResolvedConfig): vite.Plugin {
...
+        return (await exists(appConfigFile))
+          ? appConfigFile
+          : resolvedVirtualModuleId;
+      },
     },
-    load(id) {
-      if (id === resolvedVirtualModuleId) return `export default {}`;
+    load: {
+      filter: {
+        id: new RegExp(`^${resolvedVirtualModuleId}$`),
```
aklinker1, line 38:
```suggestion
        id: resolvedVirtualModuleId,
```

packages/wxt/src/core/builders/vite/plugins/wxtPluginLoader.ts
```diff
@@ -53,18 +70,20 @@ try {
...
 
-        const script = document.createElement('script');
-        script.type = 'module';
-        script.src = src;
+        if (!existing) {
+          const script = document.createElement('script');
+          script.type = 'module';
+          script.src = src;
+
+          if (document.head === null) {
```
aklinker1: Please revert unless it's important to the PR.

```suggestion
          if (document.head == null) {
```

aklinker1 (commented):
Looks good, thanks for figuring this out. Couple of comments, not sure if we need to use regex for exact matches.

PatrykKuniczak:
@aklinker1 As you can see it doesn't accept strings, i've wanted to do it in more simply way, but it isn't possible.

I had plan to check issues and if not open, open it on `rolldown` repo, but i haven't time, i'm able to do it now -_-

PatrykKuniczak:
@aklinker1 That's in some kind of FAQ:
`id is treated as a glob pattern when you pass a string, and treated as a regular expression when you pass a RegExp. In the resolve hook, id must be a RegExp. strings are not allowed. This is because the id value in resolveId is the exact text written in the import statement and usually not an absolute path, while glob patterns are designed to match absolute paths.
`

In the very bottom of:
https://rolldown.rs/guide/plugin-development#plugin-hook-filters

And i haven't understand it but Perplexity helps me 😆 
And that make sens, why Regexp is necessary for `resolveId` for rest i'm able to simplify it, but i need to check if it'll be working well, if tests pass.

packages/wxt/src/core/builders/vite/plugins/devHtmlPrerender.ts
PatrykKuniczak: No problem :)

packages/wxt/src/core/builders/vite/plugins/devHtmlPrerender.ts
PatrykKuniczak: No problem :)

packages/wxt/src/core/builders/vite/plugins/extensionApiMock.ts
PatrykKuniczak: TLDR;

We won't able to use strings as i've though in:
https://github.com/wxt-dev/wxt/pull/1787#issuecomment-3190878288

It's because we want to check direct string in imports, not resolved absolute path from bundling process.

We need to use RegExp everywhere as it's now.

I was trying to prettify it to make it with `/expression/`, but it wasn't working with vars, that's why i need to use `new RegExp()` syntax :)

PatrykKuniczak:
@aklinker1 Should be all good right now, let's approve workflow :)

PatrykKuniczak:
@aklinker1 let's check it if you'll have time, if you want to change sth there, i'll have time next week to work a little.

aklinker1:
I'll be available the next few weeks, running them now.

packages/wxt/src/core/builders/vite/plugins/devHtmlPrerender.ts
```diff
@@ -113,29 +113,35 @@ export function devHtmlPrerender(
...
+            new RegExp('^\x00noop'),
+          ],
+        },
+        handler(id) {
+          // Ignore chunks during HTML file pre-rendering
+          if (id === '\0noop') {
+            return '';
+          }
+
           const key = id.substring(id.indexOf('?') + 1);
```
aklinker1, line 145: Add this comment back

```suggestion
          // id="virtual:wxt-inline-script?<key>"
          const key = id.substring(id.indexOf('?') + 1);
```

packages/wxt/src/core/builders/vite/plugins/wxtPluginLoader.ts
```diff
@@ -14,33 +14,50 @@ export function wxtPluginLoader(config: ResolvedConfig): vite.Plugin {
...
-          .join('\n');
-        return `${imports}\n\nexport function initPlugins() {\n${initCalls}\n}`;
-      }
-      if (id === resolvedVirtualHtmlModuleId) {
-        return `import { initPlugins } from '${virtualModuleId}';
+    resolveId: {
+      filter: {
+        id: [
+          new RegExp(`^${virtualModuleId}$`),
+          new RegExp(`^${virtualHtmlModuleId}$`),
```
aklinker1, line 23: Random thought: It's interesting that regex objects can be passed into the native rust code...

aklinker1: What am I talking about lol, obviously it can be serialized and deserialized 🤦

packages/wxt/src/core/builders/vite/plugins/wxtPluginLoader.ts
```diff
@@ -53,18 +70,20 @@ try {
 
         const { document } = parseHTML(html);
         const existing = document.querySelector(`script[src='${src}']`);
-        if (existing) return;
 
-        const script = document.createElement('script');
-        script.type = 'module';
-        script.src = src;
+        if (!existing) {
```
aklinker1: I don't think this change is meaningful. Can it be reverted?

aklinker1 (changes requested):
OK, code looks looks good, only a few small comments, nothing major! 

Looks like there are some lint errors, checks are failing.

But tests are passing. I need to manually test dev mode, it's not well covered by unit tests.... it's something I keep putting off because there's a lot of setup and utils I need to create to unit test dev mode...

packages/wxt/src/core/builders/vite/plugins/wxtPluginLoader.ts
PatrykKuniczak: No probl3m, i'll do it after work :)

PatrykKuniczak: No probl3m, i'll do it after work :)

PatrykKuniczak: Done :)

PatrykKuniczak:
@aklinker1 Let's go with this.
It's complete.

aklinker1:
Didn't get to this this weekend, it's a high priority for me since Vite 8 is in beta.

PR #1792, opened 2025-07-04
fix(manifest): ensure manifest author is string, support object with email

### LINKED ISSUE #1770: Invalid Firefox manifest.json
### Describe the bug

When preparing an extension to be published through Firefox it fails Firefox's validation.

Image

Looking at the validation output it appears Firefox expects the "author" field for add-ons to be a string not an object that is defined by wxt:

https://github.com/wxt-dev/wxt/blob/7c783fe56524e5e0e5fa2599832eecf4f8765835/packages/browser/src/gen/index.d.ts#L9335C1-L9337C27

[MDN manifest.json Reference](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/author)

### Reproduction

https://github.com/justinforlenza/wxt-firefox-manifest

### Steps to reproduce

- Create a new wxt project
- set author under manifest in `wxt.config.ts` 
- npm install
- npm zip:firefox
- Upload to Firefox console

### System Info

```shell
System:
    OS: Linux 6.8 Ubuntu 24.04.2 LTS 24.04.2 LTS (Noble Numbat)
    CPU: (2) x64 AMD EPYC 7763 64-Core Processor
    Memory: 5.08 GB / 7.75 GB
    Container: Yes
    Shell: 5.2.21 - /bin/bash
  Binaries:
    Node: 22.16.0 - ~/nvm/current/bin/node
    Yarn: 1.22.22 - /usr/bin/yarn
    npm: 9.8.1 - ~/nvm/current/bin/npm
    pnpm: 10.11.0 - ~/nvm/current/bin/pnpm
  npmPackages:
    wxt: ^0.20.6 => 0.20.7
```

### Used Package Manager

npm

### Validations

hellozzm:
### Overview

Convert author field in manifest to string if it is an object, using the email property. Add test to verify author object is correctly converted to string in the generated manifest. Prevents invalid manifest format in Firefox and other browsers.

### Related Issue

This PR closes #1770


aiktb:
No browser other than Firefox needs the `author`. Some time ago Google removed this field from their MV3 manifest documentation reference.

> As of May 2025, Chrome and the Chrome Web Store don't use this key.
> Ref https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/author

I think these codes should be firefox only.

hellozzm:
Yes, before making the revisions, I consulted the documentation:
[https://developer.chrome.google.cn/docs/extensions/reference/manifest/author?hl=en](https://developer.chrome.google.cn/docs/extensions/reference/manifest/author?hl=en)

But perhaps this way of handling it can avoid reporting errors on Firefox

Timeraa:
Does this also apply to stores like Opera, Edge Addons?

hellozzm:
> Does this also apply to stores like Opera, Edge Addons?

image

As shown in the attached image, the Opera browser extension also encounters the same issue, so the solution used for Firefox can also be applied to Opera. Edge, like Chrome, does not have this problem.

Timeraa:
Hey @hellozzm I added some more tests and added a log as from my research as well no one uses the object version of this anymore and only Gecko based browsers use the plain string for display purposes. Could you review what I changed and give your feedback on it?

Timeraa (approved):
@aklinker1 would like your approval on this change, I don't consider it a breaking change since all browsers don't use it except gecko and we do migration  but even then it's not correctly set I think. So I'd like you to merge this

hellozzm:
> Hey @hellozzm I added some more tests and added a log as from my research as well no one uses the object version of this anymore and only Gecko based browsers use the plain string for display purposes. Could you review what I changed and give your feedback on it?

I’ve reviewed your changes—your test cases are much more comprehensive than mine. Overall, everything looks good now.

packages/browser/src/gen/index.d.ts
aklinker1, line 1: Revert this file, it will be overwritten next time the `@types/chrome` package changes. Make the change to that package instead of here.

aklinker1:
See https://github.com/wxt-dev/wxt/issues/1770#issuecomment-3148429940

Issue #1738, opened 2025-06-13
Using defineUnlistedScript for scripts injected into the main world overwrites a global variable


tstehr:
### Describe the bug

We're using `defineUnlistedScript` to create a script that is subsequently inserted into websites via the content script, allowing it to interact with the main world, like done in the [inject-script](https://github.com/wxt-dev/examples/tree/main/examples/inject-script) example.

We noticed that on particular websites having a script injected can lead to errors in page JavaScript. The problem happens even for empty scripts. 

Investigating the error points to the output generated by `defineUnlistedScript`. In [inject-script](https://github.com/wxt-dev/examples/tree/main/examples/inject-script) for example the generated file `injected.js` starts as follows:

```js
var injected=function(){<...>}();
```

This leads to a property `window.injected` being set to `undefined`, since that is the return type of the function. This overwrite can interfere with a page script that also uses `window.injected` (or another variable that matches the name of an unlisted script being injected).

### Reproduction

https://github.com/tstehr/repro-wxt-inject

### Steps to reproduce

1. Check out the test repo linked above
2. run `npm install` and then `npm run demo`
3. Run the [inject-script](https://github.com/wxt-dev/examples/tree/main/examples/inject-script) example in a browser. I used `pnpm dev:firefox`.
4. In that browser, visit the local server started in step 2.

Expected: The page displays "✅ The function has run"

Actual: The page continues to display "❌ The function has not yet run" and errors are shown in the developer console.

### System Info

```shell
System:
    OS: macOS 15.5
    CPU: (10) arm64 Apple M2 Pro
    Memory: 80.66 MB / 32.00 GB
    Shell: 4.0.1 - /opt/homebrew/bin/fish
  Binaries:
    Node: 23.10.0 - ~/.nodenv/shims/node
    Yarn: 1.22.22 - /opt/homebrew/bin/yarn
    npm: 10.9.2 - ~/.nodenv/shims/npm
    pnpm: 10.7.0 - /usr/local/bin/pnpm
  Browsers:
    Chrome: 137.0.7151.104
    Chrome Canary: 139.0.7236.0
    Firefox Nightly: 141.0a1
    Safari: 18.5
    Safari Technology Preview: 18.4
  npmPackages:
    wxt: ^0.20.5 => 0.20.5
```

### Used Package Manager

npm

### Validations

- [x] Read the [Contributing Guidelines](https://github.com/wxt-dev/wxt/blob/main/CONTRIBUTING.md).
[...]

PR #1793, opened 2025-07-05
fix(rolldown-compat): Only apply footer to content and unlisted scripts via plugin

### LINKED ISSUE #1717: [Rolldown] `esbuild.footer` is deprecated
### Feature Request

> `esbuild.footer` option was specified. But this option is deprecated and will be removed in future versions. This option can be achieved by using a plugin with transform hook, please use that instead.
> &ndash; https://github.com/wxt-dev/wxt/pull/1715

We need to create a custom plugin to add a footer. We currently use a footer to add code at the bottom of background, unlisted, and content scripts so they have a return value.

When we update to a plugin, we can fix some issues around the footer being applied too broadly. It should only be applied to unlisted and content scripts, not background or web workers.

- #942 
- Apart of https://github.com/wxt-dev/wxt/issues/1677

N/A

PatrykKuniczak:
### Overview

I hope that's it :)

### Manual Testing

Let's run `pnpm -F wxt-demo dev` and then check if `wxt-demo/.output/chrome-mv3-dev/content-scripts/automount.js` have `automount` on the very last line.

### Related Issue

This PR closes #1717


ion1:
A related issue: <https://github.com/wxt-dev/wxt/issues/1738>. The script does not need to set a global variable at all for the return value from the IIFE to be returned by executeScript.

PatrykKuniczak:
> A related issue: <https://github.com/wxt-dev/wxt/issues/1738>. The script does not need to set a global variable at all for the return value from the IIFE to be returned by executeScript.

If i'm understand you right, this plugin is unnecessary at all, yeah?

I can simply remove all of that, include esbuild prop from original solution, yes?

ion1:
We should probably get input from the project authors on this.

packages/wxt/src/core/builders/vite/plugins/iifeFooter.ts
```diff
@@ -0,0 +1,20 @@
...
+ * get's removed if added to `build.rollupOptions.output.footer`
+ * See https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/scripting/executeScript#return_value
+ */
+export function iifeFooter(iifeReturnValueName: string): Plugin {
+  return {
+    name: 'wxt:iife-footer',
+    generateBundle(_, bundle) {
+      for (const chunk of Object.values(bundle)) {
+        if (chunk.type === 'chunk' && chunk.isEntry) {
+          chunk.code += `${iifeReturnValueName};`;
```
sm17p, line 15: This works, but linking relevant issues for support at bundler level

1. https://github.com/rollup/rollup/issues/5509#issuecomment-2092109483
 2. > this.parse now only supports the allowReturnOutsideFunction option for now (https://github.com/rollup/rollup/pull/5073)
3. https://github.com/vitejs/vite/pull/14508

PatrykKuniczak: @sm17p I don't understand.
What are you want to achive, i need to adjust my code or what?

sm17p: What I found is that there is a rollup specific option called `allowReturnOutsideFunction` which does what this plugin is trying to do. 

The first issue has usage examples for areas where we need that option for code transformation.

The 2nd and 3rd are just supporting issues indicating that it's available for use for `Vite v6`.

**Why I mentioned it?**
- I thought if it's available at a bundler level, then it's better to use that to avoid maintenance overhead.

> i need to adjust my code or what?
- Upto the owner's descretion
- Normally, what I would do is check if I can get it working with already existing API but that's your call

PatrykKuniczak: @sm17p Now i see, what are you want to achive.

But this prop could be use only inside plugin context, i need to make plugin after all, especially if we want to run it under condition(Only for `content-script` and `unlisted-script`)

I was studying the structure of project and of building process, and there's no place do enable it for plugin of one of entrypoints, because it's fetching dynamically, and there's no plugin for e.g `content-script` in which plugin i can enable `allowReturnOutsideFunction`.

Maybe there's a chance. owner will answer, i'm still begginer in this project :)

Now it's simply IMO, but if it could be more simply, let's feel free to open PR to my branch with your solution or wait for owner answer, we knows more than us, about his product :)

PR #1806, opened 2025-07-13
Keep `content_script` `matches` unsorted in manifest

kachkaev:
### Overview

I’ve noticed that host names in `matches` end up being sorted. This affects what extension users see in the store, so sorting may be therefore undesired. For example, if we have

```
www.example.com
m.example.com
another-less-important-example.com
```

, users will end up being presented with

```
another-less-important-example.com
m.example.com
www.example.com
```

That's somewhat less logical than the original sequence.

### Manual Testing

I tested the changes locally via `pnpm patch`.

### Related Issue

I haven’t found or created one. The change is quite small so felt worth a PR right away. If you disagree with the idea, feel free to ignore this change. `pnpm patch` gives a good enough workaround, so host sorting is not blocking.

### DIFF
--- a/packages/wxt/src/core/utils/content-scripts.ts
+++ b/packages/wxt/src/core/utils/content-scripts.ts
@@ -39,7 +39,8 @@ export function hashContentScriptOptions(
     Object.entries(withDefaults)
       // Sort any arrays so their values are consistent
       .map<[string, unknown]>(([key, value]) => {
-        if (Array.isArray(value)) return [key, value.sort()];
+      // Keeping matches unsorted gives developers more control on what users see in extension stores
+      if (Array.isArray(value) && key !== "matches") return [key, value.sort()];
         else return [key, value];
       })
       // Sort all the fields alphabetically


Timeraa (approved):

aklinker1:
I'm good with this change as long it doesn't cause tests to become flakey in this repo (the only reason I sort everything lol, things start failing real fast if you don't ensure an order).

Patching works, but if you don't want to deal with conflicts with the patch when upgrading WXT, you could use the [`build:manifestGenerated` hook](https://wxt.dev/api/reference/wxt/interfaces/WxtHooks.html#build-manifestgenerated) in your config or a WXT module to update the manifest how you want: https://wxt.dev/guide/essentials/config/hooks.html

kachkaev:
Happy to make additional changes if needed. What would you suggest?

aklinker1:
Sorry for the delay getting back to you. After re-reviewing this PR 4 months later, it appears the only change now is removing the sorting in the `hashContentScriptOptions` function, which is not used to generate the manifest. Should the change go somewhere in the actual manifest generation code? 

https://github.com/wxt-dev/wxt/blob/adeef6eb73a245a5f2717532dc26113408d2bf42/packages/wxt/src/core/utils/manifest.ts#L50

Hashing needs to sort the entries to accurately check if two sets of options are different.

kachkaev:
👋 @aklinker1! If I understand correctly, manifest does not contain host `matches` by itself – these are grabbed from content script metadata. In here we hash metadata for individual scripts and iterate over them to call `addHostPermission`:

https://github.com/wxt-dev/wxt/blob/5c77cf23b85ed8c469f1ddbf5272b46b52032f38/packages/wxt/src/core/utils/manifest.ts#L380-L413

This explains why patching `hashContentScriptOptions` works for me (I can still confirm it).

After learning WXT a bit more, I now think that my solution is not generic and should not be used. The number of content scripts can vary and they can have different sets of hosts. So unless we build content scripts synchronously in alphabetical order, we can't get a deterministic list anyway. 

What one can do instead is using [a hook named `build:manifestGenerated`](https://wxt.dev/guide/essentials/config/hooks.html#adding-hooks). We can take an already sorted list of hosts and re-sort it how we want it. Importantly this list will be full, i.e. combined from individual content scripts.

Let's close this PR. I'm happy with using `build:manifestGenerated`, but if anyone has better thoughts, happy to try something else. Cheers!

PR #1823, opened 2025-07-24
fix: improve Chrome path search

kachkaev:
### Overview

When I tried `pnpm test` locally for #1806, I saw this error:

```
FAIL  src/__tests__/options.test.ts > Options > firefoxArgs > should log a warning when --remote-debugging-port is passed in
Error: Could not find "chrome" binary.

If it is installed in a custom location, you can specify the path with the browserPaths option.
 ❯ Module.resolveRunOptions src/options.ts:63:11
```

My `/Applications` folder contained `Google Chrome.app` instead of `Chrome.app` (it was installed via `homebrew` a while ago). Adding an alternative path to `KNOWN_BROWSER_PATHS` resolved the error. I also added the second path for Chrome Canary just in case.

### Manual Testing

`pnpm test` stopped showing the above error.

### Related Issue

NA

### DIFF
--- a/packages/runner/src/browser-paths.ts
+++ b/packages/runner/src/browser-paths.ts
@@ -30,7 +30,10 @@ export const KNOWN_BROWSER_PATHS: Record<
     windows: [],
   },
   chrome: {
-    mac: ['/Applications/Chrome.app/Contents/MacOS/Google Chrome'],
+    mac: [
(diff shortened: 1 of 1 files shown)


aklinker1:
I fixed checks on `main` this morning, merged to get your PR passing.

PR #1825, opened 2025-07-25
feat: Add overlay option for dev icons

zizzfizzix:
### Overview

Adds an icon overlay option in addition to grayscale in development environments.

>When an icon is black and white, the grayscale indicator doesn't cut it. This new one should be more obvious, and work for bw icons.

Here's the sample output:

Deprecates the previous option `grayscaleOnDevelopment` but is backward compatible, and retains the same default.

### Manual Testing

Add this option in your extension's `wxt.config.ts`

```
  autoIcons: {
    developmentIndicator: 'overlay',
  }
```

### DIFF
--- a/packages/auto-icons/README.md
+++ b/packages/auto-icons/README.md
@@ -5,7 +5,7 @@
 ## Features
 
 - Generate extension icons with the correct sizes
-- Make the icon greyscale during development
+- Make the icon greyscale or include a visible overlay during development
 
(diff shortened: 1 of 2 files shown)


zizzfizzix:
FYI @Timeraa

Timeraa (approved):

aklinker1:
Updated with latest `main` to fix checks.

PR #1826, opened 2025-07-25
feat: add experimental autoIncludeExternalSources option for monorepo support

### LINKED ISSUE #1505: Automatic gathering of used files for `zip`
### Feature Request

I added the extension as a child to another project, kind of like a monorepo. The problem is that when I `zip` for Firefox it only includes the source files for the extension, but not files that the extension imports from the parent (or sibling) project. A possible way to do this might be; since Vite/Rollup should be aware of which files actually end up in the compiled output, `zip` could follow the paths and add those files.

"N/A"

Adding the files manually via `defineConfig({ zip: {} })` which is very tedious and doesn't scale.

### Additional context

"N/A"

johnnyfekete:
First real contribution here, hope I understood the issue correctly 🙂

### Overview

Adds experimental `autoIncludeExternalSources` option to automatically include external source files in zip archives for monorepo setups.  
When enabled, WXT analyzes the build output to find imported files from outside the extension directory and includes them in the sources zip. This eliminates the need for manual configuration when extensions import from parent/sibling packages.

The feature is disabled by default for backward compatibility and marked as experimental.

### Manual Testing

1. Create a shared file outside your extension: `../shared-lib/utils.ts`
2. Import it in your extension: `import { util } from '../../../shared-lib/utils';`
3. Enable in config: `zip: { autoIncludeExternalSources: true }`
4. Run `wxt zip -b firefox`
5. Check sources zip contains the external file's contents

### Related Issue

This PR closes #1505

### DIFF
--- a/docs/guide/essentials/publishing.md
+++ b/docs/guide/essentials/publishing.md
@@ -143,6 +143,24 @@ export default defineConfig({
 });
 ```
 
+#### Monorepo Support (Experimental)
+
+If your extension is part of a monorepo and imports files from outside the extension directory (like shared libraries), you can enable automatic inclusion of these external files:
+
+```ts [wxt.config.ts]
+export default defineConfig({
+  zip: {
+    autoIncludeExternalSources: true, // EXPERIMENTAL
+  },
+});
+```
+
+When enabled, WXT will analyze your build output to find all imported files from outside the extension's source directory and automatically include them in the sources zip. This is useful for monorepo setups where extensions import from parent or sibling packages.
+
+:::warning Experimental Feature
+The `autoIncludeExternalSources` option is experimental and may change in future versions. Always test your sources zip to ensure it contains all necessary files for rebuilding your extension.
+:::
+
 If it's your first time submitting to the Firefox Addon Store, or if you've updated your project layout, always test your sources ZIP! The commands below should allow you to rebuild your extension from inside the extracted ZIP.
 
 :::code-group
--- a/docs/guide/essentials/publishing.md
+++ b/docs/guide/essentials/publishing.md
@@ -200,7 +200,7 @@ Depending on your package manager, the `package.json` in the sources zip will be
 WXT uses the command `npm pack <package-name>` to download the package. That means regardless of your package manager, you need to properly setup a `.npmrc` file. NPM and PNPM both respect `.npmrc` files, but Yarn and Bun have their own ways of authorizing private registries, so you'll need to add a `.npmrc` file.
 :::
 
-#### Monorepo Support (Experimental)
+#### Include External Sources (Experimental)
 
 If your extension is part of a monorepo and imports files from outside the extension directory (like shared libraries), you can enable automatic inclusion of these external files:
 
--- a/docs/guide/essentials/publishing.md
+++ b/docs/guide/essentials/publishing.md
@@ -143,24 +143,6 @@ export default defineConfig({
 });
 ```
 
-#### Monorepo Support (Experimental)
-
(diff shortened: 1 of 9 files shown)


docs/guide/essentials/publishing.md
```diff
@@ -200,6 +200,24 @@ Depending on your package manager, the `package.json` in the sources zip will be
 WXT uses the command `npm pack <package-name>` to download the package. That means regardless of your package manager, you need to properly setup a `.npmrc` file. NPM and PNPM both respect `.npmrc` files, but Yarn and Bun have their own ways of authorizing private registries, so you'll need to add a `.npmrc` file.
 :::
 
+#### Include External Sources (Experimental)
```
johnnyfekete, line 203: Not sure, if this is the right place to add the documentation 🤷‍♂️

packages/wxt/e2e/tests/zip.test.ts
aklinker1, line 1: These tests don't actually test the new behavior. Please actually include a external file in project, then extract and make sure the are included.

Here's an example of how to unzip and check for file existence: https://github.com/johnnyfekete/wxt/blob/d632e5c3cdd96421662b791c6d6008b32dde666e/packages/wxt/e2e/tests/zip.test.ts#L103-L123

docs/guide/essentials/publishing.md
aklinker1: Yeah, probably fine.

aklinker1 (changes requested):
Like the idea, just need to see a better E2E test.

packages/wxt/e2e/tests/zip.test.ts
johnnyfekete: You're right, I changed them. It was a lot more complicated than I thought, so maybe you can think of a nicer way to test files that are outside of the project folder, but I'm using the zip test functionalities that you recommended.

st3h3n:
This seems to generally work quite well, but I came across a conceptual problem:

Let's say the WXT root is in `monorepo/packages/wxt`. If you run `wxt zip` the resulting zip file will include the sources from the other packages in a `..` directory. But because of the [Zip Slip Vulnerability](https://security.snyk.io/research/zip-slip-vulnerability) most zip tools will ignore this directory when unzipping the file.

No problem, just use `sourcesRoot: '../'`, and in order to not include all packages: `{ excludeSources: '**', includeSources: ['wxt/**'] }`. But now the collected external sources will be omitted.

I think `excludeSources` should either ignore the auto collected sources, or be ignored, or there could be seperate controls for it. While this was also mentioned in [this post](https://github.com/wxt-dev/wxt/issues/1505#issuecomment-2817472470), I would disagree regarding the `includeSources` option being ignored, as there are common usecases like `includeSources: ['../tsconfig.json', '../package.json', ../assets/**]`.

st3h3n:
I've had some more time to look into this. I've found some files are missing:

1) A file which only contains a re-export: `export { foo as bar } from 'foobar';`
2) Files that only contain build-time content, so `.ts` files that only export interfaces or types. I would expect `.d.ts` to also be affected.
3) All `.html` and `.less` files. The reason for that is me using Angular. But the plugin that makes it Vite compatible uses Vite to resolve those files so a mechanism that hooks into Vite's resolver should still be able to catch these.

I'm a complete noob at Vite, hence why I didn't try to implement this feature myself, but I suspect that the approach of working from the build chunks backwards cannot solve this.

PatrykKuniczak:
@johnnyfekete Are you have plan to continue work?

aklinker1:
Gonna close this as stale. We can re-open if someone wants to continue working on it.

PR #1833, opened 2025-08-01
feat: Support `.wxtrc` config file

dword-design:
### Overview

Nuxt supports `.nuxtrc`, WXT can support `.wxtrc`. Or does something speak against it?

### DIFF
--- a/packages/wxt/e2e/tests/user-config.test.ts
+++ b/packages/wxt/e2e/tests/user-config.test.ts
@@ -142,14 +142,12 @@ describe('User Config', () => {
   it('should not throw error when config file not exist', async () => {
     const project = new TestProject();
     project.addFile(
-      'src/entrypoints/background.ts',
+      'entrypoints/background.ts',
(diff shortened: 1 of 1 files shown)


Timeraa (approved):
From my side I don't see why we wouldn't allow so but @aklinker1 would/should know if they did so on purpose

aklinker1:
Looks like a test is failing. If it's invalid and the expectation just needs to change, please update it. Compare the actual value to `main`'s actual value for reference.

dword-design:
I updated the test. Looks like the wording of the error changed or sth.

dword-design:
Now the tests should run through. The issue was that the condition if the config exists or not is different depending on if `rcFile` is set or not. I fiddled around with the `loadConfig` results and then asked myself if the check should actually be there at all. I mean why an error if the config file isn't there? So I removed the check and adjusted the test to expect to only throw the entrypoints error.

PR #1852, opened 2025-08-20
feat: allow setting `alias: false` to disable default aliases

aleclarson:
I want to use my own alias patterns without interference from wxt.

### DIFF
--- a/packages/wxt/e2e/tests/typescript-project.test.ts
+++ b/packages/wxt/e2e/tests/typescript-project.test.ts
@@ -397,6 +397,44 @@ describe('TypeScript Project', () => {
     expect(output).toContain('./example.ts');
   });
 
+  it('should allow disabling the default aliases', async () => {
+    const project = new TestProject();
+    project.addFile('src/entrypoints/unlisted.html', '<html></html>');
+    project.setConfigFileConfig({
+      srcDir: 'src',
+      alias: false,
+    });
+
+    await project.prepare();
+
+    const output = await project.serializeFile('.wxt/tsconfig.json');
+    expect(output).toMatchInlineSnapshot(`
+      ".wxt/tsconfig.json
+      ----------------------------------------
+      {
+        "compilerOptions": {
+          "target": "ESNext",
+          "module": "ESNext",
+          "moduleResolution": "Bundler",
+          "noEmit": true,
+          "esModuleInterop": true,
+          "forceConsistentCasingInFileNames": true,
+          "resolveJsonModule": true,
+          "strict": true,
+          "skipLibCheck": true,
+          "paths": {
+
+          }
+        },
+        "include": [
+          "../**/*",
+          "./wxt.d.ts"
+        ],
+        "exclude": ["../.output"]
+      }"
+    `);
+  });
+
   it('should set correct import.meta.env.BROWSER type based on targetBrowsers', async () => {
     const project = new TestProject();
     project.addFile('entrypoints/unlisted.html', '<html></html>');
--- a/packages/wxt/src/core/resolve-config.ts
+++ b/packages/wxt/src/core/resolve-config.ts
@@ -133,15 +133,18 @@ export async function resolveConfig(
     defaults: userConfig.webExt ?? userConfig.runner,
   });
   // Make sure alias are absolute
-  const alias = Object.fromEntries(
-    Object.entries({
-      ...mergedConfig.alias,
-      '@': srcDir,
-      '~': srcDir,
-      '@@': root,
-      '~~': root,
-    }).map(([key, value]) => [key, path.resolve(root, value)]),
-  );
+  const alias =
+    mergedConfig.alias !== false
+      ? Object.fromEntries(
+          Object.entries({
+            ...mergedConfig.alias,
+            '@': srcDir,
+            '~': srcDir,
+            '@@': root,
+            '~~': root,
+          }).map(([key, value]) => [key, path.resolve(root, value)]),
+        )
+      : {};
 
   let devServerConfig: ResolvedConfig['dev']['server'];
   if (command === 'serve') {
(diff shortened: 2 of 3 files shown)


aleclarson:
I've rebased and added a test.

packages/wxt/src/core/resolve-config.ts
```diff
@@ -133,15 +133,18 @@ export async function resolveConfig(
...
+      ? Object.fromEntries(
+          Object.entries({
+            ...mergedConfig.alias,
+            '@': srcDir,
+            '~': srcDir,
+            '@@': root,
+            '~~': root,
+          }).map(([key, value]) => [key, path.resolve(root, value)]),
+        )
+      : {};
```
PatrykKuniczak, line 147: Because alias can be `undefined`, you can do it like:

```suggestion
  const alias =
    mergedConfig.alias ?? Object.fromEntries(
          Object.entries({
            ...mergedConfig.alias,
            '@': srcDir,
            '~': srcDir,
            '@@': root,
            '~~': root,
          }).map(([key, value]) => [key, path.resolve(root, value)]),
        )
```

packages/wxt/e2e/tests/typescript-project.test.ts
```diff
@@ -397,6 +397,44 @@ describe('TypeScript Project', () => {
...
+          "module": "ESNext",
+          "moduleResolution": "Bundler",
+          "noEmit": true,
+          "esModuleInterop": true,
+          "forceConsistentCasingInFileNames": true,
+          "resolveJsonModule": true,
+          "strict": true,
+          "skipLibCheck": true,
+          "paths": {
+
```
PatrykKuniczak, line 426: This blank space is necessary?

PatrykKuniczak (changes requested):

packages/wxt/e2e/tests/typescript-project.test.ts
aleclarson: that is vitest's doing. generated snapshot

aleclarson: or it might be `project.serializeFile()` actually

packages/wxt/src/core/resolve-config.ts
aleclarson: i think the `alias` variable is meant to be an object. whereas `??` can result in `false`

packages/wxt/e2e/tests/typescript-project.test.ts
aklinker1: https://github.com/wxt-dev/wxt/blob/9470d2b6968300d9d4a0abfecc0f9a776c00c0cd/packages/wxt/src/core/generate-wxt-dir.ts#L246-L266

aklinker1 (commented):
@aleclarson Have you tried this by patching the changes into a real extension? I just ran the actions, you should be able to install this PR [via pkg.pr.new](https://github.com/wxt-dev/wxt/pull/1852#issuecomment-3882081195) now.

For some reason my gut is telling me WXT relies on some of these aliases internally and dev or production builds would crash, but I can't find those references in the code anymore... Maybe they were there and I've removed them naturally over time? Hmm 🤔 

Once we merge this, maintainers will need to keep in mind that the default aliases will no longer always exist.

The other thing to consider is that WXT is opinionated, if you don't like it's opinions... to bad? Convention over configuration - all WXT projects use the same aliases, and other frameworks like Nuxt use them too.

Can I ask what your preferred aliases are?

packages/wxt/src/core/resolve-config.ts
PatrykKuniczak: @aleclarson If this will be `false` everything will work as expected. let's read:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing

It works ONLY for `null` or `undefined`

packages/wxt/e2e/tests/typescript-project.test.ts
PatrykKuniczak: @aklinker1 Are you want to read this file, instead of redefining it, or only i'm don't understand, what do you mean? 😆

PatrykKuniczak:
@aklinker1 
~~If i'm right.
Those aliases are used not only in WXT(As lib), but only after init project on it?
If yes, anybody are able to overwrite it by passing empty object and there's no `us` aliases in their project, yeah??
In that case, there's no different on empty object and `false`, it's only syntaxis sugar(or sth on that shape), i'm right or completely no?~~

Don't mind

packages/wxt/src/core/resolve-config.ts
aleclarson: yeah that's how `??` works. for `a ?? b`, it results in `false` if `a === false`.

`ResolvedConfig` type expects `alias` to be an object, as i said.

https://github.com/wxt-dev/wxt/blob/2164a95b55ea0a091dfc02d0cc53efcc26192b0f/packages/wxt/src/types.ts#L1355-L1358

packages/wxt/e2e/tests/typescript-project.test.ts
aleclarson: if that will be fixed, it won't be in this PR

packages/wxt/src/core/resolve-config.ts
PatrykKuniczak: @aleclarson OKKK, yeahh my bad, i'm resolving this right now.

PatrykKuniczak:
@aklinker1 This is mergable?

aklinker1:
I want an answer before making this change, it will effect WXT devs and module authors, changing their assumptions for what generated code needs to look like.

PatrykKuniczak:

PR #1858, opened 2025-08-24
fix: should be youtube.injected

lookfirst:
Small typo fix

### DIFF
--- a/docs/guide/essentials/entrypoints.md
+++ b/docs/guide/essentials/entrypoints.md
@@ -106,7 +106,7 @@ Entrypoints must be zero or one levels deep for WXT to discover and build them:
    📂 youtube.content/ <!-- [!code ++] -->
       📄 index.ts <!-- [!code ++] -->
       📄 ... <!-- [!code ++] -->
-   📂 youtube-injected/ <!-- [!code ++] -->
+   📂 youtube.injected/ <!-- [!code ++] -->
       📄 index.ts <!-- [!code ++] -->
       📄 ... <!-- [!code ++] -->
 ```


aklinker1:
Could be, but it doesn't have to be. Either name works as an unlisted script.

jon-hotaisle:
@aklinker1 Where is that documented then? It was a source of confusion for me.

lookfirst:
@aklinker1 Where is that documented then? It was a source of confusion for me.

aklinker1:
- https://wxt.dev/guide/essentials/content-scripts.html#isolated-world-vs-main-world
- https://wxt.dev/api/reference/wxt/utils/inject-script/functions/injectScript.html#function-injectscript
- https://wxt.dev/guide/essentials/entrypoints.html#unlisted-scripts

lookfirst:
@aklinker1 Thanks, but I don't see where it says `"-"` vs. `"."` is ok in the name of the folder. That's where the confusion stems.

aklinker1:
> @aklinker1 Thanks, but I don't see where it says `"-"` vs. `"."` is ok in the name of the folder. That's where the confusion stems.

@lookfirst The docs state that you use `injectScript` with an unlisted script. Unlisted script paths can look like this:

- `entrypoints/{name}.[jt]sx?`
- `entrypoints/{name}/index.[jt]sx?`

The `{name}` can be anything you want. It can include alphanumerics, periods, dashes, other punctuation, etc. It should all work.

`.background` and `.content` are the only recognized dot suffixes for JS files, corresponding with [background](https://wxt.dev/guide/essentials/entrypoints.html#background) and [content script](https://wxt.dev/guide/essentials/entrypoints.html#content-scripts) entrypoints. Any other JS file matching the unlisted script path patterns above will be categorized as an unlisted script.

I didn't document what could go in the `{name}` because it can be anything. Not sure where a good place for that would be, it's true for all types of entrypoints.

aklinker1:
You can see the type of entrypoint your files are categorized by running the following and searching your console for "All entrypoints"

```sh
wxt prepare --debug
```

lookfirst:
@aklinker1 

> `I didn't document what could go in the {name} because it can be anything.`

The easiest way to avoid any sort of confusion would be to document that {name} can be anything.

PR #1881, opened 2025-09-04
fix: input_components is supported by mv3

eagleoflqj:
### Overview

Guess I'm the first one trying to use wxt to build an input method for ChromeOS.
The `input_components` should be kept in manifest.
It's documented by Google https://developer.chrome.com/docs/extensions/reference/manifest and used by an input method developed by Google https://github.com/google/jscin/blob/4fd228d2ef445ed615d8f9a5217a1dfa867ba6db/src/chrome/manifest.json#L18

### Related Issue


aklinker1:
Awesome, thanks for the PR! Haven't dealt much with ChromeOS

PR #1884, opened 2025-09-08
fix: `wxt prepare` fails with the error "__vite_ssr_exportName__ is not defined" when using Vite 7

aiktb:
### Related Issue


aklinker1 (approved):
> Can vite-node 2 be removed from scope?

~~Not for now, no. Since WXT supports vite 5-7, we need vite-node 2 for vite 5~~

Yes, we can. Seems like vite-node 3.2.5 supports vite ≥ 5!

https://www.npmjs.com/package/vite-node/v/3.2.4?activeTab=code

aiktb:
> > Can vite-node 2 be removed from scope?
> 
> ~Not for now, no. Since WXT supports vite 5-7, we need vite-node 2 for vite 5~
> 
> Yes, we can. Seems like vite-node 3.2.5 supports vite ≥ 5!
> 
> https://www.npmjs.com/package/vite-node/v/3.2.4?activeTab=code

I manually tested vite@5 and vite-node@3 and found no issues, so I removed vite-node@2 from the dependencies.
Requires re-approval

oeyoews:
It works for me. Thanks!

aklinker1:
Thanks for verifying!

PR #1899, opened 2025-09-16
feat: add `onBeforeMount` callback to `createIframeUi` options

bohuslavsemenov:
### Overview

This PR introduces an `onBeforeMount` callback to the `createIframeUi` function. The main motivation is that the `allow` attribute must be set before the iframe is injected into the DOM, otherwise browser will not include permission policies specified in it.

### Manual Testing

1. Go to the `wxt-demo` project.
2. Add the following code to `entrypoints/iframe-src/index.html`
```ts
navigator.mediaDevices.getUserMedia({
  audio: true,
  video: true,
});
```
3. Add the `onMount` callback to set `allow` attribute to `iframe.content.ts`:
```ts
onMount: (wrapper, iframe) => {
  iframe.setAttribute('allow', 'camera;microphone');
}
```
4. Go to google.com and try to refresh the page. You will get inconsistent permission prompts (Actually, they will work only when iframe is hot reloaded or on "reload iframe" click in browser context menu).
5. Change `onMount` to `onBeforeMount` and it will prompt you every time you refresh the page.


blackwolf3342:
@aklinker1 Pls can you review it, since we're having with media permissions

aklinker1 (approved):

PR #1907, opened 2025-09-19
fix: Don't return promises from unlisted scripts that do not have an async `main` function

### LINKED ISSUE #1906: bug: Scripts defined with defineUnlistedScript always return a Promise
### Describe the bug

For example, the following inject.ts script will be executed using `browser.devtools.inspectedWindow.eval`.

```ts
  // @ts-expect-error
  if (!globalThis.$0) {
    return
  }
  // @ts-expect-error
  return globalThis.$0.childElementCount
})
```

The expected return value is number/undefined, but it always returns `{}`. After careful inspection, I found this script is compiled into

```js
var inject = (function() {
  "use strict";
  function defineUnlistedScript(arg) {
    if (arg == null || typeof arg === "function") return { main: arg };
    return arg;
  }
  const definition = defineUnlistedScript(() => {
    if (!globalThis.$0) {
      return;
    }
    return globalThis.$0.childElementCount;
  });
  function initPlugins() {
  }
  const result = (async () => {
    try {
      initPlugins();
      return await definition.main();
    } catch (err) {
      logger.error(
        `The unlisted script "${"inject"}" crashed on startup!`,
        err
      );
      throw err;
    }
  })();
  return result;
})();
```

The key issue is that `await definition.main()` assumes all defined functions return Promises, when in fact they don't. The expected behavior should properly handle both Promise and non-Promise scenarios.

Image

### Reproduction

https://github.com/rxliuli/wxt-error-define-unlisted-script-promise

### Steps to reproduce

git clone https://github.com/rxliuli/wxt-error-define-unlisted-script-promise
pnpm i && pnpm dev
Open any webpage
Open Devtools > Element
Select Element

### System Info

```shell
System:
    OS: macOS 26.0
    CPU: (12) arm64 Apple M2 Max
    Memory: 21.25 GB / 64.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.14.0 - /usr/local/bin/node
    npm: 10.9.2 - /usr/local/bin/npm
    pnpm: 10.10.0 - /usr/local/bin/pnpm
    bun: 1.1.6 - ~/.bun/bin/bun
  Browsers:
    Brave Browser: 131.1.73.105
    Chrome: 140.0.7339.186
    Edge: 128.0.2739.54
    Safari: 26.0
  npmPackages:
    wxt: ^0.20.6 => 0.20.11
```

### Used Package Manager

pnpm

### Validations

rxliuli:
### Related Issue

This PR closes #1906


packages/wxt/src/utils/__tests__/define-unlisted-script.test.ts
```diff
@@ -21,4 +21,22 @@ describe('defineUnlistedScript', () => {
 
     expect(actual).toEqual({ main });
   });
+
+  it('should return the object definition when given a main function that returns a value', () => {
```
aklinker1:
```suggestion
  it('should return the result without awaiting for synchronous main functions', () => {
```

packages/wxt/src/utils/__tests__/define-unlisted-script.test.ts
```diff
@@ -21,4 +21,22 @@ describe('defineUnlistedScript', () => {
...
+    const main = vi.fn(() => 'test');
+
+    const actual = defineUnlistedScript(main);
+
+    expect(actual).toEqual({ main });
+    expect(actual.main()).eq('test');
+  });
+
+  it('should return the object definition when given a main function that returns a promise', async () => {
```
aklinker1:
```suggestion
  it('should return a promise of a result for async main functions', async () => {
```

packages/wxt/src/virtual/unlisted-script-entrypoint.ts
```diff
@@ -2,17 +2,37 @@ import definition from 'virtual:user-unlisted-script-entrypoint';
...
+      `Failed to initialize plugins for "${import.meta.env.ENTRYPOINT}"`,
+      err,
+    );
+    throw err;
+  }
+  let result;
+  try {
+    result = definition.main();
+
+    if (result && typeof result === 'object' && 'then' in result) {
```
aklinker1: I've seen other devs check for promises like this, but never understand why. Why not just check `result instanceof Promise`?

```suggestion
    if (result instanceof Promise) {
```

aklinker1 (changes requested):
I don't like how many try/catches and error rethrowing we're doing, was much cleaner before... but I don't see a way to support promises without doing it this way :/

Looks good for me, couple of small changes and questions before I merge it.

aklinker1 (approved):

PR #1909, opened 2025-09-24
fix: Add another `defineItem` signature when `init` function is passed

toofishes:
### Overview

https://github.com/wxt-dev/wxt/pull/1601 adjusted the typing of `defineItem` to examine whether `defaultValue` or `fallback` were passed, and only then, remove the `null` from the possible returned value. However, it did not accommodate the case where an `init` function is passed to prevent a `null` return, which was previously allowed.

Add a new type signature for this case, and add some tests for type expectations to catch the issue.

### Manual Testing

Added type tests that fail before this change, and succeed after.

### Related Issue

* #1601

PR #1914, opened 2025-10-01
type(storage): when having `init` option, the value type should not contain `null`

aiktb:
### Related Issue


aiktb:
same https://github.com/wxt-dev/wxt/pull/1909

PR #1932, opened 2025-10-11
ci: Support NPM OIDC publishing flow and add provenance

aklinker1:
### Overview

NPM is making changes to how to authenticate publishing packages: https://github.blog/changelog/2025-09-29-strengthening-npm-security-important-changes-to-authentication-and-token-management/

Since WXT used a token, this PR converts it to use OIDC. I have already:

- [x] Revoked the token used to publish in the past.
- [x] Disallowed using tokens to publish each package.
- [x] Added this repo as trusted publisher for all WXT packages.

It also adds provenance automatically as a side effect!

image

### Manual Testing

N/A - This will be tested after merging. Practiced this change on 5 other repos before doing WXT... so it should work.

### Related Issue

N/A

### DIFF
--- a/.github/workflows/update-browser-package.yml
+++ b/.github/workflows/update-browser-package.yml
@@ -12,7 +12,8 @@ jobs:
     name: 'Sync with @types/chrome'
     runs-on: ubuntu-latest
     permissions:
-      contents: write
+      contents: write # Push version changes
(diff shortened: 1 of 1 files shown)


aklinker1:
image

Well, it took like 10 commits... but it's working now.

PR #1938, opened 2025-10-15
fix: content script is incorrectly invalidated when injected multiple times

### LINKED ISSUE #884: Firefox: Sometimes the content script is invalidated almost immediately after execution
### Describe the bug

There is some strange floating bug - sometimes `ContentScriptContext` invalidates itself: content script starts execution, and after a short time `ctx.onInvalidated` is triggered.

Judging by the code, `ContentScriptContext` should work like this:
1. `#stopOldScripts` calls `window.postMessage`, which should invalidate all old scripts by the `wxt:content-script-started` message
2. `#listenForNewerScripts` is called in `setTimeout`, which subscribes to `wxt:content-script-started` (it is subscribed via `setTimeout` in order not to catch its own message sent from `#stopOldScripts`)

Depending on the site and the page, usually the correct order of their triggering is preserved (first the message from `#stopOldScripts` is posted, and only then the context subscribes to `#listenForNewerScripts`), but sometimes the message from `#stopOldScripts` comes later - first the `#listenForNewerScripts` listener is added, and then this listener catches the message sent from `#stopOldScripts` of the listener's own context.

I tried test this on different sites and devices, and:
- it seems that the heavier the JS on a site, the more likely it is to catch the bug
  for example, on https://example.com I've never been able to reproduce it, but on https://openweathermap.org I get it almost every time
- bug is reproducible in different versions of Firefox (I tried FF and LibreWolf 128.0.3, FF 101.0.1, all with clean profiles without extensions), but in Chrome I never got it
- it can be reproduced both with `runAt = document_idle` and with `runAt = document_start`

### Reproduction

Repo: https://github.com/hikiko4ern/wxt-invalidated-cs

There's a commit that adds a patch with additional logs and a unique UUID for clarity, but the bug is reproducible even without it.

### Steps to reproduce

1. `pnpm i`
2. `pnpm dev:firefox`
3. load extension
4. open any page with heavy enough JS

   I've been able to reproduce on pages:
   - https://openweathermap.org (about 90% of the time)
   - https://coub.com/view/zxmi7 (about 50% of the time)

5. open the console and see if there is any log about script invalidation

   if it is invalidated, the logs will show this:
   ```
   [wxt] stopping old scripts 0fbc90a3-df51-4620-9413-8e29a46d6be4
   -- main --
   [wxt] will attach newer scripts listener 0fbc90a3-df51-4620-9413-8e29a46d6be4
   [wxt] invalidated from 0fbc90a3-df51-4620-9413-8e29a46d6be4
   invalidated
   [wxt] Content script "content" context invalidated
   ```
   otherwise it will be like this:
   ```
   [wxt] stopping old scripts c3b0d205-883f-4b21-98c4-5cc0f455015f
   -- main --
   [wxt] will attach newer scripts listener c3b0d205-883f-4b21-98c4-5cc0f455015f
   ```

6. try refreshing the page a few times until the bug is reproduced

### System Info

```shell
System:
  OS: Linux 5.15 Debian GNU/Linux trixie/sid
  CPU: (32) x64 AMD Ryzen 9 7950X3D 16-Core Processor
  Memory: 53.69 GB / 64.76 GB
  Container: Yes
  Shell: 5.9 - /bin/zsh
Binaries:
  Node: 20.15.1 - /mnt/wslg/runtime-dir/fnm_multishells/27245_1722387911992/bin/node
  npm: 10.7.0 - /mnt/wslg/runtime-dir/fnm_multishells/27245_1722387911992/bin/npm
  pnpm: 9.6.0 - /mnt/wslg/runtime-dir/fnm_multishells/27245_1722387911992/bin/pnpm
  bun: 1.1.8 - ~/.bun/bin/bun
Browsers:
  Chrome: 127.0.6533.72
  Chromium: 123.0.6312.105
npmPackages:
  wxt: ^0.19.1 => 0.19.2
```

### Used Package Manager

pnpm

### Validations

### LINKED ISSUE #1937: Content script is incorrectly invalidated when injected multiple times
### Describe the bug

When a content script is injected multiple times, all but the most recent one should be invalidated. In practice however, all content scripts, including the most recent one, are invalidated.

Video going over the issue, the root cause, and the fix: https://www.loom.com/share/97983e5d38904af594d5f486bbf9f1f4?sid=e588c652-c868-4cb0-8db2-521650726919

### Reproduction

Minimal reproduction: https://github.com/namukang/wxt-invalidation-bug

### Steps to reproduce:

1. Clone https://github.com/namukang/wxt-invalidation-bug
2. Run `wxt dev`
3. Navigate to http://example.com
4. Open the console
5. Click the extension icon

### Expected behavior

Only the first content script is invalidated.

### Actual behavior

Both content scripts are invalidated.

### System Info

```shell
System:
    OS: macOS 15.6.1
    CPU: (10) arm64 Apple M1 Max
    Memory: 3.33 GB / 64.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.12.0 - /Users/dk/.nvm/versions/node/v22.12.0/bin/node
    npm: 10.9.0 - /Users/dk/.nvm/versions/node/v22.12.0/bin/npm
    pnpm: 10.6.1 - /Users/dk/.nvm/versions/node/v22.12.0/bin/pnpm
    bun: 1.2.23 - /opt/homebrew/bin/bun
    Deno: 2.5.3 - /opt/homebrew/bin/deno
  Browsers:
    Brave Browser: 100.1.37.109
    Chrome: 141.0.7390.107
    Edge: 141.0.3537.71
    Firefox: 133.0.3
    Safari: 18.6
  npmPackages:
    wxt: ^0.20.11 => 0.20.11
```

### Used Package Manager

pnpm

### Validations

namuorg:
### Overview

Video of issue and fix: https://www.loom.com/share/97983e5d38904af594d5f486bbf9f1f4?sid=e588c652-c868-4cb0-8db2-521650726919

This PR replaces `window.postMessage` with `document.dispatchEvent` in order to avoid race conditions when invalidating content scripts.

Since [`document.dispatchEvent`](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/dispatchEvent) invokes event handlers synchronously, it prevents issues with invalidation due to the asynchronous nature of `window.postMessage`.

I cleaned up related code in `ContentScriptContext` that followed as a result of this change:
- Due to the removal of the race condition, the logic of ignoring the first event used to resolve #884 is no longer necessary.
- The premature invalidation in #1359 should not happen, but to handle the case of a rogue website that somehow replays the custom event, the content script context ignores messages from itself. However, it should not ignore messages from other instances of the same script being injected, so the comparison was changed to simply check the message ID against its own context ID.
- Since `document.dispatchEvent` operates within its frame, `stopOldScripts` is called in all cases, not just in the top frame so that duplicate content scripts are invalidated when they're injected into a frame.

### Related Issue

This PR closes #1937

### DIFF
--- a/packages/wxt/src/utils/content-script-context.ts
+++ b/packages/wxt/src/utils/content-script-context.ts
@@ -247,6 +247,17 @@ export class ContentScriptContext implements AbortController {
         },
       }),
     );
+
+    // Send message using `window.postMessage` for backwards compatibility to invalidate old versions before WXT changed to `document.dispatchEvent`
+    // TODO: Remove this once WXT version using `document.dispatchEvent` has been released for a while
+    window.postMessage(
+      {
+        type: ContentScriptContext.SCRIPT_STARTED_MESSAGE_TYPE,
+        contentScriptName: this.contentScriptName,
+        messageId: this.id,
+      },
+      '*',
+    );
   }
 
   verifyScriptStartedEvent(event: CustomEvent) {


packages/wxt/src/utils/content-script-context.ts
```diff
@@ -243,43 +244,45 @@ export class ContentScriptContext implements AbortController {
   }
 
   stopOldScripts() {
-    // Use postMessage so it get's sent to all the frames of the page.
```
namuorg, line 246: In case there's concern that changing from `window.postMessage` to `document.dispatchEvent` will change behavior when it comes to other frames: This comment is incorrect because `window.postMessage` only sends a message to the targeted window. It doesn't automatically send messages to other frames — you'd need to specifically target those windows in order to send a message.

aklinker1: Doesn't the `*` target mean we're sending messages to all frames?

aklinker1: This is my main concern. Basically, this class is meant to properly stop and shut down a content script when an extension updates. In production or in development, you will encounter bugs if we don't properly stop ALL the content scripts running on a page. So if in one update, a content script changes from `allFrames: true` to `allFrames: false`, some scripts running in iframes won't be stopped by using custom events.

Because I'm 99% sure targeting `"*"` sends a message to all iframes on the page, and any level.

Now, that said, the current approach has tradeoffs as well, this race condition you found. If we don't send a message to all frames, then technically those scripts will still stop after they try and use an extension API, only to find that they can't because the context was invalidated after the update.

On a separate note, another thing to consider is that this is a breaking change that will effect all WXT extensions on their next update. We're going to send a custom event, but the old version of the extension will be listening for messages. This could be resolved by continuing to send the message...

Can I get some more context around why you would inject the same script multiple times in quick succession, like from the reproduction? I need more info to decide how likely this race condition is to  weigh it against merging this PR and changing the behavior.

namuorg: ### `window.postMessage` and iframes
Targeting `"*"` doesn't send a message to all iframes on the page. The `targetOrigin` parameter of `window.postMessage` specifies which origin is *allowed* to receive the message.

In order to send a message to an iframe, the iframe's window needs to be targeted directly: e.g. `iframe.contentWindow.postMessage(message, targetOrigin);`

From [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage):
> Furthermore, an accessing script must have obtained the window object of the accessed document beforehand. This can occur through methods such as [window.open()](https://developer.mozilla.org/en-US/docs/Web/API/Window/open) for popups or [iframe.contentWindow](https://developer.mozilla.org/en-US/docs/Web/API/HTMLIFrameElement/contentWindow) for iframes.

The current code of `window.postMessage(message, "*")` simply sends a message to the current window and allows any origin to receive it, but it doesn't send a message to iframes.

That means the current approach actually has a bug where content scripts inside iframes are not invalidated at all when the content script is injected again (e.g. when the extension updates).

I've created a video demonstration of the behavior with `allFrames: true` with the current implementation and the fix proposed in this PR: https://www.loom.com/share/a172168ee3744979b949ef8411d126bc

### Backwards compatibility

I've added to the PR to temporarily continue sending the message via `window.postMessage` as well in order to invalidate content scripts from before this change. This makes this change backwards compatible and not a breaking change.

For the particular scenario you gave: The new implementation doesn't invalidate content scripts inside iframes when injecting a content script is changed from `allFrames: true` to `allFrames: false`. But this wouldn't be breaking change since the current implementation doesn't handle that case either. I personally think this is an edge case that's not worth worrying about (at least in the scope of this PR), in the same way that WXT doesn't currently try to invalidate content scripts that were injected in previous versions of an extension but were removed or renamed in the current version. I think it'd be fine to let them stop when trying to use an extension API.

Overall, I think this change is worth making because it (1) fixes the bug with content scripts inside iframes not being invalidated and (2) entirely eliminates race conditions with invalidations given the synchronous nature of `document.dispatchEvent`.

### Context for multiple injections

Here's why the content script is injected multiple times in quick succession in my extension ([Easy Scraper](https://easyscraper.com)):

When the user clicks the extension icon to open the popup window, a content script is injected into the current tab and is injected again whenever a new page is loaded in the current tab. The popup window does this by having a listener for `browser.tabs.onUpdated` and checking for `changeInfo.status === "complete"` to detect when a new page is loaded.

However, there are cases when a page causes `browser.tabs.onUpdated` to fire multiple times with `changeInfo.status === "complete"`. For example, it's fired with status `complete` again when the page sets the hash (e.g. `window.location.hash = "section"`) or changes the URL without reloading the page (e.g. `window.history.pushState({}, '', '/new-url');`.

So there are cases where the initial page load causes `browser.tabs.onUpdated` to fire in quick succession, which causes the popup window to inject the content script multiple times. (There's a check that runs to see if the current tab already has the content script before it tries to inject it, but in this case, that check fails because no content script has been injected yet when `onUpdated` is fired multiple times.)

For extensions that dynamically inject the content script using `browser.tabs.onUpdated` (as in my case where only the `activeTab` permission is used along with a popup), this is a fairly unavoidable scenario.

namuorg: @aklinker1 Just seeing if you had a chance to look at this. Thanks!

namuorg:
Oops, I didn't realize that deleting my fork would automatically close the issue.

Reopened as #2035

packages/wxt/src/utils/content-script-context.ts
aklinker1: You are correct about `window.postMessage`'s behavior, I learned something new today! So that means I don't have a concern about switching to custom events, but my breaking change concern is still valid.

The solution there is easy, just keep calling `window.postMessage`.

Let's continue this conversation on the new PR

Issue #1631, opened 2025-04-30
Make Vite a Peer Dependency 


aklinker1:
### Feature Request

I get lots of comments about this issue:

https://github.com/wxt-dev/wxt/issues/1460#issuecomment-2811260819

Which would be solved by making Vite a peer and requiring all projects list a version of it in their `package.json`.

#### Is your feature request related to a bug?

N/A

### What are the alternatives?

Keep it a subdependency. Only benefit here is WXT has control over the version of vite it uses, and can prevent people from trying a newer version, like v7, before it's ready.

It's easy to ignore peer dependency version warnings, while those should tell the user t hey have an invalid version installed.

lishaduck:
One idea that might help, Yarn & pnpm support "peer dependencies with defaults". If you include a package in both `peerDependencies` and `dependencies`, it'll act like a peer, but it'll be auto-installed npm-style unless you install your own version of Vite, in which case it'll inherit your version. In npm, it'll fall back to a normal dependency (status quo).

<details><summary>References</summary>
<p>

Docs:
- [Yarn supports "peer dependencies with default": dependencies listed in both the dependencies and a peerDependencies fields will try to solve the peer dependency first, but will fallback to the regular dependency if it can't be satisfied otherwise.](https://yarnpkg.com/configuration/manifest#peerDependencies:~:text=As%20an%20extension%2C%20Yarn%20supports%20%22peer%20dependencies%20with%20default%22%3A%20dependencies%20listed%20in%20both%20the%20dependencies%20and%20a%20peerDependencies%20fields%20will%20try%20to%20solve%20the%20peer%20dependency%20first%2C%20but%20will%20fallback%20to%20the%20regular%20dependency%20if%20it%20can%27t%20be%20satisfied%20otherwise.)
- [The preferred one is to list the dependency (in Next.js's case, webpack) as both a regular dependency and a peer dependency. Yarn will interpret this pattern as "peer dependency with a default", meaning that your users will be able to take ownership of the Webpack package if they need to, while still giving the package manager the ability to emit a warning if the provided version is incompatible with the one your package expects.](https://yarnpkg.com/advanced/rulebook#:~:text=The%20preferred%20one,your%20package%20expects.)
- [pnpm PR](https://github.com/pnpm/pnpm/pull/5210)

</p>
</details> 

_P.S. I'm back for at least a bit! 😛_

aklinker1:
[...]

PR #1945, opened 2025-10-22
refactor: make vite a default peer

lishaduck:
### Overview

I think this could be breaking in theory, but...
- If you're using the default setup, you're already good.
- If you've customized your Vite version, you probably thought it was a peer.

### Manual Testing

Make some projects and ensure the correct versions are installed.

### Related Issue

This PR closes https://github.com/wxt-dev/wxt/issues/1631

### DIFF
--- a/packages/wxt/package.json
+++ b/packages/wxt/package.json
@@ -58,7 +58,6 @@
     "publish-browser-extension": "^2.3.0 || ^3.0.0",
     "scule": "^1.3.0",
     "unimport": "^3.13.1 || ^4.0.0 || ^5.0.0",
-    "vite": "^5.4.19 || ^6.3.4 || ^7.0.0",
     "vite-node": "^2.1.4 || ^3.1.2",
     "web-ext-run": "^0.2.3"
   },
@@ -80,6 +79,7 @@
     "publint": "^0.3.12",
     "typescript": "^5.8.3",
     "unbuild": "^3.5.0",
+    "vite": "^6.3.5",
     "vitest": "^3.1.2",
     "vitest-plugin-random-seed": "^1.1.1"
   },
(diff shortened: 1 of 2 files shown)


aklinker1:
Could you rebase this off of `major` instead of `main`? This is a breaking change, and that's where I stage all breaking changes before the next major release. It's breaking for projects that don't hoist dependencies, like pnpm.

lishaduck:
> Could you rebase this off of `major` instead of `main`? This is a breaking change, and that's where I stage all breaking changes before the next major release.

Rebased! (fyi, would probably be good for major to get updated from main, it's pretty far back right now)

packages/wxt/package.json
aklinker1: Was about to merge and saw this. We also need to remove it from here, or better, move it to `devDependencies` and specify a single version to use for development.

aklinker1:
I'll rebase `major` after merging this PR so it doesn't effect it.

packages/wxt/package.json
lishaduck: I was making it default peer, see https://github.com/wxt-dev/wxt/issues/1631#issuecomment-3430177592

Happy to make it a normal peer if you'd prefer, but I thought you'd prefer it to be less breaking (and less intrusive)

aklinker1: Oh I see... didn't know that was a thing. I guess that would mean it's not a breaking change, like you originally said.

Hmm... I'd rather stick to a normal peer and make the breaking change. That seems complicated and I don't know how bun handles that, which this project needs to support.

aklinker1 (approved):

aklinker1:
Alright, `major` is rebased with the latest `main`. @lishaduck

PR #1946, opened 2025-10-22
refactor!: move to module-runner

lishaduck:
> [!WARNING]
>
> **Breaking Change**
>
> Drop support for Vite <6

### Overview

Drop `vite-node` in favor of Vite-native `runnerImport`, which is based on `ModuleRunner`.

### Manual Testing

Good question... not entirely sure what fails the treeshake in the first place, but all the tests pass so I assume it loads the entrypoints at least. 🤷🏻‍♂️

### Related Issue

Stacked on #1945.
Supersedes/closes #1884.

### DIFF
--- a/packages/wxt/package.json
+++ b/packages/wxt/package.json
@@ -58,11 +58,10 @@
     "publish-browser-extension": "^2.3.0 || ^3.0.2",
     "scule": "^1.3.0",
     "unimport": "^3.13.1 || ^4.0.0 || ^5.0.0",
-    "vite-node": "^2.1.4 || ^3.1.2",
     "web-ext-run": "^0.2.4"
   },
   "peerDependencies": {
-    "vite": "^5.4.19 || ^6.3.4 || ^7.0.0"
+    "vite": "^6.3.4 || ^7.0.0"
   },
   "devDependencies": {
     "@aklinker1/check": "^2.1.0",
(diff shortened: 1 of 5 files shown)


aklinker1:
If tests pass, that should be good enough.

packages/wxt/package.json
```diff
@@ -58,11 +58,10 @@
     "publish-browser-extension": "^2.3.0 || ^3.0.2",
     "scule": "^1.3.0",
     "unimport": "^3.13.1 || ^4.0.0 || ^5.0.0",
-    "vite-node": "^2.1.4 || ^3.1.2",
     "web-ext-run": "^0.2.4"
   },
   "peerDependencies": {
-    "vite": "^5.4.19 || ^6.3.4 || ^7.0.0"
+    "vite": "^6.3.4 || ^7.0.0"
```
aklinker1, line 64: Perfect, was gonna call this out.

aklinker1 (approved):

lishaduck:
~~This PR is an absolute mess, but it works. Cleanup later if I have time, I need to hit the hay~~

I say that and then I whip out `grbi`. Life is better now, I can sleep in good conscience.
Apparently the issue was that #1945 updated the locked vite version but didn't sync it right, which I tried to fix in here but downgraded instead of upgrading. Fixed that 😅 And guess what? Wouldn't have been (as much of) an issue if Vitepress used a peer... lol

aklinker1:
> Wouldn't have been (as much of) an issue if Vitepress used a peer... lol

Haha yeah, vite should always be a peer is what I've discovered.

PR #1950, opened 2025-10-24
refactor: cleanup esm

lishaduck:
### Overview

A few tiny cleanups I noticed could be made while perusing around the codebase.

### Manual Testing

Bundle, check it still works.

### Related Issue

This PR is stacked on #1945.

### DIFF
--- a/packages/wxt/src/core/runners/__tests__/index.test.ts
+++ b/packages/wxt/src/core/runners/__tests__/index.test.ts
@@ -38,7 +38,7 @@ describe('createExtensionRunner', () => {
   });
 
   it('should return a WSL runner when `is-wsl` is true', async () => {
-    isWslMock.mockReturnValue(true);
+    isWslMock.mockReturnValueOnce(true);
     setFakeWxt({
       config: {
         browser: 'chrome',
@@ -51,7 +51,7 @@ describe('createExtensionRunner', () => {
   });
 
   it('should return a manual runner when `runner.disabled` is true', async () => {
-    isWslMock.mockReturnValue(false);
+    isWslMock.mockReturnValueOnce(false);
     setFakeWxt({
       config: {
         browser: 'chrome',


packages/wxt/src/core/resolve-config.ts
```diff
@@ -536,17 +536,10 @@ async function getUnimportEslintOptions(
  * Returns the path to `node_modules/wxt`.
  */
 function resolveWxtModuleDir() {
-  // TODO: Drop the __filename expression once we're fully running in ESM
-  // (see https://github.com/wxt-dev/wxt/issues/277)
```
lishaduck, line 536: Read this issue a bit closer, is the reason we dynamically import vite to suppress the cjs warning? If so, we can switch to static imports now (there's already static imports in the codebase, so presumably it's just wasting cycles now).

packages/wxt/package.json
```diff
@@ -70,6 +79,7 @@
     "@types/node": "^20.17.6",
     "@types/normalize-path": "^3.0.2",
     "@types/prompts": "^2.4.9",
+    "eslint": "^9.38.0",
```
lishaduck, line 80: Or we could just shim out a `.d.ts`, but that feels overly brittle.

aklinker1:
Ugh, this diff is massive, did you rebase it correctly?

lishaduck:
> Ugh, this diff is massive, did you rebase it correctly?

Haven't rebased it yet

packages/wxt/src/core/builders/vite/index.ts
aklinker1, line 1: I'd like to keep this file as-is, so that vite's not a required dependency. Eventually, I want to add support for more bundlers than just vite, which is why it's a dynamic import here. I'd like to keep the function async so trying to figure out how to make it async again is not a blocker whenever that happens.

packages/wxt/src/core/runners/index.ts
```diff
@@ -3,13 +3,14 @@ import { createWslRunner } from './wsl';
 import { createWebExtRunner } from './web-ext';
 import { createSafariRunner } from './safari';
 import { createManualRunner } from './manual';
-import { isWsl } from '../utils/wsl';
+
```
aklinker1:
```suggestion
```

packages/wxt/src/core/utils/wsl.ts
aklinker1, line 1: We can just delete this file?

packages/wxt/src/core/resolve-config.ts
aklinker1, line 1: Seems like this file is also only about importing `vite` up top. Let's revert it as well now, since Vite will be a peer soon.

packages/wxt/src/core/wxt.ts
```diff
@@ -26,7 +26,7 @@ export async function registerWxt(
 
   const hooks = createHooks<WxtHooks>();
   const config = await resolveConfig(inlineConfig, command);
-  const builder = await createViteBuilder(config, hooks, () => wxt.server);
+  const builder = createViteBuilder(config, hooks, () => wxt.server);
```
aklinker1: Hmm, and here I'm assuming vite exists lol. We should probably add an if statement and dynamic import `createViteBuilder` here as well.

packages/wxt/package.json
```diff
@@ -72,6 +78,7 @@
...
```
aklinker1, line 80: We test this by spinning up NPM projects that have eslint as a dependency, so we don't technically need it installed here. I would prefer not to, just so it's not installed as apart of the monorepo.

lishaduck: The reason we needed it is because otherwise TS errors that there's no types for the `import("eslint")`.

I'll try to shim out a `.d.ts` instead.

packages/wxt/src/core/resolve-config.ts
lishaduck: Even when Vite is a peer, it'll still be installed (either by error [yarn, pnpm sometimes] or by autoinstall [npm, bun, pnpm usually]).

packages/wxt/src/core/utils/wsl.ts
lishaduck: I was having trouble mocking it, but I'll give it another shot

lishaduck (commented):
Oh, apparently I didn't hit submit on my review, sorry for leaving you without guidance.

I was a bit confused, like didn't I just clarify + ask you a followup question? Makes sense.

lishaduck:
> We can also cleanup this file: [main/packages/wxt/src/core/utils/eslint.ts](https://github.com/wxt-dev/wxt/blob/main/packages/wxt/src/core/utils/eslint.ts?rgh-link-date=2025-10-27T23%3A11%3A10Z)

Not sure what you mean, I did, right? `eslint` is an `optional` `peerDependency`, so it still needs the dynamic import.

> Other than the changes that make Vite a required dependency, the rest looks good.

Vite is already a required dependency, is still a required dependency following #1945, and I'm skeptical of the multiple builders argument (I'll follow up on that above).
If you're referencing non-dynamic imports of Vite, those also already exist. I thought it might be intentional, but I decided it was just for CJS-compat because there are a few (two?) files that just directly (non-`type`) import `'vite'`.

I'll revert the changes, but I'm pretty lost as to the reasoning.

> Also, should this be based on the main branch? Or did you intend this to be a breaking change as well?

I didn't want to target changes that might be wrong, but having messed around with the codebase more, I'm pretty confident with them.

packages/wxt/src/core/builders/vite/index.ts
lishaduck: So... I shouldn't work on tying WXT more closely to Vite using the Env API? ... was going to work on that next.

Can you explain why we'd want more bundlers than Vite? I'd understand if it was a webpack project and we were adding in support for RSPack or Turbopack, but there's not really Vite alternatives out there beyond Rolldown-Vite, which'll just be vanilla Vite pretty soon.

packages/wxt/src/core/wxt.ts
lishaduck: But leave that for a different PR?

And I'm still not sure why it matters if Vite exists. It does exist, and acting like it doesn't just feels like premature abstraction, but 🤷🏻‍♂️

packages/wxt/src/core/builders/vite/index.ts
aklinker1: I just don't want to lock WXT into only using vite. If something better comes along in the next 5 years, I want there to be an easy way to add support for it. Nuxt does this, supporting both webpack and vite. While I don't plan to add support for webpack, some new tool like it could pop up and replace Vite.

It's fine to replace `vite-node` with the env API, because I've already organized things so that the entrypoint importers are in the `Builder` interface. So, just as an example, if someone wanted to use webpack and WXT supported it, we wouldn't be using vite to import those files, it would be up to the webpack builder to import them somehow.

packages/wxt/src/core/utils/wsl.ts
aklinker1: I see... That might be why I also created it lol. I just saw we weren't using it anymore, and importing directly from `is-wsl` in `packages/wxt/src/core/runners/index.ts`. I didn't check for other usages.

packages/wxt/src/core/resolve-config.ts
aklinker1: I would prefer to gracefully hanlde the situation where vite is not installed and provide an error message telling the user what is required. If we import it at the top level, we'll just be throwing a module not found error.

packages/wxt/src/core/wxt.ts
aklinker1: Yeah, separate PR

packages/wxt/package.json
aklinker1: Hmm... yeah I guess that's true. Could just ignore the type error too. I'm good with just leaving it then.

I just don't like [how many dependencies it has](https://pkg-size.dev/eslint) that we would otherwise not need to install... But I'm probably being overly picky. Most those dependencies are probably already installed.

packages/wxt/package.json
aklinker1: let's leave it

aklinker1:
The other imports to vite are because I got lazy... I'd prefer to keep it abstracted for now even if it's the only builder WXT supports. Maybe it's about time I clean those other imports up as well...

Undoing the builder abstraction or more tightly coupling Vite to the project is something I will stubbornly refuse to budge on right now, sorry.

packages/wxt/package.json
lishaduck: > Hmm... yeah I guess that's true. Could just ignore the type error too. I'm good with just leaving it then.

👍🏻

> I just don't like [how many dependencies it has](https://pkg-size.dev/eslint) that we would otherwise not need to install... But I'm probably being overly picky. Most those dependencies are probably already installed.

If you look at the diff, it does actually add a significant number of (relatively small, primarily 1st party) new modules, so 🤷🏻‍♂️

packages/wxt/src/core/builders/vite/index.ts
lishaduck: Nuxt still supports webpack‽ 🤯

> It's fine to replace vite-node with the env API, because I've already organized things so that the entrypoint importers are in the Builder interface.

Yeah, but I was going to give a shot at moving the rest of wxt into environments to orchestrate the build.
~~I get out of work, yay!~~ :P

packages/wxt/package.json
lishaduck: I tried to drop it but pnpm wasn't happy with that and I decided it was better to stick on the yellow brick road than fight it on this.

lishaduck:
> The other imports to vite are because I got lazy... I'd prefer to keep it abstracted for now even if it's the only builder WXT supports. Maybe it's about time I clean those other imports up as well...

✅ It looked like more but they were just types. Could we turn on `verbatimModuleSyntax` now?

> Undoing the builder abstraction or more tightly coupling Vite to the project is something I will stubbornly refuse to budge on right now, sorry.

No need to be sorry! I'm just curious why it's built this way, I was surprised because I thought WXT was all-in on Vite. Sorry if it comes off confrontational (and I'll try not to, but knowing me I'll probably manage to complain whenever I touch these files ~~for all of eternity~~ until I decide to add support for some shiny new bundler, please just ignore me)

packages/wxt/src/core/resolve-config.ts
```diff
@@ -584,8 +577,8 @@ export async function mergeBuilderConfig(
   if (vite) {
     return {
       vite: async (env) => {
-        const resolvedInlineConfig = (await inlineConfig.vite?.(env)) ?? {};
-        const resolvedUserConfig = (await userConfig.vite?.(env)) ?? {};
+        const [resolvedInlineConfig = {}, resolvedUserConfig = {}] =
+          await Promise.all([inlineConfig.vite?.(env), userConfig.vite?.(env)]);
```
lishaduck, line 577: This could potentially be a race condition if someone is doing something stupid in their config file. Probably fine.

packages/wxt/src/modules.ts
```diff
@@ -109,7 +109,10 @@ export function addViteConfig(
   wxt.hooks.hook('config:resolved', (wxt) => {
     const userVite = wxt.config.vite;
     wxt.config.vite = async (env) => {
-      const fromUser = await userVite(env);
+      const [vite, fromUser] = await Promise.all([
+        import('vite'),
+        userVite(env),
+      ]);
```
lishaduck, line 115: Ditto

packages/wxt/src/core/wxt.ts
```diff
@@ -26,8 +26,10 @@ export async function registerWxt(
 
   const hooks = createHooks<WxtHooks>();
   const config = await resolveConfig(inlineConfig, command);
-  const builder = await createViteBuilder(config, hooks, () => wxt.server);
-  const pm = await createWxtPackageManager(config.root);
+  const [builder, pm] = await Promise.all([
+    createViteBuilder(config, hooks, () => wxt.server),
+    createWxtPackageManager(config.root),
+  ]);
```
lishaduck, line 32: We own these, should be fine, but could potentially call something user created and cause a new race condition... yeah this is ridiculous.

scripts/upgrade-deps.ts
```diff
@@ -253,7 +253,6 @@ async function detectUpgrades(
     const latestVersion = dep.info['dist-tags'].latest;
     const latestVersionReleasedAt = new Date(dep.info.time[latestVersion]);
 
-    semver.RELEASE_TYPES;
```
lishaduck: oxclint was complaining and I figured I'd be nice to it.
Not sure why check didn't notice it tho, a _bit_ concerning (cause it's in the root prob?).

aklinker1: Yeah, `oxlint` isn't a dependency in the root `package.json`, so `check` doesn't run it... Options to fix this:

1. Just add oxlint once in the root `package.json` and let it run once for all files everywhere
2. Add it to the root and add a `.oxlintignore` file ignoring the `packages` directory
3. Leave it as a future problem lol

I'll leave it as a future problem and track this in an issue: https://github.com/wxt-dev/wxt/issues/1999

aklinker1:
Looks like there are two test failures

PatrykKuniczak:
@aklinker1 I see there's more people, which want to make this code base, a little more readable.
I don't think more strict `eslint` and additional `stylelint` is a bad idea, but you're THE BOSS.

lishaduck:
Thanks for sorting that out for me Aaron!

Kept forgetting I had this sittin' around.

aklinker1:
No worries :+1: Glad we could get it merged.

PR #1957, opened 2025-10-27
chore: Use `linkedom` in a test instead of `happy-dom`

aklinker1:
### Overview

I just realized we had `happy-dom` installed to test a `linkedom` function... So I just removed it and used `linkedom` in the test file.

I also disabled peer installs by default so developers don't have to install as many dependencies to contribute.

### Manual Testing

```sh
cd packages/wxt
pnpm test run devHtmlPrerender.test.ts
```

### Related Issue

N/A

### DIFF
--- a/packages/wxt/package.json
+++ b/packages/wxt/package.json
@@ -71,6 +71,7 @@
     "@types/normalize-path": "^3.0.2",
     "@types/prompts": "^2.4.9",
     "extract-zip": "^2.0.1",
+    "happy-dom": "^18.0.1",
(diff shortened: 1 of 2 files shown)


aklinker1:
We do use happy-dom for some tests... Gonna keep th test file change, but leave the others that use the happy-dom vitest environment as-is.

aklinker1:
Moved the peer auto-install change to a different commit: https://github.com/wxt-dev/wxt/commit/bfa9c1e1c8b81562d8d4a8c2e4f969f987bb8705

PR #1973, opened 2025-11-08
feat: Add log `--level` cli option

### LINKED ISSUE #1942: add log level option to wxt --debug mode
### add log level option to wxt --debug mode

I am using `WXT Modules` to port a website into Chrome extension as running the website in Chrome extension options.

In dev mode, my wxt modules use `addViteConfig` to add vite plugin to support virtual scripts. ex, in `htmlScriptToVirtual` plugin use `config.logger.info` and `config.logger.debug` two log functions. I would like `wxt --debug` command to add  log level support, so that in dev mode only log the `config.logger.info` content. 

```ts
    addViteConfig(wxt, () => ({
      plugins: [
        htmlScriptToVirtual(wxt.config, () => wxt.server),
        vueDevtoolsHack(wxt.config, () => wxt.server),
        wxt.config.command === `build`
          ? htmlScriptToLocal(wxt)
          : undefined,
      ],
    }))
```

"N/A"

comment out the `config.logger.debug` lines, only keep the `config.logger.info` lines.

### Additional context
codes,
```ts
{
      name: `md:dev-html-prerender`,
      apply: `build`,
      transformIndexHtml: {
        order: `post`,
        async handler(html) {
          if (server == null) {
            return html
          }
          const { document } = parseHTML(html)
          // Replace inline script with virtual module served via dev server.
          // Extension CSP blocks inline scripts, so that's why we're pulling them out.
          const promises: Promise<void>[] = []
          const inlineScripts = document.querySelectorAll(`script[src^=http]`)
          inlineScripts.forEach(async (script) => {
            promises.push(new Promise<void>((resolve) => {
              const url = script.getAttribute(`src`) ?? ``
              if (url?.startsWith(`http://localhost`)) {
                resolve()
                return
              }
              doFetch(url).then((textContent) => {
                const key = hash(textContent)
                inlineScriptContents[key] = textContent
                config.logger.info(`key-content`, key, textContent.slice(0, 10))
                script.setAttribute(`src`, `${server.origin}/@id/${virtualInlineScript}?${key}`)
                if (script.hasAttribute(`id`)) {
                  script.setAttribute(`type`, `module`)
                }
                resolve()
              })
            }))
          })
          await Promise.all(promises)
          const newHtml = document.toString()
          config.logger.debug(`\nhtmlScriptToVirtual Old HTML:\n${html}`)
          config.logger.debug(`\nhtmlScriptToVirtual New HTML:\n${newHtml}`)
          return newHtml
        },
      },
    },
```

nickbar01234:
### Overview

Add a global option to set consola log-level. `--debug` mode will take priority if set in conjunction.

### Related Issue

This PR closes #1942

### DIFF
--- a/packages/wxt/src/cli/cli-utils.ts
+++ b/packages/wxt/src/cli/cli-utils.ts
@@ -23,7 +23,7 @@ export function wrapAction(
     if (level && Object.keys(LogLevels).includes(level)) {
       consola.level = LogLevels[level];
     }
-    
+
     // Enable consola's debug mode globally at the start of all commands when
     // the `--debug` flag is passed
     const isDebug = !!args.find((arg) => arg?.debug);


nickbar01234:
Need to also update docs on the new CLI option, but would like your input first @aklinker1

aklinker1 (dismissed):
Nice idea, thanks for the PR!

packages/wxt/src/cli/commands.ts
```diff
@@ -9,6 +9,7 @@ import {
 const cli = cac('wxt');
 
 cli.option('--debug', 'enable debug mode');
+cli.option('--level <level>', 'specify log level');
```
aklinker1: Do we need to document this more? Specifically, which values are available? Either the numbers or the string literals?

nickbar01234: yep. Do you prefer this be part of the cli or in the public docs?

aklinker1: I don't document the CLI flags other than what's printed when running a command with --help, so lets put them in the CLI.

https://wxt.dev/api/cli/wxt-build.html

It will be a bit longer than the rest of the flag descriptions, but that's fine.

> `specify log level (debug|info|...?|error|fatal)`

nickbar01234: Pushed the changes specifying the option

PatrykKuniczak (approved):

PatrykKuniczak:
@aklinker1 Let's take a look on that, i think he done all necessary steps

aklinker1 (approved):

aklinker1:
Yup, looks good, thanks for the ping

PR #1981, opened 2025-11-27
fix(imports): generate d.ts file for eslint gte 9 auto imports

unrevised6419:
### Overview

When having the ESLint config in a `*.ts` file, we have a TS error `Could not find a declaration file for module './.wxt/eslint-auto-imports.mjs'`

| Before | After |
|--------|--------|
| image | image | 

### Manual Testing

1. Have ESLint 9 in your project
2. have ESLint config be in a TS file
3. Use WXT `autoImports` ESLint Plugin

### DIFF
--- a/packages/wxt-demo/wxt.config.ts
+++ b/packages/wxt-demo/wxt.config.ts
@@ -1,5 +1,6 @@
 import { defineConfig } from 'wxt';
 import { presetUno } from 'unocss';
+import path from 'node:path';
 
 export default defineConfig({
   srcDir: 'src',
@@ -54,6 +55,12 @@ export default defineConfig({
         },
       },
       presets: [presetUno()],
+    },
+  },
+  imports: {
+    eslintrc: {
+      enabled: true,
+      filePath: path.resolve('.wxt/eslintrc-auto-import.json'),
     },
   },
 });
--- a/packages/wxt/src/builtin-modules/unimport.ts
+++ b/packages/wxt/src/builtin-modules/unimport.ts
@@ -1,3 +1,4 @@
+import path from 'node:path';
 import { addViteConfig, defineWxtModule } from '../modules';
 import type {
   EslintGlobalsPropValue,
@@ -49,16 +50,16 @@ export default defineWxtModule({
       // Only create global types when user has enabled auto-imports
       entries.push(await getImportsDeclarationEntry(unimport));
 
+      // Only generate ESLint config if that feature is enabled
       if (wxt.config.imports.eslintrc.enabled === false) return;
 
-      // Only generate ESLint config if that feature is enabled
-      entries.push(
-        await getEslintConfigEntry(
-          unimport,
-          wxt.config.imports.eslintrc.enabled,
-          wxt.config.imports,
-        ),
+      const eslintConfigEntries = await getEslintConfigEntry(
+        unimport,
+        wxt.config.imports.eslintrc.enabled,
+        wxt.config.imports,
       );
+
+      entries.push(...eslintConfigEntries);
     });
 
     // Add vite plugin
@@ -105,7 +106,7 @@ async function getEslintConfigEntry(
   unimport: Unimport,
   version: 8 | 9,
   options: WxtResolvedUnimportOptions,
-): Promise<WxtDirFileEntry> {
+): Promise<WxtDirFileEntry[]> {
   const globals = (await unimport.getImports())
     .map((i) => i.as ?? i.name)
     .filter(Boolean)
@@ -115,7 +116,7 @@ async function getEslintConfigEntry(
       return globals;
     }, {});
 
-  if (version <= 8) return getEslint8ConfigEntry(options, globals);
+  if (version <= 8) return [getEslint8ConfigEntry(options, globals)];
   else return getEslint9ConfigEntry(options, globals);
 }
 
@@ -132,8 +133,8 @@ export function getEslint8ConfigEntry(
 export function getEslint9ConfigEntry(
   options: WxtResolvedUnimportOptions,
   globals: Record<string, EslintGlobalsPropValue>,
-): WxtDirFileEntry {
-  return {
+): WxtDirFileEntry[] {
+  const javaScriptFileEntry: WxtDirFileEntry = {
     path: options.eslintrc.filePath,
     text: `const globals = ${JSON.stringify(globals, null, 2)}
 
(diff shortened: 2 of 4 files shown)


packages/wxt/src/types.ts
```diff
@@ -1442,17 +1443,34 @@ export interface Eslintrc {
...
   filePath?: string;
+
+  /**
+   * File path to the TypeScript definition file to use for auto-imported variables.
+   * Needs to match the definition file path for `filePath`
+   *
+   * Default:
+   * - 9 and above: './.wxt/types/wxt-auto-imports.d.mts'
+   */
+  definitionPath?: string;
```
aklinker1: `.d.ts` files are "declaration" files, not "definition" files. 

```suggestion
  declarationFilePath?: string;
```

aklinker1 (changes requested):
Thanks for the PR. I like the addition of a declaration file, but there are lots of extra changes here that are not needed, and some that make this a breaking change:

1. I would prefer to keep the "8" and "9" labeling over "legacy". I expect ESLint will make more breaking changes to their config in future and using version numbers will be better for identifying versions in the long term. I don't see a reason to rename change change the functions related to this.
2. Making changes to the `ResolvedConfig` type is a breaking change. WXT Modules depend on this type, so renaming or changing properties is a breaking change. I'm fine with making breaking changes in general, but they need to be in isolated, small PRs into the [`major` branch](https://github.com/wxt-dev/wxt/tree/major).

    That being said, in this case, I don't think the rename provides much value, not enough to warrant a breaking change in the first place, so I'm recommending we just not make this change.

packages/wxt/src/types.ts
aklinker1: Actually, do we even need this config? It should always be the same as the JS file, right? Just with a different extension.

unrevised6419: > I would prefer to keep the "8" and "9" labeling over "legacy"

Will revert to 8 & 9, but not sure if this will work with upcoming eslint@10

> Making changes to the ResolvedConfig type is a breaking change.

Got it, will revert. I thought is only internal.

unrevised6419: > Actually, do we even need this config?

maybe not, but each type of js file needs its own counterpart of d.ts file

.js -> .d.ts
.mjs -> .d.mts
.cjs -> .d.cts

unrevised6419: Probably we can have an internal mapper, and based on the extension to chose ts,cts,mts

unrevised6419: PS: actually doing some tests now, and it seems that `.d.ts` could cover all JS extensions

aklinker1: Awesome, thanks!

unrevised6419: Question: When user defines `filePath` does it have some restrictions (relative or absolute) regarding the `.wxt` directory? Because I see in the code it always tries to resolve to the `.wxt` directory.

unrevised6419: Because `getUnimportEslintOptions` will resolve it to `.wxt` directory, and later after returning the resovled config from `getUnimportOptions`, the `defu` overrides it again with the user value. It has strange behaviour.

unrevised6419: For example

```js
filePath: '.wxt/eslint-auto-imports.cjs',
```

image

unrevised6419: From my understanding the user config file path needs to be a resolved path, because the default value is a resolved one.

pnpm-lock.yaml
unrevised6419, line 1: Note: Netlify failed because this file was not in sync with package.json

aklinker1: The rest of the checks will fail as well. You'll need to run:

```sh
git fetch
git checkout origin/main pnpm-lock.yaml
pnpm i
```

to reset it.

packages/wxt-demo/wxt.config.ts
```diff
@@ -56,4 +57,10 @@ export default defineConfig({
       presets: [presetUno()],
     },
   },
+  imports: {
+    eslintrc: {
+      enabled: true,
+      filePath: path.resolve('.wxt/eslintrc-auto-import.json'),
+    },
+  },
```
aklinker1, line 65: Instead of changing the demo project, please write tests to test different config. ESLint tests are here: https://github.com/wxt-dev/wxt/blob/04ae542b13329b9580955fa58d0301f05e85aae1/packages/wxt/e2e/tests/auto-imports.test.ts#L219

packages/wxt/src/types.ts
aklinker1: The intention is to allow the user to use either a relative or absolute path in the config. The relative path should be relative to the `.wxt` directory. In the final resolved config passed around internally, the path should be absolute.

So it looks like there's a bug in `resolve-config` module responsible for this. You can see how we resolve paths manually outside of `defu` for all the other major file paths/directories: https://github.com/wxt-dev/wxt/blob/04ae542b13329b9580955fa58d0301f05e85aae1/packages/wxt/src/core/resolve-config.ts#L88

Shouldn't effect this PR, I'd prefer we fix it in a separate PR. I can do it unless you want to.

packages/wxt/src/builtin-modules/unimport.ts
```diff
@@ -147,4 +148,26 @@ export default {
...
+  const javaScriptFileExtension = path.extname(options.eslintrc.filePath);
+  const javaScriptFileBasename = path.basename(
+    options.eslintrc.filePath,
+    javaScriptFileExtension,
+  );
+
+  const typeScriptFilePath = path.join(
+    javaScriptFileDirname,
+    `${javaScriptFileBasename}.d.ts`,
+  );
```
aklinker1, line 162: This could be simplified a little bit

```suggestion
  const typeScriptFilePath = options.eslintrc.filePath.slice(0, -extname(options.eslintrc.filePath)) + ".d.ts"
```

The second slice argument might be off by one, I can never remember if `extname` returns the extension with or without the dot. Or just remove the first dot from `".d.ts"`.

aklinker1 (changes requested):
Nice, I like the new changes. Let's just add tests then the PR is good to go.

pnpm-lock.yaml
aklinker1: Oh shoot sorry, `main` is broken!

PatrykKuniczak:
@unrevised6419 Are you want to go on?

unrevised6419:
Hey @PatrykKuniczak, no, thanks for asking! 
This was a low priority for me because I encountered it during an interview, and decided to fix it, but got blocked at adding the tests for this.

If anyone wants to take over this, feel free.

PR #1992, opened 2025-12-11
ci: Fix update-browser-package action

alexkhrisanfov-st:
### Overview

This PR includes changes that fix `update-browser-package` action in the protected branch.


.github/workflows/update-browser-package.yml
```diff
@@ -18,6 +18,9 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@ff7abcd0c3c05ccf6adc123a8cd1fd4fb30fb493 # v5.0.0
+        with:
+          fetch-depth: 0
```
aklinker1: Don't need `fetch-depth`, that's only for jobs that need the full git history, like for generating changelogs.

```suggestion
```

aklinker1 (changes requested):

PR #1993, opened 2025-12-12
feat: Default to using `use_dynamic_url: true` for content script css files

aeharding:
### Overview

The property [`use_dynamic_url`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/web_accessible_resources) can be set to prevent webpages from fingerprinting users by checking for the existence of extension CSS files. 

[I made a PR](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/73809) to upstream @chrome/types so this is now available!

### Manual Testing

Try building an extension with CSS and injection mode UI.

### Related Issue

None


davestewart:
That is a useful PR! LinkedIn checks 100s, if not 1000s of extension ids, and potentially will ban users for using specific extensions 😮

PR #1996, opened 2025-12-15
fix(storage): Set version number on init

42willow:
### Overview

Sets version number in meta if storage is initiated with a version number greater than 1.

When the storage is first accessed/initiated, the v2 migration runs even though the storage item was first set as v2. This seems to occur because the version meta key remains unset after the first migration (when storage is null) / when initiating the storage item in storage for the first time.

### Manual Testing

Tested in https://github.com/schooltape/schooltape/pull/324

### DIFF
--- a/packages/storage/src/index.ts
+++ b/packages/storage/src/index.ts
@@ -711,7 +711,7 @@ export interface WxtStorage {
   /**
    * Get the metadata of multiple storage items.
    *
-   * @param items List of keys or items to get the metadata of.
+   * @param keys List of keys or items to get the metadata of.
    * @returns An array containing storage keys and their metadata.
    */
   getMetas(
@@ -758,7 +758,7 @@ export interface WxtStorage {
   /**
    * Set the metadata of multiple storage items.
    *
-   * @param items List of storage keys or items and metadata to set for each.
+   * @param metas List of storage keys or items and metadata to set for each.
    */
   setMetas(
     metas: Array<
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "engines": {
     "node": ">=18.20.3"
   },
-  "packageManager": "pnpm@10.15.1",
+  "packageManager": "pnpm@10.29.2+sha512.bef43fa759d91fd2da4b319a5a0d13ef7a45bb985a3d7342058470f9d2051a3ba8674e629672654686ef9443ad13a82da2beb9eeb3e0221c87b8154fff9d74b8",
   "scripts": {
     "check": "check && pnpm -r --sequential run check",
     "test": "pnpm -r --sequential run test run",
--- a/packages/wxt-demo/wxt.config.ts
+++ b/packages/wxt-demo/wxt.config.ts
@@ -1,5 +1,5 @@
 import { defineConfig } from 'wxt';
-import { presetUno } from 'unocss';
+import { presetWind3 } from 'unocss';
 
 export default defineConfig({
   srcDir: 'src',
@@ -53,7 +53,7 @@ export default defineConfig({
           ],
         },
       },
-      presets: [presetUno()],
+      presets: [presetWind3()],
     },
   },
 });
--- a/packages/wxt/src/cli/__tests__/index.test.ts
+++ b/packages/wxt/src/cli/__tests__/index.test.ts
@@ -1,10 +1,12 @@
 import { describe, it, vi, beforeEach, expect } from 'vitest';
-import { build } from '../../core/build';
-import { createServer } from '../../core/create-server';
-import { zip } from '../../core/zip';
-import { prepare } from '../../core/prepare';
-import { clean } from '../../core/clean';
-import { initialize } from '../../core/initialize';
+import {
+  build,
+  createServer,
+  zip,
+  prepare,
+  clean,
+  initialize,
+} from '../../core';
 import { mock } from 'vitest-mock-extended';
 import consola from 'consola';
 


42willow:
this is only a problem when storage items are being initiated/accessed for the first time with a version greater than 1 - so should only set meta in those cases

aklinker1:
Should the fix actually be to set the version number during the `init` logic, not during the migration logic?

I'm having a hard time understanding why this fix is necessary in the first place, could you point to the exact diff that lead to you encountering a bug?

42willow:
> Should the fix actually be to set the version number during the `init` logic, not during the migration logic?

yes, I will make that change
edit: was I correct in the assumption that meta should only be set if the version is greater than 1 (because of the fallback)?

> I'm having a hard time understanding why this fix is necessary in the first place, could you point to the exact diff that lead to you encountering a bug?

```ts
import { nanoid } from 'nanoid'; 

// Retroactively add a type for the first version
type IgnoredWebsiteV1 = string; 
interface IgnoredWebsiteV2 { 
  id: string; 
  website: string; 
} 

export const ignoredWebsites = storage.defineItem<string[]>( 
export const ignoredWebsites = storage.defineItem<IgnoredWebsiteV2[]>( 
  'local:ignoredWebsites',
  {
    fallback: [],
    version: 2, 
    migrations: { 
      // Ran when migrating from v1 to v2
      2: (websites: IgnoredWebsiteV1[]): IgnoredWebsiteV2[] => { 
        return websites.map((website) => ({ id: nanoid(), website })); 
      }, 
    }, 
  },
);
```

taking this example from the docs, if the storage item has not been initiated yet (i.e. neither meta or value are not set in browser storage):
1. on first set value migration `v2` will skip as value is null - meta won't be set as the logic returns before that point
    - https://github.com/wxt-dev/wxt/blob/main/packages/storage/src/index.ts#L371
    - https://github.com/wxt-dev/wxt/blob/main/packages/storage/src/index.ts#L411
    - value is now set and meta is unset
2. on second access, migration `v2` will run as meta is unset, and it falls back to setting `currentVersion` to `1` in this case
    - https://github.com/wxt-dev/wxt/blob/main/packages/storage/src/index.ts#L373

so the websites parameter of the `v2` migration could be of type `IgnoredWebsiteV1[]` or `IgnoredWebsiteV2[]`

this isn't an issue if the storage item has already been initiated with the v1 schema

packages/storage/src/__tests__/index.test.ts
```diff
@@ -706,6 +706,35 @@ describe('Storage Utils', () => {
...
+      it('should not run migrations if initialized on a version greater than 1', async () => {
+        const migrateToV2 = vi.fn((oldValue) => oldValue * 2);
+
+        const item = storage.defineItem<number, { v: number }>(`local:count`, {
+          defaultValue: 0,
+          version: 2,
+          migrations: {
+            2: migrateToV2,
+          },
+        });
```
aklinker1: This test does not use `init` when defining the item... doesn't seem like it's testing the PR's change? Or am I missing something and still don't understand what's going on here?

42willow: I tested this test and it fails on `main`.

I made the change inside `getOrInitValue`, which runs after the migration is finished. Migrate is called by setValue (i think?) https://github.com/wxt-dev/wxt/blob/43f6c062718f7e82eeb51e74f5ad80b7d483eb3e/packages/storage/src/index.ts#L471 but for some reason this doesn't happen in the test environment, so I added the migration call manually.

```ts
await item.migrate();
await waitForMigrations();
```

https://github.com/wxt-dev/wxt/blob/43f6c062718f7e82eeb51e74f5ad80b7d483eb3e/packages/storage/src/index.ts#L447-L448

When I first found this bug I wasn't actually using the `storageItem.init()` method but was using storage item fallbacks and `setValue`.

I may be misunderstanding something though as I don't *really* know what I'm doing :)

aklinker1:
@42willow I think I created more relevant test cases:

```ts
    describe("init option", () => {
      // ...
      
      it('should set the version without running migrations for empty storage items', async () => {
        const migrate = vi.fn((n: number) => n * 2);

        const item = storage.defineItem<number>('local:key', {
          init: () => 1,
          version: 2,
          migrations: {
            2: migrate,
          },
        });
        const value = await item.getValue();
        const meta = await item.getMeta();

        expect(value).toEqual(1);
        expect(meta).toEqual({ v: 2 });
        expect(migrate).toBeCalledTimes(0);
      });
      
      it('should set the version and run migrations if a storage value already exists and v2 is added', async () => {
        const key = 'local:key';

        // Mimic as if an initial value was set by a previous version. We expect
        // this value to be migrated and doubled to V2.
        await storage.setItem(key, 1);

        const migrate = vi.fn((n: number) => n * 2);
        const item = storage.defineItem<number>(key, {
          init: () => 1,
          version: 2,
          migrations: {
            2: migrate,
          },
        });
        const value = await item.getValue();
        const meta = await item.getMeta();

        expect(value).toEqual(2);
        expect(meta).toEqual({ v: 2 });
        expect(migrate).toBeCalledTimes(1);
      });
    });
```

For the first one, we verify that the migration is NOT ran and that the version is set to 2 after the value is initialized.

For the second one, we test what happens in the same case but where a previous version of the extension had already initialized the value to `1`. In that case, we do expect the migration to run and a value of `2`.

Does that match the setup of the bug you're seeing?

42willow:
@aklinker1 these are perfect, thanks! the first one fails on `main` and the second one passes on both.

I have updated the test cases accordingly :)

packages/storage/src/__tests__/index.test.ts
```diff
@@ -748,7 +789,7 @@ describe('Storage Utils', () => {
         const actualMeta = await item.getMeta();
 
         expect(actualValue).toEqual(0);
-        expect(actualMeta).toEqual({});
+        expect(actualMeta).toEqual({ v: 3 });
```
aklinker1: I don't think this test change is correct - if the default value is set, nothing is written to storage, and there is no value to "version" yet.


```suggestion
        expect(actualMeta).toEqual({});
```

aklinker1: I don't think this test change is correct - if the default value is set, nothing is written to storage, and there is no value to "version" yet.

```suggestion
        expect(actualMeta).toEqual({});
```

aklinker1: So I think this means the fix is in the wrong place.

packages/storage/src/index.ts
aklinker1, line 493: Move the new if statement down to here, after `driver.setItem`. I think that should fix the existing test that shouldn't be changed.

42willow: thanks! I have made the changes

42willow:
just tested this with the problem i had in https://github.com/schooltape/schooltape/pull/324 and it still is running the v2 migration even though it is being initialised as v2. I ran the low quality initial test and it seems to be failing after https://github.com/wxt-dev/wxt/pull/1996/commits/c4577d72263a4b9e61605656be81d25df5a2a734 :(

aklinker1:
Did I release this yet?

aklinker1:
Running now: https://github.com/wxt-dev/wxt/actions/runs/21969045007

aklinker1:
Looks like that package was quite out of date. Try it again with https://github.com/wxt-dev/wxt/releases/tag/storage-v1.2.7

42willow:
@aklinker1 ah thanks! will check later. I got confused as the main package was updated 👍

aklinker1:
Yup, it's a sub dependency, so make sure you actually have the latest version installed

42willow:
I'm trying to pin `@wxt-dev/storage` to `1.2.7` but it keeps installing `1.1.1` as a sub dependency of `@wxt`, do you know why this might be the case?

edit: I think this might be an issue with my project- `wxt prepare` fails after i update my dependencies, https://discord.com/channels/1212416027611365476/1471755129421168650

42willow:
I have updated `@wxt-dev/storage` to `1.2.7` and am still getting the issue.
https://github.com/schooltape/schooltape/commit/c52ef2416c7bb1d18890c6bf3c7afe6f63509d1a

42willow:
I'll see if I can write a more appropriate test for this problem to see if reverting https://github.com/wxt-dev/wxt/commit/c4577d72263a4b9e61605656be81d25df5a2a734 fixes the problem and submit a new PR.

PR #2000, opened 2025-12-16
chore: Upgrade dev and non-major prod dependencies

aklinker1:


packages/wxt/e2e/tests/remote-code.test.ts
```diff
@@ -16,7 +16,7 @@ describe('Remote Code', () => {
     const output = await project.serializeFile('.output/chrome-mv3/popup.js');
     expect(output).toContain(
       // Some text that will hopefully be in future versions of this script
-      'lodash.com',
+      '__lodash_placeholder__',
```
aklinker1, line 19: "lodash.com" was in a comment, and it appears a new version of... vite? esbuild? is causing the comment to not be included.

PR #1934, opened 2025-10-13
fix: Optimize `splitShadowRootCss`

ZerGo0:
### Overview

I did some benchmarking in my prod. ext. and noticed that ~8% (~65ms) of the initial page render time is spent in `splitShadowRootCss`. It is pretty unoptimized right now and this PR changes that.

image

Main problem with the original one seems to be GC and regex related:

Original:
image

Optimized:
image

### Manual Testing

Benchmark script (copy and paste this in the Chrome dev tools, ai generated):
```
function splitShadowRootCss(css) {
  let shadowCss = css;
  let documentCss = "";

  const rulesRegex = /(\s*@(property|font-face)[\s\S]*?{[\s\S]*?})/gm;
  let match;
  while ((match = rulesRegex.exec(css)) !== null) {
    documentCss += match[1];
    shadowCss = shadowCss.replace(match[1], "");
  }

  return {
    documentCss: documentCss.trim(),
    shadowCss: shadowCss.trim(),
  };
}

const AT_RULE_BLOCKS = /(\s*@(property|font-face)[\s\S]*?{[\s\S]*?})/gm;

function splitShadowRootCssOptimized(css) {
  const documentCss = Array.from(css.matchAll(AT_RULE_BLOCKS), (m) => m[0])
    .join("")
    .trim();
  const shadowCss = css.replace(AT_RULE_BLOCKS, "").trim();

  return {
    documentCss,
    shadowCss,
  };
}

function generateTestCss(size = 5000) {
  const rule =
    `  @property --x { syntax: "<color>"; inherits: true; initial-value: red; }\n` +
    `\t@font-face { font-family: "Test"; src: url("test.woff2"); }\n` +
    `.foo { color: blue; background: var(--x); }\n` +
    `\n/* comment */\n  .bar{font-family:"Test"}\n\n` +
    `@supports (font-variation-settings: "wght" 400) { .baz { letter-spacing: 0; } }\n`;
  return rule.repeat(size);
}

function assertEqual(a, b, label) {
  const same = a.documentCss === b.documentCss && a.shadowCss === b.shadowCss;
  if (!same) {
    console.error(`❌ Validation failed: ${label}`);
    process.exit(1);
  }
}

function benchmark(fn, css, label, iterations = 1000) {
  for (let i = 0; i < 5; i++) fn(css);

  const start = performance.now();
  for (let i = 0; i < iterations; i++) fn(css);
  const end = performance.now();

  const total = end - start;
  const avg = total / iterations;
  console.log(
[...]

PR #2005, opened 2025-12-17
feat: add cssPropertyRename option to prevent CSS @property conflicts

### LINKED ISSUE #1955: CSS @property conflicts when extension uses Tailwind v4 and host page uses Tailwind v3
### Describe the bug

When using Shadow Root UI with Tailwind CSS v4, the extension's `@property` rules conflict with host pages that use Tailwind CSS v3, breaking their gradient styles and other CSS properties.

**Root Cause:**
WXT's `splitShadowRootCss` function (introduced in PR #1594) extracts `@property` rules from shadow DOM CSS and injects them into the host page's `<head>`. This is necessary because `@property` doesn't work inside shadow DOM.

However, when the extension uses Tailwind v4 (which defines typed `@property` rules like `syntax: "<color>"`) and the host page uses Tailwind v3 (which uses composite CSS variable values), conflicts occur:

- **Extension (Tailwind v4)**:
  ```css
  @property --tw-gradient-from {
    syntax: "<color>";
    inherits: false;
    initial-value: #0000;
  }
  ```

- **Host Page (Tailwind v3)**:
  ```css
  --tw-gradient-from: rgba(251, 113, 133, 0.8) var(--tw-gradient-from-position);
  ```

The browser enforces the `<color>` type constraint globally, rejecting the host page's composite value that includes `var(--tw-gradient-from-position)`, causing the host page's gradients to break.

**Real-world Impact:**
Sites like https://yesicon.app/ph (and any site using Tailwind v3) have their gradient buttons and other styles broken when our extension is active. 

For more sites has the problem, please refer this issue: https://github.com/mengxi-ream/read-frog/issues/643

## Suggested Solution

WXT could provide a built-in configuration option to rename custom properties before injecting them into the host page. For example:

```typescript
createShadowRootUi(ctx, {
  name: 'my-ui',
  // New option to prevent conflicts
  cssPropertyPrefix: 'wxt-', // or extension-specific prefix
  // ...
});
```

This would rename `--tw-*` properties to `--wxt-tw-*` (or custom prefix) at the WXT framework level, preventing naming conflicts with host pages.

## Current Workaround

We've implemented a PostCSS plugin to rename properties at build time:
- Issue: https://github.com/mengxi-ream/read-frog/issues/643
- Solution PR: https://github.com/mengxi-ream/read-frog/pull/655

While this works, it would be better if WXT handled this internally to:
1. Prevent other extension developers from hitting the same issue
2. Make shadow root UI safer by default
3. Reduce boilerplate configuration for each extension

### Reproduction

1. Create a WXT extension with Shadow Root UI using Tailwind CSS v4
2. Visit a website that uses Tailwind CSS v3 with gradients (e.g., https://yesicon.app/ph)
3. Observe the host page's gradient styles are broken with console error: "Invalid property value, expected type '<color>'"

**Expected behavior:**
Extension's CSS should not interfere with host page styles.

### Steps to reproduce

any WXT extension with shadow root ui can trigger the bug

### System Info

```shell
System:
    OS: macOS 26.0.1
    CPU: (14) arm64 Apple M4 Pro
    Memory: 315.22 MB / 24.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 24.10.0 - /Users/leonz/Github/ghost/read-frog-ghost/node_modules/.bin/node
    npm: 10.9.3 - /Users/leonz/Library/pnpm/npm
    pnpm: 10.16.1 - /opt/homebrew/bin/pnpm
    Deno: 2.3.1 - /Users/leonz/.deno/bin/deno
  Browsers:
    Brave Browser: 140.1.82.166
    Chrome: 141.0.7390.123
    Firefox: 142.0
    Safari: 26.0.1
  npmPackages:
    wxt: 0.20.10 => 0.20.10 

- WXT version: (any version with PR #1594 merged)
- Framework: Tailwind CSS v4
- Conflict occurs with: Any host page using Tailwind CSS v3
```

### Used Package Manager

pnpm

### Validations

taiiiyang:
### Overview

  Add `cssPropertyRename` option to `createShadowRootUi` that allows renaming CSS custom property prefixes to prevent
  conflicts with host page styles.

  **Problem:**
  When extensions use Tailwind CSS v4 with Shadow Root UI, the `@property` rules (e.g., `@property --tw-gradient-from {
  syntax: "<color>"; }`) are extracted and injected into the host page's `<head>`. This causes conflicts with host pages
  using Tailwind CSS v3, which use composite CSS variable values that don't match the typed syntax constraint.

  **Solution:**
  - Added `renameCssCustomProperties()` utility function that renames:
    - `@property` rules: `@property --tw-xxx` → `@property --wxt-tw-xxx`
    - Property declarations: `--tw-xxx: value` → `--wxt-tw-xxx: value`
    - `var()` references: `var(--tw-xxx)` → `var(--wxt-tw-xxx)`
  - Added `cssPropertyRename` option to `ShadowRootContentScriptUiOptions`
  - Both `fromPrefix` and `toPrefix` are optional:
    - If only `toPrefix` is provided: prepends prefix to all custom properties
    - If only `fromPrefix` is provided: removes the matching prefix
    - If both are provided: replaces `fromPrefix` with `toPrefix`

  ### Manual Testing

  1. Create a WXT extension with Shadow Root UI using Tailwind CSS v4:
  ```ts
  createShadowRootUi(ctx, {
    name: 'my-ui',
    cssPropertyRename: {
      fromPrefix: '--tw-',
      toPrefix: '--my-ext-tw-',
    },
    onMount(container) {
      // ...
    },
  });
```
  2. Build the extension and load it in the browser
  3. Visit a website using Tailwind CSS v3 with gradients (e.g., https://yesicon.app/ph)
  4. Verify that the host page's gradient styles are NOT broken
  5. Verify that the extension's styles work correctly

  ### Related Issue

  This PR closes #1955

### CODE REVIEW

File: packages/wxt/src/utils/__tests__/rename-css-custom-properties.test.ts [hunk 1]
```diff
@@ -0,0 +1,251 @@
...
+      const css = `.class { --prop: value; background: var(--prop); }`;
+      const expected = `.class { --wxt-prop: value; background: var(--wxt-prop); }`;
+      expect(renameCssCustomProperties(css, { toPrefix: '--wxt-' })).toBe(
+        expected,
+      );
+    });
+
+    it('should remove fromPrefix when toPrefix is undefined', () => {
+      const css = `.class { --tw-prop: value; }`;
+      const expected = `.class { prop: value; }`;
```
aklinker1: This is... probably fine? Trying to decide if throwing an error would be better than producing this CSS. Probably not? This isn't valid CSS but having weird styles is better than not showing your UI at all? Hmm, not sure. Did you consider throwing an error in this case?

packages/wxt/src/utils/content-script-ui/shadow-root.ts
```diff
@@ -207,4 +217,25 @@ export type ShadowRootContentScriptUiOptions<TMounted> =
       shadow: ShadowRoot,
       shadowHost: HTMLElement,
     ) => TMounted;
+    /**
+     * Rename CSS custom property prefixes to prevent conflicts with host page styles.
```
aklinker1:
```suggestion
     * Rename CSS custom property and var prefixes to prevent conflicts with host page styles.
```

aklinker1 (changes requested):
Code looks good, thanks for writing tests. 

We had really bad performance with the `splitShadowRootCss` function that [some else found and improved](https://github.com/wxt-dev/wxt/pull/1934). Since that function and this new one both are apart of the same function, `createShadowRootUi`, and slow performance here is known to cause a bad UX, can you benchmark the `createShadowRootUi` function using real CSS from tailwind? 

Compare how many times per second `createShadowRootUi` can run with or without `cssPropertyRename`. If it's slow, try and optimize it. If it's fast and the hit is negligable, no need to make any improvements.

You can use Vitest's [`bench` utils](https://vitest.dev/api/#bench) to write the benchmark, please include it in the PR, don't just delete it after writing it.

packages/wxt/src/utils/__tests__/rename-css-custom-properties.test.ts
taiiiyang: What do you think would be a better way to handle the situation when toPrefix or fromPrefix is empty? Maybe we could set both toPrefix and fromPrefix as required fields? If either of them is not passed in, no processing would be done?

aklinker1: So I think there are three options:

1. Throw an error and make devs look in the console to figure out why their content script UI isn't showing up
2. Use what you have now and devs will see the UI, but things like borders and rings that depend on custom properties will be styled wrong and they have to debug why it's not showing up correctly.
3. Don't change the CSS at all if there are invalid options and devs will never know their styles mess up tailwind V3 sites unless they are developing on one directly.

Now that I've written all that out... I think option 1 where we throw an error would be the best DX. It's clear something is broken because their UI wouldn't show up and we can clearly state the problem and solution in the error message without the dev having to go through the CSS and debug which properties are the cause of the bad styles.

taiiiyang: Okay, I also think option 1 is excellent.

taiiiyang: I added non-empty validation for toPrefix and fromPrefix. Both of them need to be required fields.

taiiiyang:
> Code looks good, thanks for writing tests.
> 
> We had really bad performance with the `splitShadowRootCss` function that [some else found and improved](https://github.com/wxt-dev/wxt/pull/1934). Since that function and this new one both are apart of the same function, `createShadowRootUi`, and slow performance here is known to cause a bad UX, can you benchmark the `createShadowRootUi` function using real CSS from tailwind?
> 
> Compare how many times per second `createShadowRootUi` can run with or without `cssPropertyRename`. If it's slow, try and optimize it. If it's fast and the hit is negligable, no need to make any improvements.
> 
> You can use Vitest's [`bench` utils](https://vitest.dev/api/#bench) to write the benchmark, please include it in the PR, don't just delete it after writing it.

I have made adjustments to the current `renameCssCustomProperties` method, and now there are no performance issues. Here are the test results obtained using vitest bench.

PatrykKuniczak:
@aklinker1 This one, still wait for your review, i think

PR #2006, opened 2025-12-18
chore: Use `tsdown` to build packages

aklinker1:
### Overview

I've started using it in other projects and it's way faster and nicer to work with compared to `unbuild`.

Here's some comparisons between the two:

| Comparison | `unbuild` | `tsdown` | Diff |
| --- | --- | --- | --- |
| `pnpm -r build` timing (on my PC) | 21.637 s | 17.103 s | 21.0% faster |
| Sum of all `pnpm pack`ed TGZ files | 354,603 B | 296,413 B | 16.4% smaller |

Other differences:

- Unbuild's `mkdist` mode vs TSDown's `unbundle` mode
   - `tsdown` only includes files that were actually used by the specified entrypoints, meaning tests and other non-imported packages (like WXT's virtual entrypoints) are not transpiled
   - `tsdown` doesn't generate declaration files for files not imported by the entrypoint's types, meaning dirs like the `cli` go from 6 files (3 js, 3 declaration) to 4 files (3 js, 1 declaration)
- `tsdown` strips triple-slash references - not good in our case, but workable.
- `unbuild` created both `.d.ts` and `.d.mts` files, which were exact copies, `tsdown only creates `.d.mjs` files.

Want to be careful to not make any breaking changes here:

- [x] Continue building CJS if the package is doing that (these should be removed in 1.0)
- [x] Don't standardize the use of `main`/`module` top-level `package.json` fields (this should be fixed in 1.0)
- [x] WXT's build is very sensitive - lots of things going on. Already fixed one issue (related to triple-slash comments), wouldn't be surprised if there were more issues. Try via pkg.pr.new before merging

### Manual Testing

Run checks. This will build all packages and verify the outputs with publint.

```sh
pnpm check
```

### Related Issue

N/A

### DIFF
--- a/packages/wxt/e2e/tests/output-structure.test.ts
+++ b/packages/wxt/e2e/tests/output-structure.test.ts
@@ -343,44 +343,40 @@ describe('Output Directory Structure', () => {
 
     expect(await project.serializeFile('.output/chrome-mv3/background.js'))
       .toMatchInlineSnapshot(`
-        ".output/chrome-mv3/background.js
-        ----------------------------------------
-        import { l as logHello, i as initPlugins } from "./chunks/_virtual_wxt-plugins-OjKtWpmY.js";
-        function defineBackground(arg) {
-          if (arg == null || typeof arg === "function") return { main: arg };
-          return arg;
-        }
-        const definition = defineBackground({
-          type: "module",
-          main() {
-            logHello("background");
+          ".output/chrome-mv3/background.js
+          ----------------------------------------
+          import { l as logHello, i as initPlugins } from "./chunks/_virtual_wxt-plugins-OjKtWpmY.js";
+          function defineBackground(arg) {
+            if (arg == null || typeof arg === "function") return { main: arg };
+            return arg;
           }
-        });
-        globalThis.browser?.runtime?.id ? globalThis.browser : globalThis.chrome;
-        function print(method, ...args) {
-          return;
-        }
-        const logger = {
-          debug: (...args) => print(console.debug, ...args),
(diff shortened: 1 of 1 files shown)


packages/analytics/package.json
```diff
@@ -34,16 +34,16 @@
       "default": "./dist/providers/umami.mjs"
     }
   },
-  "module": "./dist/index.mjs",
-  "types": "./dist/index.d.ts",
+  "main": "./dist/index.mjs",
```
aklinker1: Revert

packages/i18n/package.json
```diff
@@ -45,44 +45,24 @@
     "oxlint": "^1.33.0",
     "publint": "^0.3.16",
     "typescript": "^5.9.3",
-    "unbuild": "^3.6.1",
     "vitest": "^4.0.16",
     "vitest-plugin-random-seed": "^1.1.2",
     "wxt": "workspace:*"
   },
-  "main": "./dist/index.cjs",
-  "module": "./dist/index.mjs",
```
aklinker1, line 54: Add back CJS

.gitignore
```diff
@@ -23,3 +23,6 @@ stats.html
 .tool-versions
 .cache
 *-stats.txt
+
+# TODO: remove
+dist-old
```
aklinker1: TODO

TODO
```diff
@@ -0,0 +1 @@
+TODO: Remove dev deps used in all packages and move them to the root package.json
```
aklinker1: TODO - future PR

PR #2010, opened 2025-12-19
feat: add default_state option to popup actions

kairi003:
## Overview

This PR adds support for the `default_state` option in popup actions (browser_action/page_action for MV2, action for MV3). This option controls the initial enabled/disabled state of the extension's action button, as documented in the [Chrome Extension Action API](https://developer.chrome.com/docs/extensions/reference/api/action#enabled_state).

Previously, WXT did not support configuring this property in the manifest, making it impossible to set the action button to a disabled state before any runtime methods are called. This PR allows setting the initial state declaratively in the entrypoint options, following the same pattern as the existing options.

## Changes

- Added `defaultState` property to `PopupEntrypointOptions` type
- Updated manifest generation logic to include `default_state` when specified
- Updated browser API type definitions to support the property
- Added comprehensive tests for MV2 and MV3 manifest generation

## Manual Testing

Verified in `wxt-demo`:
- Added `default_state: "disabled"` to `wxt.config.ts` manifest configuration
- Confirmed TypeScript type checking passes without errors
- Built the extension and verified `default_state` property is correctly included in the generated `manifest.json`

## Related Issue

No related issue - this is a standalone enhancement to improve manifest configuration options.

### DIFF
--- a/packages/browser/src/gen/index.d.ts
+++ b/packages/browser/src/gen/index.d.ts
@@ -9132,6 +9132,7 @@ export namespace Browser {
             default_icon?: ManifestIcons | undefined;
             default_title?: string | undefined;
             default_popup?: string | undefined;
+            default_state?: "enabled" | "disabled" | undefined;
         }
 
         /** Source: https://developer.chrome.com/docs/extensions/reference/permissions-list */
--- a/packages/browser/src/gen/index.d.ts
+++ b/packages/browser/src/gen/index.d.ts
@@ -9132,7 +9132,6 @@ export namespace Browser {
             default_icon?: ManifestIcons | undefined;
             default_title?: string | undefined;
             default_popup?: string | undefined;
-            default_state?: "enabled" | "disabled" | undefined;
         }
 
         /** Source: https://developer.chrome.com/docs/extensions/reference/permissions-list */
--- a/packages/wxt/src/core/utils/__tests__/manifest.test.ts
+++ b/packages/wxt/src/core/utils/__tests__/manifest.test.ts
@@ -18,6 +18,7 @@ import {
   ContentScriptEntrypoint,
   Entrypoint,
   OutputAsset,
+  TargetManifestVersion,
 } from '../../../types';
 import { wxt } from '../../wxt';
 import { mock } from 'vitest-mock-extended';
@@ -44,7 +45,6 @@ describe('Manifest Utils', () => {
               '16': '/icon/16.png',
             },
             defaultTitle: 'Default Title',
-            defaultState: 'enabled',
           },
           outputDir: outDir,
           skipped: false,
@@ -64,7 +64,6 @@ describe('Manifest Utils', () => {
           action: {
             default_icon: popup.options.defaultIcon,
             default_title: popup.options.defaultTitle,
-            default_state: popup.options.defaultState,
             default_popup: 'popup.html',
           },
         };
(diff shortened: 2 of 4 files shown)


packages/browser/src/gen/index.d.ts
aklinker1, line 1: This is a generated file based on `@types/chrome`. To make changes to it's types, open a PR in that repo instead.

For this PR, you're welcome to just `@ts-expect-error` like this PR does: https://github.com/wxt-dev/wxt/pull/2097/changes#diff-50df625c6f95cac7dcf47882681f0bc55fb27a66858fc50256f26fb541b25c43

aklinker1 (changes requested):
Just undo your change to the browser package and add a `@ts-expect-error` like mentioned in the comemnt. Then I'll approve!

Sorry for the delay, and thanks for opening a PR 😄

PatrykKuniczak:
@kairi003 Are you want to go on?

PatrykKuniczak:
@kairi003 One more poke :)

kairi003:
@aklinker1 @PatrykKuniczak 
Sorry for the delay. I’ve made the requested changes based on your comments.
Could you please confirm whether this looks good?

packages/wxt/src/types.ts
```diff
@@ -760,6 +760,7 @@ export interface PopupEntrypointOptions extends BaseEntrypointOptions {
   mv2Key?: PerBrowserOption<'browser_action' | 'page_action'>;
   defaultIcon?: Record<string, string>;
   defaultTitle?: PerBrowserOption<string>;
+  defaultState?: PerBrowserOption<'enabled' | 'disabled'>;
```
PatrykKuniczak, line 768: Let's add info that's only available for Chrome, like there:
https://github.com/wxt-dev/wxt/pull/2010/changes#diff-d28c2725d1b8ea5238aee0b3d8e6c8aabdee0772ea03a560fcfe685561727f0cR765-R770

PatrykKuniczak (changes requested):

packages/wxt/src/types.ts
kairi003: Thank you for your review!
Added a Chrome-only note for defaultState at 7fff02e .

PatrykKuniczak (approved):
LGTM

PatrykKuniczak:
@kairi003

PatrykKuniczak:
@kairi003 WAIT, tests issue.
Let's adjust it

kairi003:
> @kairi003 WAIT, tests issue. Let's adjust it

@PatrykKuniczak 
I fixed a test failure by making default_state only emitted for non-Firefox builds. Does this look good?

PatrykKuniczak:
@kairi003 Let's see details of this error, fix from previous commit doesn't work.
Let's revert this commit, because we'll handle removing props in other way someday #2301

packages/wxt/src/core/utils/__tests__/manifest.test.ts
aklinker1, line 1: Can you add a new test for this instead of updating existsing ones? There's still value in making sure when a state isn't set, nothing gets set in the manifest.

If you revert your changes to the existing tests, they will verify that. Then one small test will make sure it shows up when set.

aklinker1 (approved):

packages/wxt/src/core/utils/__tests__/manifest.test.ts
aklinker1: I'll handle this, I've made you wait long enough

PR #2013, opened 2025-12-21
fix: normalize path for createAnalytics of analytics/index.ts

### LINKED ISSUE #2008: Error on run `pnpm check` for `analytics` package
### Describe the bug

You can see error, on the screenshot:

Image

Problem is in `packages/analytics/.wxt/analytics/index.ts`

Image

### Reproduction

1. Clone repo 
2. Run `pnpm i`
3. Run `pnpm check`

### Steps to reproduce

_No response_

### System Info

```shell
System:
    OS: Windows 11 10.0.26200
    CPU: (12) x64 AMD Ryzen 5 3600 6-Core Processor              
    Memory: 13.30 GB / 31.91 GB
  Binaries:
    Node: 25.1.0
    npm: 11.6.2
    pnpm: 10.15.1
    bun: 1.3.3
  Browsers:
    Chrome: 143.0.7499.110
    Edge: Chromium (140.0.3485.54)
    Firefox: 144.0
    Internet Explorer: 11.0.26100.7309
  npmPackages:
    wxt: workspace:* => 0.20.13
```

### Used Package Manager

pnpm

### Validations

PatrykKuniczak:
### Overview

I've use normalize path for `createAnalysis` func in output.

### Manual Testing

Let's run `pnpm -F analytics prepare` and see new output of `.wxt/analytics/index.ts`, should be valid independent of OS.

### Related Issue

This PR closes #2008

### DIFF
--- a/packages/analytics/modules/analytics/index.ts
+++ b/packages/analytics/modules/analytics/index.ts
@@ -8,6 +8,7 @@ import {
 } from 'wxt/modules';
 import { relative, resolve } from 'node:path';
 import type { AnalyticsConfig } from './types';
+import { normalizePath } from 'wxt';
 
 declare module 'wxt/utils/define-app-config' {
   export interface WxtAppConfig {
@@ -42,7 +43,7 @@ export default defineWxtModule({
       `import { createAnalytics } from '${
         process.env.NPM
           ? clientModuleId
-          : relative(wxtAnalyticsFolder, clientModuleId)
+          : normalizePath(relative(wxtAnalyticsFolder, clientModuleId))
       }';`,
       `import { useAppConfig } from '#imports';`,
       ``,


packages/analytics/modules/analytics/index.ts
```diff
@@ -8,6 +8,7 @@ import {
 } from 'wxt/modules';
 import { relative, resolve } from 'node:path';
```
aklinker1, line 9: Should we just use `node:path/unix` instead of exposing the `normalizePath` function?

PatrykKuniczak: I think you mean about `posix`, but it:
``` import { relative, resolve } from 'node:path/posix'; ```
doesn't work, because it still puts `\`

And afterall(From `.wxt/analytics/index.ts`):
image
But i don't know how to achive that.

And i was wondering if `clientModuleId` and `pluginModuleId` should be also normalize?

Comment (aklinker1) [reply packages/analytics/modules/analytics/index.ts:9]: > `import { relative, resolve } from 'node:path/posix';`
> doesn't work, because it still puts `\`

You sure about that?

Ran a simple test in github actions via https://github.com/wxt-dev/wxt/pull/2074. Both the linux and windows tests output with `/` when using `node:path/posix`.

Image

Comment (PatrykKuniczak) [reply packages/analytics/modules/analytics/index.ts:9]: @aklinker1 I'll try it one more time tomorrow, but i've looked at output file and there was wrong `\`

Comment (PatrykKuniczak) [reply packages/analytics/modules/analytics/index.ts:9]: @aklinker1 
I've removed `.wxt` and try to rebuild, but without `normalizePath` it doesn't work, i have error:
image

Comment (PatrykKuniczak) [reply packages/analytics/modules/analytics/index.ts:9]: I'm also tried to import `resolve` from `node:path` and only `relative` from `posix` but, then i have:
image

In the output file

**Comment by PatrykKuniczak** [comment]
@aklinker1 I'm also simplified import path in output file:
image

**Comment by PatrykKuniczak** [comment]
@aklinker1 OK, IT DOESN'T WORK xD
Reverted

**Comment by aklinker1** [comment]
Ok. Export `normalizePath`, but don't add a new export path for it - those are for runtime utils only. Export it from `wxt` directly since it's a build time util.

Edit: Actually, we should break this PR apart, one for exporting `normalizePath` from WXT so it get's a dedicated entry in the changelog next release.

File: packages/analytics/modules/analytics/index.ts [hunk 2]
```diff
@@ -8,6 +8,7 @@ import {
...
 import type { AnalyticsConfig } from './types';
+import { normalizePath } from 'wxt/utils/paths';
```
aklinker1:
```suggestion
import { normalizePath } from 'wxt';
```

aklinker1 (changes requested):

packages/analytics/modules/analytics/index.ts
PatrykKuniczak: Done in #2080 let's check it

PatrykKuniczak:
@aklinker1 Are you able to enable `update with rebase` as default and only option?
This change can enforce devs to resolve conflicts before merge PR :)
Because of:
image

PatrykKuniczak:
@aklinker1 Finally DONE :)

aklinker1:
> @aklinker1 Are you able to enable `update with rebase` as default and only option? This change can enforce devs to resolve conflicts before merge PR :)
> Because of:
> image

No, that's not a setting on github. You can't merge a PR without resolving conflicts anyways, either a merge commit or a rebase both do the job.

PatrykKuniczak:
@aklinker1 But if you can(if you didn't done it already), let's enable auto removing branches after merge, it'll be useful at least on our branches(because with auto-merge) i usually doesn't check PR after merge, with auto branch remove, we can avoid a loot of old unnecessary branches :)

aklinker1:
It's already enabled. Same problem, I think it's related to your PRs being from a fork.

PR #2014, opened 2025-12-21
fix: Errors and remove unused code of docs and blog

PatrykKuniczak:
### Overview

1. Removed unused code
2. Add missing types and removed unnecessary
3. Removed duplicated CSS props
4. Make real const UPPER_CASE
5. Resolve conflict of vitepress vite version and version on our project (Add `@ts-expect-error`, for permanent solution, we need to wait till [vitepress v2.0](https://github.com/vuejs/vitepress/blob/main/CHANGELOG.md#200-alpha7-2025-06-24) will be released, that's bizzare, all of those changes are in alpha stage for a year)
6. I've wanted to declare breakpoints vals as vars in `:root` and use it inside `@media` but it isn't possible in native CSS and possible solutions isn't good enough, and i need to give up with this idea, only good solution IMHO is to use Tailwindcss to achive this in good quality :)

### Manual Testing

Let's check if blog works well :)

### DIFF
--- a/docs/.vitepress/components/Icon.vue
+++ b/docs/.vitepress/components/Icon.vue
@@ -1,19 +1,19 @@
 <script lang="ts" setup>
 import { computed } from 'vue';
 
-const props = defineProps<{
+const { name, icon } = defineProps<{
   name: string;
   icon?: string;
 }>();
 
 const src = computed(() => {
-  if (props.icon) return props.icon;
-  return `https://raw.githubusercontent.com/PKief/vscode-material-icon-theme/main/icons/${props.name.toLowerCase()}.svg`;
+  if (icon) return icon;
+  return `https://raw.githubusercontent.com/PKief/vscode-material-icon-theme/main/icons/${name.toLowerCase()}.svg`;
 });
 </script>
 
 <template>
-  <img :src="src" :alt="`${name} Logo`" />
+  <img :alt="`${name} Logo`" :src="src" />
 </template>
 
 <style scoped>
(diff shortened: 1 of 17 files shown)


PatrykKuniczak:
@aklinker1 I want to configure `Stylelint`(maybe on seperate PR and then rebase here) for prettify CSS of Vue components, but i don't know how to integrate it with `check` lib, are you able to help me, here or on Discord DM(I've sent it :))

Second approach could be to use plugin for prettier, but stylelint works like eslint + prettier for CSS and IMHO it could be better for that case.

docs/.vitepress/components/Icon.vue
```diff
@@ -1,19 +1,19 @@
 <script lang="ts" setup>
 import { computed } from 'vue';
 
-const props = defineProps<{
+const { name, icon } = defineProps<{
```
aklinker1: Is vitepress using vue 3.5 internally?

https://vuejs.org/guide/components/props.html#reactive-props-destructure

PatrykKuniczak:
@aklinker1 Also take a look on that one

PatrykKuniczak:
@aklinker1 You've already answered about `stylelint`, only think it's `TODO` with this is review, let's take your time ❤️

PatrykKuniczak:
@aklinker1 Are you able to take a look on that?

aklinker1:
Could you break this PR up too?

PatrykKuniczak:
@aklinker1 Yeah, i'll do it in next few days

PatrykKuniczak:
@aklinker1 I split it, i'll review it one more time in next hour and create 1 more and when it'll be merged i'll rebase it here and check if i haven't missed sth :)

PR #2015, opened 2025-12-21
chore: fix recognize lack of argument for scripts

PatrykKuniczak:
### Overview

I fixed check for no argument for scripts

```ts
if (pkg == null) 
```
which don't work(because if user don't type nothing `undefined` is return
with
```ts
if(!pkg)
```
which works :)

And i cleaned code a little, of unnecessary console.log()

### Manual Testing

Try to use all of those scripts without argument.

### DIFF
--- a/scripts/bump-package-version.ts
+++ b/scripts/bump-package-version.ts
@@ -8,8 +8,7 @@ import {
 import spawn from 'nano-spawn';
 import { getPkgTag, grabPackageDetails, listCommitsInDir } from './git';
 import { consola } from 'consola';
-import fs from 'fs/promises';
-import fsExtra from 'fs-extra';
+import fs from 'fs-extra';
 import path from 'node:path';
 
 const pkg = process.argv[2];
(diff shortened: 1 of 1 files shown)


PatrykKuniczak:
Sorry for a little mess, i haven't checked rest of the code for `fs-extra` references, we need to fix it in another way, change `fs-extra` to `fs/promises` isn't solutions :(

aklinker1:
Use `fix: ...` when the PR fixes a bug in a package, use `chore: ...` when making a change that only effects contributors, like updating a script, or refactors that shouldn't change runtime behavior.

PatrykKuniczak:
@aklinker1 Are you able to enable `auto merge` for repo?

aklinker1:
> @aklinker1 Are you able to enable `auto merge` for repo?

Down to give it a go. Just enabled it.

PatrykKuniczak:
@aklinker1 You can also enable auto deleting branches after merge.

aklinker1:
> @aklinker1 You can also enable auto deleting branches after merge.

It is. But since this is a fork, I can't delete your branch.

Now that you're a maintainer, you don't need to use a fork.

PatrykKuniczak:
@aklinker1 Yeah your, right, i know but it can work for internal branches.

And yeah second right, i can, but those PRs was opened, when i wasn't.
I'll create maybe a couple more as forks, because it can be easier to split #2022 :)

aklinker1:
Also, auto-merge is really nice, thanks for the suggestion!

Issue #2004, opened 2025-12-16
Config to specify IIFE name or disable it altogether


aklinker1:
> Before that, I want to run an alternative by you: would it make more sense to add this to `defineContentScript` (as an option like `cssInjectionMode`, `matches`, etc) instead of `wxt.config.ts`?
> 
> ```ts
> export default defineContentScript({
>   iffeName: "example",
>   // OR `globalName: string | boolean`, letting people disable the iffe name in the  output as well as give it a different name.
> 
>   main() {
>     // ...
>   }
> })
> ```
> 
> In this version, it would default to `true` and the name would be generated based on the content script name, but in the next major version would could change the default to `false` so a name isn't even generated in the first place.
> 

&mdash; _Originally posted by @aklinker1 in https://github.com/wxt-dev/wxt/pull/1897#pullrequestreview-3385640996_

aklinker1:
So this issue needs two PRs:

1. Into `main`: Add the `iifeName: boolean | string` option and default it to `true`
2. Into `major`: Change the default to `false`

@jviney [offered to implement this](https://github.com/wxt-dev/wxt/pull/1897#issuecomment-3459219514)... 2 months ago, so I don't know if that offer still stands or if someone else can pick it up. I just created this issue to track it and make sure it's not forgotten since I think it's a really good idea.

Riddhish1:
Hey @aklinker1 , I’d like to pick this up if no one is currently working on it. I’m planning to start with the PR to `main` (adding `iifeName: boolean | string` defaulting to `true`). Let me know if that sounds good.

jviney:
> Hey [@aklinker1](https://github.com/aklinker1) , I’d like to pick this up if no one is currently working on it. I’m planning to start with the PR to `main` (adding `iifeName: boolean | string` defaulting to `true`). Let me know if that sounds good.

I posted a draft of this a few days ago over here https://github.com/wxt-dev/wxt/pull/2017. Any feedback welcome.

aklinker1:
@jviney PR looks really good, left a few comments with some small changes.

PR #2017, opened 2025-12-24
feat: Add `globalName` entrypoint option.

jviney:
### Overview

Implements https://github.com/wxt-dev/wxt/issues/2004

The IIFE name is not currently configurable for content scripts and unlisted scripts. The variable name can be important when inserting scripts with world=MAIN or directly via a `<script>` tag to ensure there is no conflict with an existing variable name on the page.

This PR is a followup to the discussion in #1897 where we decided to add a `globalName` option.

Questions:
* The `globalName` property is available on all entrypoints - should we restrict this to just content scripts and unlisted scripts in the type definitions?
* Is there any benefit to having the `globalName` option available for other entrypoint types?
* How shall we indicate that the default value will switch to the behaviour of `false` in a future WXT version? Would this version change the generated output for other script types too (eg: background, popup), or will we leave them generating a named IIFE.

### Manual Testing

Specify the `globalName` option on a content script or unlisted script entrypoint and check the generated JS output matches what's expected for the given option.

### DIFF
--- a/packages/wxt/e2e/tests/output-structure.test.ts
+++ b/packages/wxt/e2e/tests/output-structure.test.ts
@@ -459,4 +459,104 @@ describe('Output Directory Structure', () => {
       "
     `);
   });
+
+  describe('globalName option', () => {
+    it('generates an IIFE with a default name', async () => {
+      const project = new TestProject();
+      project.addFile(
+        'entrypoints/content.js',
+        `export default defineContentScript({
+          matches: ["*://*/*"],
+          main() {},
+        })`,
+      );
+
+      await project.build({ vite: () => ({ build: { minify: false } }) });
+
+      const output = await project.serializeFile(
+        '.output/chrome-mv3/content-scripts/content.js',
+      );
+      expect(output).toMatch(/^var content\s?=[\s\S]*^content;$/gm);
+    });
+
+    it('generates an IIFE with a specific name', async () => {
+      const project = new TestProject();
+      project.addFile(
+        'entrypoints/content.js',
+        `export default defineContentScript({
+          globalName: "MyContentScript",
+          matches: ["*://*/*"],
+          main() {},
+        })`,
+      );
+
+      await project.build({ vite: () => ({ build: { minify: false } }) });
+
+      const output = await project.serializeFile(
+        '.output/chrome-mv3/content-scripts/content.js',
+      );
+      expect(output).toMatch(
+        /^var MyContentScript =[\s\S]*^MyContentScript;$/gm,
+      );
+    });
+
+    it('generates an IIFE with a specific name provided by a function', async () => {
+      const project = new TestProject();
+      project.addFile(
+        'entrypoints/content.js',
+        `export default defineContentScript({
+          globalName: () => "MyContentScript",
+          matches: ["*://*/*"],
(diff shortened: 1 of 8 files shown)


packages/wxt/e2e/tests/output-structure.test.ts
```diff
@@ -459,4 +459,377 @@ describe('Output Directory Structure', () => {
...
+        content_dynamic;"
+      `);
+    });
+
+    it('generates an anonymous IIFE when not minified', async () => {
```
aklinker1: Let's remove the unminified example, it's not necessary.

packages/wxt/src/core/builders/vite/plugins/iifeAnonymous.ts
aklinker1, line 1: ~~I don't think we need a full plugin to do this... If I remember correctly, vite will already output a anonymous IIFE if we don't specify a global name?~~

packages/wxt/src/types.ts
```diff
@@ -551,6 +551,25 @@ export interface BaseEntrypointOptions {
...
+   * Available options:
+   * - `true`: automatically generate a name for the IIFE based on the entrypoint name
+   * - `false`: Output the IIFE without a variable name, making it anonymous. This is the safest option
+   *  to avoid conflicts with existing variables on the page. This will become the default in a future version of WXT.
+   * - `string`: Use the provided string as the global variable name.
+   * - `function`: A function that receives the entrypoint and returns a string to use as the variable name.
+   *
+   * @default true
+   */
+  globalName?: string | boolean | ((entrypoint: Entrypoint) => string);
```
aklinker1, line 587: Yeah, this needs to be moved to the content script and unlisted scripts types. I think it's fine if you duplicate this or feel free to create a shared base type.

packages/wxt/e2e/tests/output-structure.test.ts
```diff
@@ -459,4 +459,377 @@ describe('Output Directory Structure', () => {
...
+          content;"
+        `);
```
aklinker1: What about using this for the expectation?

```suggestion
      const output = await project.serializeFile(
        '.output/chrome-mv3/content-scripts/content.js',
      )
      expect(output).toMatch(/^var content=[\s\S]*^content;$/gm);
```

If a test doesn't require the full file to match, I'd prefer we narrow it down to only what is required. That way if something unrelated to the global name changes and effects this test, it still passes.

packages/wxt/src/core/builders/vite/plugins/iifeAnonymous.ts
aklinker1: I am wrong. We need a plugin like you created!

jviney: Great!

packages/wxt/src/types.ts
jviney: Great - done.

packages/wxt/e2e/tests/output-structure.test.ts
jviney: Agreed - done.

jviney:
Ready for another review.

Issue #2018, opened 2025-12-24
Templates scaffolded by wxt@latest init don't use the latest WXT version


InvictusNavarchus:
### Describe the bug

When a user runs `npx wxt@latest init`, it initializes a WXT project with a chosen pre-made template. The problem is, the WXT version is hardcoded in the template (currently 0.20.6). Even the [auto upgrader script](https://github.com/wxt-dev/wxt/blob/main/scripts/upgrade-deps.ts) explicitly ignores upgrading the WXT version.

```typescript
        if (
          name === 'wxt' ||
          version.startsWith('workspace:') ||
          IGNORED_PACKAGES.includes(name)
        )
          return;
```

This is probably intentional, as highlighted in the CONTRIUBTING.md:

```
Note that templates are hardcoded to a specific version of `wxt` from NPM, they do not use the local version
```

However, it contradicts the `latest` directive in `wxt@latest`. When a user runs that command, they will expect to get the current latest WXT version, not an outdated one, let alone 7 months out of date.

I've proposed a PR #2019 that updates the WXT version in the templates to 0.20.13 (current latest), but that's just a temporary fix. A better approach would be to either

1. stop hardcoding WXT in the templates and just use the latest version. One approach would be to apply special exclusion criteria for WXT in the auto-upgrade script to allow auto-upgrade for the templates, not internal packages.
2. if auto-latest is truly unwanted, consider renaming `@latest` directive to something more accurate

### Reproduction

-

### Steps to reproduce

1. Run `npx wxt@latest init`
2. See package.json
3. You get the exact WXT version currently hardcoded in the templates

### System Info

```shell
System:
    OS: Linux 6.6 Ubuntu 24.04.3 LTS 24.04.3 LTS (Noble Numbat)
    CPU: (12) x64 AMD Ryzen 5 6600H with Radeon Graphics
    Memory: 1.70 GB / 6.65 GB
    Container: Yes
    Shell: 5.2.21 - /bin/bash
  Binaries:
    Node: 22.20.0 - /home/invictus/.nvm/versions/node/v22.20.0/bin/node
    npm: 10.9.3 - /home/invictus/.nvm/versions/node/v22.20.0/bin/npm
    pnpm: 10.15.1 - /home/invictus/.nvm/versions/node/v22.20.0/bin/pnpm
    bun: 1.3.1 - /home/invictus/.bun/bin/bun
```

### Used Package Manager

bun

### Validations

[...]

PR #2019, opened 2025-12-24
chore: update wxt version to ^0.20.13 in all templates

InvictusNavarchus:
### Overview

The WXT versions in templates were last updated on #1648. It's been 7 months since then. I believe it should be time to upgrade.

There is dependency auto-upgrade script that was introduced in #1875. It explicitly ignores WXT (intentional? refer to #2018). It was last executed for #2000, but the WXT versions in the templates were never manually updated. Potentially an oversight.

### Related Issue

Related to #2018


aklinker1:
Woops, good find! I'll merge this then update the release script to update the templates automatically, as suggested on the issue.

aklinker1 (approved):

PR #2035, opened 2026-01-11
fix: content script is incorrectly invalidated when injected multiple times

namuorg:
See #1938 for the original PR and discussion (I didn't realize that deleting my fork would automatically close the existing PR for this)

### DIFF
--- a/packages/wxt/src/utils/content-script-context.ts
+++ b/packages/wxt/src/utils/content-script-context.ts
@@ -243,43 +243,56 @@ export class ContentScriptContext implements AbortController {
   }
 
   stopOldScripts() {
-    // Use postMessage so it get's sent to all the frames of the page.
-    window.postMessage(
-      {
-        type: ContentScriptContext.SCRIPT_STARTED_MESSAGE_TYPE,
-        contentScriptName: this.contentScriptName,
-        messageId: Math.random().toString(36).slice(2),
-      },
-      '*',
+    document.dispatchEvent(
+      new CustomEvent(ContentScriptContext.SCRIPT_STARTED_MESSAGE_TYPE, {
+        detail: {
+          contentScriptName: this.contentScriptName,
+          messageId: Math.random().toString(36).slice(2),
+        },
+      }),
     );
   }
 
-  verifyScriptStartedEvent(event: MessageEvent) {
-    const isScriptStartedEvent =
-      event.data?.type === ContentScriptContext.SCRIPT_STARTED_MESSAGE_TYPE;
+  verifyScriptStartedEvent(event: CustomEvent) {
     const isSameContentScript =
-      event.data?.contentScriptName === this.contentScriptName;
-    const isNotDuplicate = !this.receivedMessageIds.has(event.data?.messageId);
-    return isScriptStartedEvent && isSameContentScript && isNotDuplicate;
+      event.detail?.contentScriptName === this.contentScriptName;
+    const isNotDuplicate = !this.receivedMessageIds.has(
+      event.detail?.messageId,
+    );
+    return isSameContentScript && isNotDuplicate;
   }
 
   listenForNewerScripts(options?: { ignoreFirstEvent?: boolean }) {
     let isFirst = true;
 
-    const cb = (event: MessageEvent) => {
-      if (this.verifyScriptStartedEvent(event)) {
-        this.receivedMessageIds.add(event.data.messageId);
-
-        const wasFirst = isFirst;
-        isFirst = false;
-        if (wasFirst && options?.ignoreFirstEvent) return;
-
-        this.notifyInvalidated();
+    const cb: EventListener = (event) => {
+      if (
+        !(event instanceof CustomEvent) ||
+        !this.verifyScriptStartedEvent(event)
+      ) {
+        return;
       }
+
+      this.receivedMessageIds.add(event.detail.messageId);
+
+      const wasFirst = isFirst;
+      isFirst = false;
+      if (wasFirst && options?.ignoreFirstEvent) {
+        return;
+      }
+      this.notifyInvalidated();
     };
 
-    addEventListener('message', cb);
-    this.onInvalidated(() => removeEventListener('message', cb));
+    document.addEventListener(
+      ContentScriptContext.SCRIPT_STARTED_MESSAGE_TYPE,
+      cb,
+    );
+    this.onInvalidated(() =>
+      document.removeEventListener(
+        ContentScriptContext.SCRIPT_STARTED_MESSAGE_TYPE,
+        cb,
+      ),
+    );
   }
 }
 
--- a/packages/wxt/src/utils/content-script-context.ts
+++ b/packages/wxt/src/utils/content-script-context.ts
@@ -247,6 +247,17 @@ export class ContentScriptContext implements AbortController {
         },
       }),
     );
+
+    // Send message using `window.postMessage` for backwards compatibility to invalidate old versions before WXT changed to `document.dispatchEvent`
+    // TODO: Remove this once WXT version using `document.dispatchEvent` has been released for a while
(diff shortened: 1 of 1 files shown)


packages/wxt/src/utils/content-script-context.ts
```diff
@@ -243,43 +239,56 @@ export class ContentScriptContext implements AbortController {
...
     window.postMessage(
       {
         type: ContentScriptContext.SCRIPT_STARTED_MESSAGE_TYPE,
         contentScriptName: this.contentScriptName,
-        messageId: Math.random().toString(36).slice(2),
+        messageId: this.id,
       },
       '*',
     );
   }
```
PatrykKuniczak, line 261: I think we can avoid it, and release only your solution, without this fallback

PatrykKuniczak:
@namuorg To be honest, i don't know why @aklinker1 worry about previous versions of scripts.

Extension update on browser restart, and then all scripts and entire browser engine shut down, and then starts with updated WXT and there's no way to have older version of content script than e.g `popup`.
Maybe i'm wrong, but IMO we shouldn't consider it as a `breaking changes`.

We have bug, let's merge it and solve the bug, if we were merge it e.g month ago, all the edge cases, about you were talking about, which can happend, will have a time to happend, and we won't remember about it.

@namuorg That was the most professional PR, which i've seen in my live, you shown us video with repro and bug fix, that's outstanding and we shouldn't block it, because of possible one edge-case.

@aklinker1 We need to go forward faster IMO, don't think too much(I don't want to say, merge without brain), but i think we need to let people know, if they'll open PR, that'll be check in e.g 1 week, maybe sth about this time, they can take any feedback(we all working, and don't have much time, to spend on open source project, but just take a look od PR, give some feedback and let people push forward.

IMO this approach, can drastically improve number of PRs by `first time contributors`, and let us(active maintainers), code less but have more work done)

@aklinker1 You have been working a looot, because i see amount of PRs, which you've merged, but i think you should do less by your own, but let people do work and only checking it.

You have done a REALLY GREAT JOB, but let be some kind of DEV-Project Manager.
Let open source be `Open Source`.

Let me know, what are you think, maybe i'm wrong.

namuorg:
Thanks @PatrykKuniczak, I appreciate the support! The videos took a while to set up and make so I'm glad they're helpful for understanding the PR and what it fixes. :)

I understand @aklinker1's concern about not wanting to introduce breaking changes, but my understanding is that the concern is due to an inaccurate understanding of how `window.postMessage` works with a target of `*`. The video in https://github.com/wxt-dev/wxt/pull/1938#discussion_r2467595859 should make it clear that the PR actually fixes an existing bug of content scripts inside iframes not being invalidated.

I'm hoping that we could get this PR merged. I'm happy to address any additional questions or concerns about this change. Thanks!

aklinker1:
> Extension update on browser restart, and then all scripts and entire browser engine shut down, and then starts with updated WXT and there's no way to have older version of content script than e.g popup.
Maybe i'm wrong, but IMO we shouldn't consider it as a breaking changes.
>
> We have bug, let's merge it and solve the bug, if we were merge it e.g month ago, all the edge cases, about you were talking about, which can happend, will have a time to happend, and we won't remember about it.

@PatrykKuniczak This is factually wrong. Content scripts do continue running after the extension is uninstalled or updated in Chrome. This is a [known problem](https://stackoverflow.com/questions/57468219/how-to-remove-orphaned-script-after-chrome-extension-update) (if not well known) that extension devs need to deal with.

The extension update scenario is the main reason this util exists and is so integrated with WXT - so that devs don't have to deal with having two versions of their content scripts being active at the same time. It also serves nicely as a solution during development for when the extension is reloaded or new version of the content script injected dynamically.

For example, I develop an extension that changes the video player on Crunchyroll. If we don't handle this scenario, when my extension is updated while people are watching anime, the custom player would remain on the page and paritally stop working. Worse, since Crunchyroll uses SPA navigation, the page won't refresh during a binge session, compounding the errors as different episodes are played. In that case, someone may report the problem or leave a bad review when things break on them. The dev of the extension (me) would then have to spend time debugging and figuring out what happened, but it would be almost impossible find since it wasn't a code change. If they did figure out the problem came from WXT, then now it's my problem as a maintainer.

So as you can see, making this breaking change would have real consequences and effect people's trust in WXT's stability.

@PatrykKuniczak if you're going to be on the team, preventing breaking changes is the number one thing to look out for in PRs. I hope you understand why after this discussion.

> We need to go forward faster IMO, don't think too much(I don't want to say, merge without brain), but i think we need to let people know, if they'll open PR, that'll be check in e.g 1 week, maybe sth about this time, they can take any feedback(we all working, and don't have much time, to spend on open source project, but just take a look od PR, give some feedback and let people push forward.

I will make no such promises. I work on WXT in my free time and when life happens, there's no guarantee I can fulfill that commitment.

I agree that we should move faster, but once again, we can't blindly merge PRs "because of one small edge case" on a critical part of the code. I was so hesitant to merge the PR at first because of how critical the `ContentScriptContext` is and I didn't understand the change completely, even if I saw it was working in one example. Then once the author  responded explaining it in more detail and... I missed both comments.

> You have been working a looot, because i see amount of PRs, which you've merged, but i think you should do less by your own, but let people do work and only checking it.
>
> You have done a REALLY GREAT JOB, but let be some kind of DEV-Project Manager.
> Let open source be Open Source.
>
> Let me know, what are you think, maybe i'm wrong.

There has not been any open source contributions towards the 1.0 milestone. No one has seen fit to update dependencies on the project. People just report bugs and sometimes fix them themselves.

With the current team, at least before I added you @PatrykKuniczak, I was the only (semi) active contributor. That means I have to balance security, maintainence, and feature work.  I was able to merge or close some 50 PRs authored by other people in the last few months across all my OSS projects, I think that's pretty good for one person in their spare time.

We just need more people on the team that I can trust to uphold the quality, stability, and vision of the framework, that's the first step to moving faster &ndash; a step I've taken with you :smile:

Let's circle back on this conversation in a month after you've had more time working on the project.

aklinker1:
Anyways... sorry for the delay @namuorg!

As mentioned [here](https://github.com/wxt-dev/wxt/pull/1938#discussion_r2780232447), my concern about `CustomEvent` behaving differently than `window.postMessage` has been resolved by your response to my initial comment. I'm on board with custom events behaving the same way as messages, but better since they're synchronous without race conditions.

~~However, one breaking change remains - that content scripts listening for messages will not be invalidated on an extension's next update.~~

~~Solution here is simple... just keep calling `postMessage` for now.~~

~~However, then we have to decide when to remove that legacy call to `postMessage`... Let's say in WXT v2.0. At that point, sometime in the far, undetermined future, the chance that someone is upgrading from 0.x to 2.0 will be minimal enough that I'd be OK with making the breaking change silently without recourse.~~

_Edit: I just looked at the changes in this PR and you've already done this! Way ahead of me. Approving and merging_ :smile:

aklinker1 (approved):

packages/wxt/src/utils/content-script-context.ts
aklinker1: As mentioned in https://github.com/wxt-dev/wxt/pull/2035#issuecomment-3868948500 and https://github.com/wxt-dev/wxt/pull/2035#issuecomment-3868961784, we need to keep the fallback.

namuorg:
Thanks @aklinker1! Appreciate you taking another look and all the work you've done on this project in your free time <3

PatrykKuniczak:
@aklinker1 i'm trying me best, to understand this project, thanks for your feedback, it's have much value :)

50 PRs is very good, as one person, and i'm appreciate it

But i want to bring this project alive, which i see is awaked right now :)

@namuorg For next time, if you deleted a branch, you can restore it and reopen PR, by this, you can avoid duplicated PRs

PR #1939, opened 2025-10-15
fix: use uiContainer to apply position when creating shadow root ui

namuorg:
### Overview

This is a companion PR for https://github.com/aklinker1/webext-core/pull/118

Once the shadow DOM uses a `<div>` instead of constructing a full HTML document, the current approach of finding the `<html>` element won't work. Instead, the `uiContainer` will be used to position the content script UI.

### Manual Testing

How to test this change:

1. Create a new WXT extension using `pnpm dlx wxt@latest init`
2. Use modified `wxt`
    - Clone `wxt` repo
    - Check out this PR
    - Run `pnpm build` in `wxt/packages/wxt`
    - In the extension directory, run `pnpm link <path-to-wxt/packages/wxt>` to use modified version of `wxt`
3. Use modified `webext-core`
    - Clone `webext-core` repo
    - Check out https://github.com/aklinker1/webext-core/pull/118
    - Run `bun run build` in `webext-core/packages/isolated-element`
    - In the `wxt` directory, run `pnpm link <path-to-webext-core/packages/isolated-element>` to use modified version of `@webext-core/isolated-element`
4. Use `createShadowRootUi` with position `overlay` or `modal` to verify that the element is position as desired

Example code for testing:
```typescript
export default defineContentScript({
  matches: ["<all_urls>"],
  cssInjectionMode: "ui",

  async main(ctx) {
    const ui = await createShadowRootUi(ctx, {
      name: "example-ui",
      position: "overlay", // <-- change between "inline", "overlay", "modal"
      anchor: "body",
      onMount(container) {
        const app = document.createElement("div");
        app.textContent = "Hello world!";
        app.style.padding = "8px";
        app.style.border = "2px solid red";
        container.append(app);
      },
    });

    ui.mount();
  },
});
```

PR #2036, opened 2026-01-11
fix: use uiContainer to apply position when creating shadow root ui

namuorg:
See #1939 for the original PR and discussion (I didn't realize that deleting my fork would automatically close the existing PR for this)


PatrykKuniczak:
@aklinker1 This is acceptable?
Because last year, you want him to implement flag instead of change it to `uiContainer`

namuorg:
@aklinker1 

To summarize the context here since it's been a few months: This is a companion PR to https://github.com/aklinker1/webext-core/pull/118 which fixes `@webext-core/isolated-element` incorrectly creating a full HTML document when setting up the shadow DOM.

This seemed to have been originally done in order to work around Tailwind not properly supporting the shadow DOM, but this was fixed upstream so we can set up the shadow DOM properly as a fragment of the DOM tree. Comment about the Tailwind fix here: https://github.com/aklinker1/webext-core/pull/118#issuecomment-3445949560

aklinker1 (approved):
This PR is fine to merge, `uiContainer` is the same element as `shadow.querySelector('html')`.

aklinker1:
Just left a comment over there: https://github.com/aklinker1/webext-core/pull/118#issuecomment-3869608698

I'm fine with switching to fragments if it's a breaking change to both `wxt` and `@webext-core/isolated-element`. Will just need a PR on WXT once the new version of isolated element is released.

aklinker1:
> This PR is fine to merge, uiContainer is the same element as shadow.querySelector('html').

Ah, wait, that's not true. We return the `body` for `isolatedElement`... https://github.com/aklinker1/webext-core/blob/a62279700692f548adde8f504e5013084b70caaf/packages/isolated-element/src/index.ts#L107

Gonna revert, we need to make this change in the breaking change PR, not a normal one.

_Edit: Done https://github.com/wxt-dev/wxt/commit/2c57840c447f57a84a74024e072369c0da16574f_

aklinker1:
Next steps for getting this added back are noted here: https://github.com/aklinker1/webext-core/pull/128

PR #2049, opened 2026-01-19
More robust way of checking the browser type to fix Chrome 144+ issues

### LINKED ISSUE #2039: browser.devtools is undefined in Chrome 144
### Describe the bug

I was using `browser.devtools.panels`... successfully in Firefox and in Chrome 143. However my code broke with Chrome 144, and according to error messages on the console, `browser.devtools` is undefined.
Here's what I get logging `browser` in Chrome 143 (couldn't manage to copy to clipboard, sorry):

Image

And here's for Chrome 144:

Image

### Reproduction

As stated above.

### Steps to reproduce

_No response_

### System Info

```shell
System:
    OS: Windows 11 10.0.26200
  Binaries:
    Node: 22.15.1
    npm: 10.9.2
  Browsers:
    Chrome: 144.0.7559.60
```

### Used Package Manager

npm

### Validations

Igloczek:
### Overview

Introduce more robust way of checking browser type, based on variables set during the build, instead of the runtime checks, to address the problems reported in #2039

### Manual Testing

```
console.log(browser.devtools)
```
shouldn't return `undefined` in Chrome 144+

### Related Issue

#2039

This PR closes #2039

### DIFF
--- a/packages/browser/src/index.mjs
+++ b/packages/browser/src/index.mjs
@@ -1,5 +1,6 @@
 // #region snippet
-export const browser = globalThis.browser?.runtime?.id
-  ? globalThis.browser
-  : globalThis.chrome;
+export const browser =
+  import.meta.env.FIREFOX || import.meta.env.SAFARI
+    ? globalThis.browser
+    : globalThis.chrome
 // #endregion snippet


aklinker1:
@Igloczek looks like Chrome fixed this on their side: https://github.com/wxt-dev/wxt/issues/2039#issuecomment-3776757832.

I'm a little hesitant to make this change, it means people can't "reuse" a single extension build between multiple browsers and have it function the same way as before... but  at the same time, I like this change. And I don't know how many people reuse builds between browsers (I would think very few, but not zero lol).

Can you rebase it off the `major` branch and target it in this PR? That way we can merge this into the next major release as a breaking change and ensure the change is documented.

aklinker1:
Actually... We can't use `import.meta.env.*` inside the `@wxt-dev/browser` package - this package is separate from WXT so other projects can use it as well. `import.meta.env.FIREFOX` and `import.meta.env.SAFARI` are WXT project specific.

Other non-WXT projects, depending on their build setup, would be able to define these as constants, but I don't want to force them to complicate their build setup if it's not necessary.

So since this was fixed by Chrome and for the above reason, I'm gonna close this PR.

PR #2055, opened 2026-01-27
feat: enable native browser launching in WSL with GUI support

### LINKED ISSUE #1931: WSL runner should check DISPLAY env var before refusing to open browser (WSLg support)
## Description

The WSL runner currently refuses to open browsers unconditionally, but this is outdated given modern WSLg (WSL GUI) support which provides native GUI capabilities with DISPLAY environment variable automatically configured.

## Current Behavior

The WSL runner ([`wsl.mjs`](https://github.com/wxt-dev/wxt/blob/main/packages/wxt/src/core/runners/wsl.ts)) always shows this warning and refuses to open browsers:

```
WARN  Cannot open browser when using WSL. Load ".output/chrome-mv3-dev" as an unpacked extension manually
```

This happens even when `DISPLAY` is set and GUI applications work perfectly.

## Expected Behavior

When `DISPLAY` environment variable is set (indicating WSLg or X11 forwarding), WXT should attempt to open the browser normally instead of refusing.

## Environment

- **OS**: WSL2 with WSLg on Windows 11
- **DISPLAY**: `:10.0` (automatically set by WSLg)
- **WXT Version**: 0.20.11
- **Browser**: Google Chrome in WSL2

## Reproduction

1. Run WXT in WSL2 with WSLg enabled (Windows 11 default)
2. Run `wxt` or `wxt build`
3. Browser fails to open despite DISPLAY being set and GUI support working

## Manual Workaround

Browser opens successfully when launched manually with DISPLAY set:
```bash
DISPLAY=:10.0 /usr/bin/google-chrome --load-extension=.output/chrome-mv3-dev
```

## Proposed Solution

Update the WSL runner to check for DISPLAY before refusing:

```typescript
export function createWslRunner() {
  return {
    async openBrowser() {
      // Check if DISPLAY is set (WSLg/X11 support)
      if (!process.env.DISPLAY) {
        wxt.logger.warn(
          \`Cannot open browser when using WSL without GUI support. Load "\${relative(
            process.cwd(),
            wxt.config.outDir
          )}" as an unpacked extension manually\`
        );
        return;
      }
      
      // DISPLAY is set, delegate to default browser runner
      // (let the normal browser opening logic handle it)
    },
    async closeBrowser() {
    }
  };
}
```

## Related Issues

- #55 - Original WSL support issue that disabled browser opening entirely
- This was appropriate at the time, but WSLg (released 2021) changes the situation

## Impact

This affects all WSL2 users on Windows 11+ who have WSLg enabled by default. The warning is misleading and forces manual browser loading despite GUI support being fully functional.

Seismix:
### Overview

 This PR adds support for launching browsers in WSL environments with GUI capabilities (e.g. WSLg).

###  Changes:
  - Updated WSL runner logic to detect GUI availability and use native browser launching when available
  - Added browser installation checks with helpful error messages when browsers are not found
  - Added tests for WSL scenarios:
    - X11 and Wayland environments
    - WSL without GUI (shows manual install warning)
    - Non-WSL environments (no behavior change)
    - Missing browser installations (helpful error messages)

### Manual Testing

Prerequisites:
  - WSL2 with GUI support (WSLg on Windows 11, or X11/Wayland server with `DISPLAY`/`WAYLAND_DISPLAY` set)
  -  Firefox or Chrome installed in WSL 

  Test steps:
  1. Follow the setup instructions in `CONTRIBUTING.md`
  2. Start the demo extension: `cd packages/wxt-demo && pnpm dev:firefox` (or `pnpm dev` for Chrome)
  3. Verify browser opens automatically with extension loaded

  Expected results:
  - Browser should open automatically in WSL with GUI
  - Extension should load correctly
  - No "Cannot open browser" warnings

### Related Issue

This PR closes #1931

### DIFF
--- a/packages/runner/src/browser-paths.ts
+++ b/packages/runner/src/browser-paths.ts
@@ -41,7 +41,13 @@ export const KNOWN_BROWSER_PATHS: Record<
       '/Applications/Google Chrome.app/Contents/MacOS/Google Chrome',
       '/Applications/Chrome.app/Contents/MacOS/Google Chrome',
     ],
-    linux: [],
+    linux: [
+      // Debian/Ubuntu official Google Chrome package
+      // Prefer the real binary over wrapper scripts to keep extra file descriptors open.
+      '/opt/google/chrome/chrome',
+      '/usr/bin/google-chrome',
+      '/usr/bin/google-chrome-stable',
+    ],
     windows: ['C:\\Program Files\\Google\\Chrome\\Application\\chrome.exe'],
   },
   'chrome-beta': {
--- a/packages/runner/src/index.ts
+++ b/packages/runner/src/index.ts
@@ -1,3 +1,4 @@
 export * from './run';
 export * from './options';
 export * from './install';
+export * from './browser-paths';
--- a/packages/wxt-demo/package.json
+++ b/packages/wxt-demo/package.json
@@ -5,6 +5,7 @@
   "type": "module",
   "scripts": {
     "dev": "buildc --deps-only -- wxt",
+    "dev:firefox": "buildc --deps-only -- wxt -b firefox --mv2",
     "build": "buildc --deps-only -- wxt build",
     "build:all": "buildc --deps-only -- pnpm run --reporter-hide-prefix /^build:all:.*/",
     "build:all:chrome-mv3": "wxt build",
--- a/packages/wxt/package.json
+++ b/packages/wxt/package.json
@@ -22,6 +22,7 @@
     "@webext-core/isolated-element": "^1.1.2",
     "@webext-core/match-patterns": "^1.0.3",
     "@wxt-dev/browser": "workspace:^",
+    "@wxt-dev/runner": "workspace:^",
     "@wxt-dev/storage": "workspace:^1.0.0",
     "async-mutex": "^0.5.0",
     "c12": "^3.3.2",
--- a/packages/wxt/src/core/runners/__tests__/index.test.ts
+++ b/packages/wxt/src/core/runners/__tests__/index.test.ts
@@ -1,4 +1,4 @@
-import { describe, expect, it, vi } from 'vitest';
+import { afterEach, describe, expect, it, vi } from 'vitest';
 import { createExtensionRunner } from '..';
 import { setFakeWxt } from '../../utils/testing/fake-objects';
 import { mock } from 'vitest-mock-extended';
@@ -7,6 +7,7 @@ import { createWslRunner } from '../wsl';
 import { createManualRunner } from '../manual';
 import { isWsl } from '../../utils/wsl';
 import { createWebExtRunner } from '../web-ext';
+import { createWxtRunner } from '../wxt-runner';
 import { ExtensionRunner } from '../../../types';
 
 vi.mock('../../utils/wsl');
@@ -24,7 +25,16 @@ const createManualRunnerMock = vi.mocked(createManualRunner);
 vi.mock('../web-ext');
 const createWebExtRunnerMock = vi.mocked(createWebExtRunner);
 
+vi.mock('../wxt-runner');
+const createWxtRunnerMock = vi.mocked(createWxtRunner);
+
 describe('createExtensionRunner', () => {
+  const originalDisplay = process.env.DISPLAY;
+
+  afterEach(() => {
+    process.env.DISPLAY = originalDisplay;
+  });
+
   it('should return a Safari runner when browser is "safari"', async () => {
     setFakeWxt({
       config: {
@@ -37,7 +47,8 @@ describe('createExtensionRunner', () => {
     await expect(createExtensionRunner()).resolves.toBe(safariRunner);
   });
 
-  it('should return a WSL runner when `is-wsl` is true', async () => {
+  it('should return a WSL runner when `is-wsl` is true and DISPLAY is not :0', async () => {
+    process.env.DISPLAY = ':1';
     isWslMock.mockResolvedValueOnce(true);
     setFakeWxt({
       config: {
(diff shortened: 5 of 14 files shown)


Seismix:
`packages/runner/src/install.ts` uses the using keyword (Explicit Resource Management):

https://github.com/wxt-dev/wxt/blob/83decf7bf5a584436f73c8f8234f9f620d64063b/packages/runner/src/install.ts#L14

During development these build failures weren't happening because of a Node version mismatch between my env and the CI env.
CI runs Node 22 which doesn't support this without the `--js-explicit-resource-management` flag. Node 24 has it enabled by default, which is why it worked locally for me.

As far as I can tell, previous builds had `@wxt-dev/runner` cached. My changes to `browser-paths.ts` invalidated that cache, so now it actually tries to rebuild and fails.

Can I get input on this? I marked the PR as draft in the meantime.

PatrykKuniczak:
@Seismix Let's rebase in the near future because we'll have changes aroud `runner` in #2079
This can help you or you'll need to change your approach.

Anyway `node 24` should be supported by us, because it was the newest LTS version.
If you need to use v24, feel free to open PR, where you change necessary version from 18 to 24, here: https://github.com/wxt-dev/wxt/blob/1fe073004bb863fe51c2d9565cc59400fac88151/package.json#L4-L6

I hope @aklinker1 won't have any arguments to block this change, but we'll see 😆

aklinker1:
This has gone way beyond the amount of work I thought was required based off the issue discussion...

Let's pause for a second, does `web-ext` support WSL? If so, just remove this if-statement:

https://github.com/wxt-dev/wxt/blob/75f01db54365ff3dee5ebe82211ba29e82fc6053/packages/wxt/src/core/runners/index.ts#L12

If not, add support for WSL over there: https://github.com/mozilla/web-ext

WXT relies on `web-ext` to open the browser, seems like this work should go there, not in WXT.

> Ignore the `@wxt-dev/runner` package for now, it's not ready for use.

Seismix:
You're right @aklinker1 , `web-ext` does partially support WSL. After testing on WSL2 + WSLg:

- **Firefox**: Works out of the box via `web-ext`. `wxt dev -b firefox` launches Firefox and installs the extension with no issues.
- **Chrome**: `web-ext run -t chromium` by itself does launch Chrome and load the extension, but it still creates a malformed UNC profile path (`\\wsl.localhost\...`) due to  `chrome-launcher`. There's no `web-ext` flag to prevent this. The conversion happens unconditionally inside `chrome-launcher`. When used through WXT's programmatic API (`web-ext-run`), this causes an `ECONNRESET` crash on the CDP pipe.

The work on the `wsl-runner` branch was trying to work around the following bugs. But it makes more sense to fix them at the source.

### Investigation
There are two upstream bugs at play:

1. `chrome-launcher` unconditionally converts `--user-data-dir` to a Windows UNC path (`\\wsl.localhost\...`) when it detects   WSL:
  https://github.com/GoogleChrome/chrome-launcher/blob/fe7654fbfd1ed651be9d576b6b46592b70e5b693/src/chrome-launcher.ts#L198
  Even when Chrome is running as a native Linux process under WSLg. This causes Chrome's remote debugging pipe to reset. Known upstream:
  -  https://github.com/GoogleChrome/chrome-launcher/issues/334
  -  https://github.com/GoogleChrome/chrome-launcher/issues/250
  -  https://github.com/GoogleChrome/chrome-launcher/issues/229
  -  https://github.com/GoogleChrome/chrome-launcher/issues/195

2. `web-ext-run` doesn't attach an error handler to the outgoing CDP pipe https://github.com/aklinker1/web-ext-run/blob/master/src/extension-runners/chromium.js#L67, compared with the incoming pipe which has error handling at https://github.com/aklinker1/web-ext-run/blob/master/src/extension-runners/chromium.js#L57. When the pipe resets, it becomes an unhandled `'error'` event that crashes the Node.js process.

### Possible solutions

1. **Minimal: Remove the WSL runner, warn for Chromium**: Delete the old WSL runner (which was a no-op for all browsers anyway), let Firefox go through `web-ext` normally, and show a warning + manual fallback for Chromium in WSL until the upstream bugs are fixed. Possibly drop the `is-wsl` dependency (use an inline `/proc/version` check instead).

1. **Patch dependencies via pnpm**: Apply `pnpm patch` to both `chrome-launcher` (skip `toWin32Path()` when a GUI display is available) and `web-ext-run` (add error handler on the outgoing CDP pipe). This would make Chrome work in WSL immediately without waiting for upstream fixes:

   possible `chrome-launcher` fix:

   ```js
   const hasDisplay = process.env.DISPLAY || process.env.WAYLAND_DISPLAY;
   flags.push(
     `--user-data-dir=${isWsl && !hasDisplay ? toWin32Path(this.userDataDir) : this.userDataDir}`,
   );
   ```

   possible web-ext-run fix:

   ```js
   outgoing.on('error', error => {
     log.error(error);
     this.#finalizeDisconnect();
   });
   outgoing.on('close', () => this.#finalizeDisconnect());

   ```

1. Fix upstream + option 1 as interim: Submit PRs to `chrome-launcher` and `web-ext-run`, use option 1 in WXT as a temporary measure, and remove the WSL check entirely once the upstream fixes land.

1. Combinations?

 ### Branches

I've created branches for demonstration purposes:
- Option 1: https://github.com/wxt-dev/wxt/compare/main...Seismix:wxt:wsl/remove-runner : minimal change, warns for Chromium, Firefox works
- Option 2: https://github.com/wxt-dev/wxt/compare/main...Seismix:wxt:wsl/patch-deps : patches both deps + wrapper script detection, both browsers work

### What next

How should I move forward?

aklinker1:
I like option 3 - interim solution + upstream support.

PatrykKuniczak:
@Seismix Are you want to go on?

PatrykKuniczak:
@Seismix One more poke :)

Seismix:
Hey @PatrykKuniczak, I will pick this up again. Busy times, sorry for the break.

PatrykKuniczak:
@Seismix No problem, take your time.
I only need to know, if you're interested on continue working on this :)

Issue #2078, opened 2026-02-07
Make the runner dependencies (`web-ext`, `web-ext-run`, `@wxt-dev/runner`) peer


aklinker1:
### Feature Request

It's really annoying to manage the `web-ext-run` fork, leading to out-of-date error and bugs already fixed in the upstream `web-ext` repo.

I created the fork to reduce dependencies and WXT's install size in `node_modules`, but if these are peer dependencies, it solves several problems:

1. Smaller install size from WXT (_technically_)
2. Projects control and manage the runner version, meaning they can get updates sooner
3. WXT can look at the installed packages and easily switch between runners for each package

#### Is your feature request related to a bug?

Related to #1700.

### What are the alternatives?

Continue shipping them as a sub-dependency.

### Additional context

N/A

PR #1700, opened 2025-06-02
feat!: Replace `web-ext` with `@wxt-dev/runner`

aklinker1:
### Overview

`web-ext` has few problems:

1. ~20MB install footprint (20% of WXT's total install footprint, even after using my lighter fork, `web-ext-run`)
2. ~~Relies on `--load-extension`, which has been deprecated ([there are plans to upgrade to using the CDP](https://github.com/mozilla/web-ext/issues/3388)~~
3. A little slower (see `./scripts/benchmarks/browser-startup.sh`)
   | Package | Time to open browser (mean, N=20) |
   | --- | :---: | 
   | `web-ext` | 634ms | 
   | `@wxt-dev/runner` | 594ms | 

### TODO

- [ ] Can you connect a remote debugger to it?

### Manual Testing

```ts
cd pacakges/wxt-demo
pnpm dev
```

This should open your browser.

### Related Issue

N/A

PR #2079, opened 2026-02-07
fix!: Make `web-ext` a peer dependency

aklinker1:
### Overview

Instead of hard switching to `@wxt-dev/runner` as in #1700, first make `web-ext` a peer dependency. Then it will be easy to add support for `@wxt-dev/runner` as a peer similar to `web-ext` without a breaking change, and give devs an easy path to migrate if they want.

- [x] See updated docs for breaking changes.
- [x] I updated the templates so new projects continue to open the browser during development by default.
- [x] Move runner construction to `resolveConfig`, allowing modules to override the default and provide their own runner. Follows standards for construction of other objects on the `wxt` instance, like `logger`
- [x] Update tests

### Manual Testing

```sh
cd packages/wxt-demo

# Make sure the browser doesn't open automatically
pnpm rm web-ext
pnpm dev

# Make sure the browser does open automatically
pnpm add web-ext
pnpm dev
```

### Related Issue

This PR closes https://github.com/wxt-dev/wxt/issues/2078

### DIFF
--- a/packages/wxt/e2e/tests/runners.test.ts
+++ b/packages/wxt/e2e/tests/runners.test.ts
@@ -9,8 +9,8 @@ import { wxt } from '../../src/core/wxt';
 
 // Globals for modifying mock behaviors
 
-var isWsl = false;
-var importWebExtRunnerError: Error | undefined = undefined;
+let isWsl = false;
+let importWebExtRunnerError: Error | undefined = undefined;
 
 // Mock runners to create constants for checking equality
 


aklinker1:
@PatrykKuniczak I'll add you as a reviewer once this is ready so you can see one of my PRs. Also if you have any feedback, please share!

packages/wxt/src/core/resolve-config.ts
```diff
@@ -217,7 +222,9 @@ export async function resolveConfig(
     publicDir,
     wxtModuleDir,
     root,
-    runnerConfig,
+    runnerConfig: webExt,
+    webExt,
+    runner: await resolveRunner(browser, logger, mergedConfig),
```
aklinker1: Hmm, I should return a manual runner for build mode, no need to spend time trying to import `web-ext` in that case.


```suggestion
    runner: command === "dev"
      ? await resolveRunner(browser, logger, mergedConfig)
      : createManualRunner(),
```

packages/wxt/e2e/tests/runners.test.ts
```diff
@@ -0,0 +1,135 @@
...
+import { createManualRunner } from '../../src/core/runners/manual';
+import { createWebExtRunner } from '../../src/core/runners/web-ext';
+import { createWslRunner } from '../../src/core/runners/wsl';
+import { TestProject } from '../utils';
+import { wxt } from '../../src/core/wxt';
+
+// Globals for modifying mock behaviors
+
+var isWsl = false;
+var isWebExtAvailable = true;
```
PatrykKuniczak: Why `var` not `let`?

packages/wxt/e2e/tests/runners.test.ts
```diff
@@ -0,0 +1,135 @@
...
+  return { createManualRunner: () => runner };
+});
+const manualRunner = createManualRunner();
+
+vi.mock('../../src/core/runners/web-ext', () => {
+  const runner = createMockExtensionRunner('web-ext');
+  return {
+    createWebExtRunner: () => {
+      if (isWebExtAvailable) return runner;
+      else throw Error('Module not found: web-ext'); // TODO: make this realistic
```
PatrykKuniczak: fix this todo, before merge

packages/wxt/e2e/tests/runners.test.ts
```diff
@@ -0,0 +1,135 @@
...
+        });
+      });
+    });
+
+    describe('web-ext is not installed', () => {
+      beforeEach(() => {
+        isWebExtAvailable = false;
+      });
+
+      it('should use the web-ext runner', async () => {
```
PatrykKuniczak:
```suggestion
      it('should use the manual runner', async () => {
```

packages/wxt/src/core/resolve-config.ts
```diff
@@ -651,3 +661,26 @@ export async function resolveWxtUserModules(
...
+
+  try {
+    // This module imports `web-ext`, so if it fails, we know `web-ext` isn't installed
+    const { createWebExtRunner } = await import('./runners/web-ext');
+    return mergedConfig.webExt?.disabled
+      ? createManualRunner()
+      : createWebExtRunner();
+  } catch (err) {
+    logger.debug('Error loading the web-ext runner', err);
+    // TODO: only ignore "web-ext module not found" errors
```
PatrykKuniczak: fix this, before merge

packages/wxt/e2e/tests/runners.test.ts
aklinker1: Vitest hoists the mocks to the top of the file before the variables are defined. Using var ensures they exist even if the mocks are hoisted, make make them usable inside the mock callbacks without a "isWsl is not defined" error. At least that was the idea, I didn't test with just `let`, I'll give it a go.

aklinker1: Works with `let`, was being too fancy lol

PR #2104, opened 2026-02-09
chore: simplify imports of `cli/__tests__/index.ts`

PatrykKuniczak:
### Overview

I've simplified imports

### DIFF
--- a/packages/wxt/src/cli/__tests__/index.test.ts
+++ b/packages/wxt/src/cli/__tests__/index.test.ts
@@ -1,10 +1,12 @@
 import { describe, it, vi, beforeEach, expect } from 'vitest';
-import { build } from '../../core/build';
-import { createServer } from '../../core/create-server';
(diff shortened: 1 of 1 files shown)


aklinker1 (approved):

PatrykKuniczak:
@aklinker1 This should be `refactor` or `chore`?

aklinker1:
Either is fine, I almost always go with chore so I don't have to make that decision lol.

I usually just stick to `fix`, `feat`, `chore`, and `perf`, but I don't care if other people use others.

PatrykKuniczak:
@aklinker1 It won't affect releease notes?

aklinker1:
Oh and I use `docs`.

> It won't affect releease notes?

It does effect release notes and version bumping, it decides what section they go under and if the version bump will be major/minor/patch. CI will tell you if you've chosen something that's invalid, we have commitlint setup for that.

PR #2105, opened 2026-02-09
docs: rename keys name of getMetas() to be proper

PatrykKuniczak:
### Overview

I've renamed keys in jsdocs, to have the same names as func has indeed

### DIFF
--- a/packages/storage/src/index.ts
+++ b/packages/storage/src/index.ts
@@ -708,7 +708,7 @@ export interface WxtStorage {
   /**
    * Get the metadata of multiple storage items.
    *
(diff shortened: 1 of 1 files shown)


aklinker1:
Now this one is only touching docs, so it should be `docs`, not `fix`.

`feat` and `fix` are the only ones that will result in a version bump. Maybe you could argue that this could be a `fix`, that'd be fine with me as well 🤷 

Not a big deal as long as you're careful using `feat` and breaking changes.

PatrykKuniczak:
@aklinker1 Let it be `docs` then :)

PR #2107, opened 2026-02-09
chore: add missing `async` to network.ts

PatrykKuniczak:


aklinker1 (changes requested):
This function doesn't `await` anything, so it doesn't need to be async. If you want to refactor the rest of the function to use try/catch and `await`, then the function needs to be `async`. But it's fine for now.

My editor also recommended this change, but I didn't add it for this reason in my async/await cleanup PR.

PatrykKuniczak:
@aklinker1 It'll be ok, if i'm add tests for `network` in new PR, and there only try/catch?
Or tests isn't necessary?

aklinker1:
>  It'll be ok, if i'm add tests for network in new PR, and there only try/catch?
> Or tests isn't necessary?

Sorry, I don't understand your question

PatrykKuniczak:
@aklinker1 I want to add try/catch in this PR.
And in other tests for entire `network.ts`(If this could be helpful, if not, don't mind)

packages/wxt/src/core/utils/network.ts
```diff
@@ -2,11 +2,16 @@ import dns from 'node:dns';
...
-  });
-  return withTimeout(isOffline, 1e3).catch(() => true);
+async function isOffline(): Promise<boolean> {
+  try {
+    const isOffline = new Promise<boolean>((res) => {
+      dns.resolve('google.com', (err) => res(err != null));
+    });
+    return await withTimeout(isOffline, 1e3);
+  } catch (error) {
+    console.error('Error checking offline status:', error);
```
aklinker1: Use `consola` if you want to log something but can't use `wxt.logger`


```suggestion
    consola.error('Error checking offline status:', error);
```

PatrykKuniczak:
@aklinker1 OH yeah

PR #2109, opened 2026-02-09
fix: Remove `prepublish` script of root `package.json`

PatrykKuniczak:
### Overview

I've removed this script


PatrykKuniczak:
> -s does exist, it makes pnpm silent so it doesn't eche the command it's about to run.
It doesn't work on my machine:
image

I'll remove this script

PR #2111, opened 2026-02-11
fix(zip): make includeSources use allowlist behavior

### LINKED ISSUE #2059: zip.includeSources does not behave intuitively, can lead to data leak
### Setup

```ts
// wxt.config.ts
export default defineConfig({
  zip: {
    includeSources: ['entrypoints/**'],
  },
})
```

Repo tree:

```
entrypoints/foo.ts
entrypoints/bar.js
cache/secrets/abc
super-secret-file
.gitignore
```

### Expected behavior

```shell
$ wxt zip:firefox
$ zipinfo .output/*sources.zip
entrypoints/foo.ts
entrypoints/bar.js
```

### Current non-intuitive, dangerously leaky behavior

```shell
$ wxt zip:firefox
$ zipinfo .output/*sources.zip
entrypoints/foo.ts
entrypoints/bar.js
cache/secrets/abc
super-secret-file
```

### Hacky workaroud

 I should not have to do this, but that's how I worked around this bug:

```ts
export default defineConfig({
  zip: {
    excludeSources: ['**'],
    includeSources: ['entrypoints/**'],
  },
})
```

I'll point out that this usually not how an allowlist/denylist behaves. The "accepted", safe, intuitive behavior is the following:

* Nothing configured: default behavior, today "*Hidden files, node_modules, and tests are ignored*". I'd personally vote for a more advanced "respect .Xignore" support for a bunch of popular VCSes, like what ripgrep does, but that's a stretch.
* Only allowlist provided (include): takes precedence since the user went through the trouble of customizing this setting; only what's "included" is included; **not what happens, and what this bug is about**.
* Only denlist provided (exclude): takes precedence, everything is included *except* what's denied.
* Both allowlist & denylist provided: only what passes the allowlist, then filter out further according to denylist. Though I accept that the opposite (first denylist, then force-allowlist), which is  WXT's current behavior ("*[includeSources] overrides excludeSources; if a file matches both lists, it is included in the ZIP*") is acceptable albeit annoying.

Pointers:
* https://wxt.dev/guide/essentials/publishing.html#firefox-addon-store
* https://wxt.dev/api/reference/wxt/interfaces/InlineConfig.html#zip
* Vaguely related: #1505

### Reproduction

[repro.zip](https://github.com/user-attachments/files/24983602/repro.zip)

### Steps to reproduce

```shell
$ mkdir /tmp/repro && cd /tmp/repro
$ unzip /tmp/repro.zip
$ npm install
$ npm run zip:firefox
$ zipinfo .output/*sources.zip
```

### System Info

```shell
System:
    OS: Linux 6.18 cpe:/o:nixos:nixos:26.05 26.05 (Yarara)
    CPU: irrelevant
    Memory: irrelevant
    Container: Yes
    Shell: 5.3.9 - /bin/bash
  Binaries:
    Node: 24.13.0 - node
    Yarn: 1.22.22 - yarn
    npm: 11.6.2 - npm
    pnpm: 10.28.0 - pnpm
  npmPackages:
    wxt: ^0.20.6 => 0.20.13
```

### Used Package Manager

pnpm

### Validations

kaigritun:
## Summary

When `includeSources` patterns are provided, only files matching those patterns should be included in the sources ZIP. Previously, `includeSources` patterns were added to the default `**/*` glob, which could leak sensitive files that weren't explicitly excluded.

## Problem

As described in #2059, the current behavior is unintuitive and can lead to data leaks:

```ts
// wxt.config.ts
export default defineConfig({
  zip: {
    includeSources: ['entrypoints/**'],
  },
})
```

**Expected:** Only `entrypoints/**` files are included
**Actual:** ALL files are included, plus the entrypoints patterns override excludeSources

This means sensitive files like `secrets/api-key.txt` or `cache/credentials.json` could accidentally be included in the sources ZIP submitted to Firefox.

## Solution

This change makes `includeSources` behave as an allowlist:
- When `includeSources` is provided: only matching files are included
- When `includeSources` is not provided: all files are included (current behavior)
- `excludeSources` still applies on top of include patterns

## Changes

1. **`packages/wxt/src/core/zip.ts`**: When `include` patterns are provided, use only those patterns for globbing instead of combining with `**/*`
2. **`packages/wxt/src/types.ts`**: Updated documentation to clarify the allowlist behavior
3. **`packages/wxt/e2e/tests/zip.test.ts`**: Added test to verify non-included files are excluded

## Breaking Change

If you were using `includeSources` to add specific files to the default set (e.g., hidden files), you now need to include all files you want in the ZIP:

```ts
// Before (old behavior: hidden files added to default)
zip: {
  includeSources: ['.env'],
}

// After (new behavior: explicit allowlist)
zip: {
  includeSources: ['**/*', '.env'],
}
```

Fixes #2059

### DIFF
--- a/packages/wxt/src/core/zip.ts
+++ b/packages/wxt/src/core/zip.ts
@@ -113,18 +113,20 @@ async function zipDir(
   },
 ): Promise<void> {
   const archive = new JSZip();
+  // When include patterns are provided, use only those patterns (allowlist behavior)
+  // When no include patterns are provided, use **/* as the default (include everything)
+  const globPatterns =
+    options?.include && options.include.length > 0 ? options.include : ['**/*'];
   const files = (
-    await glob(['**/*', ...(options?.include || [])], {
+    await glob(globPatterns, {
       cwd: directory,
       // Ignore node_modules, otherwise this glob step takes forever
       ignore: ['**/node_modules'],
       onlyFiles: true,
     })
   ).filter((relativePath) => {
-    return (
-      minimatchMultiple(relativePath, options?.include) ||
-      !minimatchMultiple(relativePath, options?.exclude)
-    );
+    // Exclude files that match any exclude pattern
+    return !minimatchMultiple(relativePath, options?.exclude);
   });
   const filesToZip = [
     ...files,
--- a/packages/wxt/e2e/tests/zip.test.ts
+++ b/packages/wxt/e2e/tests/zip.test.ts
@@ -181,6 +181,42 @@ describe('Zipping', () => {
     );
   });
 
+  it('should only include specified files when includeSources is provided (allowlist behavior)', async () => {
+    const project = new TestProject({
+      name: 'test',
+      version: '1.0.0',
+    });
+    project.addFile(
+      'entrypoints/background.ts',
+      'export default defineBackground(() => {});',
+    );
+    project.addFile('src/utils.ts', 'export const x = 1;');
+    project.addFile('secrets/api-key.txt', 'supersecret');
+    project.addFile('cache/data.json', '{}');
+    const unzipDir = project.resolvePath('.output/test-1.0.0-sources');
+    const sourcesZip = project.resolvePath('.output/test-1.0.0-sources.zip');
+
+    await project.zip({
+      browser: 'firefox',
+      zip: {
+        includeSources: ['entrypoints/**', 'src/**'],
+      },
+    });
+    await extract(sourcesZip, { dir: unzipDir });
+
+    // Included files should be present
+    expect(
+      await project.pathExists(unzipDir, 'entrypoints/background.ts'),
+    ).toBe(true);
+    expect(await project.pathExists(unzipDir, 'src/utils.ts')).toBe(true);
+
+    // Non-included files should NOT be present (allowlist behavior)
+    expect(await project.pathExists(unzipDir, 'secrets/api-key.txt')).toBe(
+      false,
+    );
+    expect(await project.pathExists(unzipDir, 'cache/data.json')).toBe(false);
+  });
+
   it('should exclude skipped entrypoints from respective browser sources zip', async () => {
     const project = new TestProject({
       name: 'test',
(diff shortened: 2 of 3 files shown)


packages/wxt/src/core/zip.ts
```diff
@@ -113,18 +113,20 @@ async function zipDir(
...
       onlyFiles: true,
     })
   ).filter((relativePath) => {
-    return (
-      minimatchMultiple(relativePath, options?.include) ||
-      !minimatchMultiple(relativePath, options?.exclude)
-    );
+    // Exclude files that match any exclude pattern
+    return !minimatchMultiple(relativePath, options?.exclude);
   });
```
aklinker1, line 130: We can simplify this and remove the whole filter, moving the `options?.exclude` globs up to here: https://github.com/wxt-dev/wxt/blob/3417bd3ed27d2e52e6b906393e0329565256f0e7/packages/wxt/src/core/zip.ts#L119-L120

aklinker1 (changes requested):
I've been trying to cleanup random config defaults outside the `resolve-config.ts` file. So there's a couple more changes this PR needs to make:

1. Can you move the `**/*` default into the `resolveZipConfig` function here:

   https://github.com/wxt-dev/wxt/blob/3417bd3ed27d2e52e6b906393e0329565256f0e7/packages/wxt/src/core/resolve-config.ts#L303

   ```ts
   includeSources: mergedConfig.zip.includeSources ?? ['**/*'], 
   ```

2. Similarly, if we move the `excludeSouces` array into the glob's `ignore` option, we can remove the hard-coded `node_modules` glob from here 

   https://github.com/wxt-dev/wxt/blob/3417bd3ed27d2e52e6b906393e0329565256f0e7/packages/wxt/src/core/zip.ts#L120

   because it will already be included. While the zipDir function is called multiple times, the other place it's called doesn't need to ignore the `node_modules` folder since it's zipping the build output.

Finally, this is a breaking change, so the PR needs to be based off the `major` branch instead of `main`, and the PR should target `major` as well.

kaigritun:
Superseded by #2114 which targets the `major` branch and incorporates the maintainer's requested changes.

PR #2112, opened 2026-02-11
fix: file reloading mechanism and add profile watch ignores

ZerGo0:
### Overview

Fixes flaky dev reload detection where WXT sometimes ignores real source file saves.

Root cause:
- Dev watcher used a debounced callback (`leading: true`, `trailing: false`) and queued events inside that callback.
- With noisy filesystem activity (especially browser profile/cache files), relevant source change events could be dropped.

Changes:
- Reworked dev file reload flow in `packages/wxt/src/core/create-server.ts` to:
  - queue every watcher event immediately
  - process the queue in a serialized loop
  - keep flushing after each debounce window until no new events arrive
- Added runner profile watch ignores in `packages/wxt/src/core/builders/vite/index.ts` so browser profile churn is not watched:
  - `chromiumProfile`
  - `firefoxProfile`
  - `chromiumArgs` `--user-data-dir` (split and `=` forms)
  - `firefoxArgs` `-profile` (split and `=` forms)
- Added a single regression test:
  - `packages/wxt/src/core/__tests__/create-server.test.ts`
  - Simulates browser-profile noise event followed by real source change and verifies the source change is detected/reloaded.

### Manual Testing

This is pretty hard/annoying to reproduce. The new test is probably the best way of reproducing this: `pnpm -C packages/wxt test --run src/core/__tests__/create-server.test.ts`

1. Use a project with browser profile inside project dir (example):
   - Chromium: `chromiumArgs: ['--user-data-dir=./private/.dev-profile']`
   - Firefox: `firefoxProfile: './private/.dev-firefox-profile'`
   - My main project does a ton of stuff in indexed dbs, this causes a ton of changes in the browser profile directory
2. Run `pnpm wxt`.
3. Save a real source file while browser is active.
4. A rebuild should be triggered, but it just doesn't sometimes.

### Related Issue

None I think.


aklinker1 (commented):
Nice, I really like these changes!

I just woke up and am on mobile, so I'll need to do some manual testing to make sure everything is working later today.

ZerGo0:
I pushed another small change. I don't really like the changes in https://github.com/wxt-dev/wxt/pull/2112/commits/0cf79bb42a9ac731702abf8d6c1b75e590f486bc but I can't really think of a better way of doing this. 

The problem is that it would show potentially wrong files in the "Changed: ..." log line and so on because they all use the "raw"/unfiltered queue. My latest commit fixes that.

PatrykKuniczak:
I'm also have 1 file on which HMR doesn't work and i was about to report it, but i'll check it on your branch, maybe your PR will solve that issue 😄

packages/wxt/src/core/builders/vite/index.ts
```diff
@@ -371,6 +375,59 @@ export async function createViteBuilder(
...
+        .map((profilePath) => `${profilePath}/**`),
+    ),
+  );
+}
+
+function extractPathArgs(args: string[] | undefined, flag: string): string[] {
+  if (!args?.length) return [];
+
+  const paths: string[] = [];
+  for (let i = 0; i < args.length; i++) {
```
PatrykKuniczak, line 442: Maybe consider `.forEach()`

ZerGo0: Needs to be like this because of L422

PatrykKuniczak: @ZerGo0 You can use index also in `.forEach()`, if you want `nextValue`.

ZerGo0: I can do that if you guys want, but that would iterate over the path 2x because I can't increase the index in foreach.

PatrykKuniczak: @ZerGo0 Yeah, that's a problem yeah.
Let's leave it what it is right now, i've missed that:
in L424:
```
      i += 1;
```

We don't want perf issue, because of a little more pretty code.

Don't mind about this change.

packages/wxt/src/core/create-server.ts
```diff
@@ -202,24 +202,31 @@ async function createServerInternal(): Promise<WxtDevServer> {
  * Returns a function responsible for reloading different parts of the extension when a file
  * changes.
  */
-function createFileReloader(server: WxtDevServer) {
+export function createFileReloader(server: WxtDevServer) {
```
aklinker1: Exporting this here will export the function from the main `wxt` module as a public API, but I don't want this to be public. Move to another file, maybe in `src/core/utils/create-file-reloader.ts`, if you want to export and test it.

https://github.com/wxt-dev/wxt/blob/50d7c1ec804576dbbaeac68b68c6d4b9ea7214f4/packages/wxt/src/core/index.ts#L5

https://github.com/wxt-dev/wxt/blob/50d7c1ec804576dbbaeac68b68c6d4b9ea7214f4/packages/wxt/src/index.ts#L9

aklinker1 (changes requested):
Tested, things seem to behave the same way for me. One requested change.

packages/wxt/src/core/create-server.ts
ZerGo0: Done, had to move a bit more than just `createFileReloader`.

PatrykKuniczak:
it's working for Windows also, but it doesn't solve my issue, i'll open new issue about it.

packages/wxt/src/core/utils/create-file-reloader.ts
```diff
@@ -0,0 +1,198 @@
...
+          .map((file) => pc.dim(relative(wxt.config.root, file)))
+          .join(', ')}`,
+      );
+
+      // Rebuild entrypoints on change
+      const allEntrypoints = await findEntrypoints();
+      try {
+        const { output: newOutput } = await rebuild(
+          allEntrypoints,
+          // TODO: this excludes new entrypoints, so they're not built until the dev command is restarted
```
PatrykKuniczak: Let's fix it or remove if it's outdated

PatrykKuniczak (changes requested):

packages/wxt/src/core/utils/create-file-reloader.ts
ZerGo0: Done

packages/wxt/src/core/utils/create-file-reloader.ts
```diff
@@ -0,0 +1,303 @@
...
+          rebuildGroups,
+          changes.type === 'no-change'
+            ? server.currentOutput
+            : changes.cachedOutput,
+        );
+        server.currentOutput = newOutput;
+
+        // Perform reloads
+        const needsFullExtensionReload =
+          newEntrypointGroups.length > 0 ||
```
PatrykKuniczak:
```suggestion
          newEntrypointGroups.length ||
```

It'll the same result 😄

PatrykKuniczak (approved):

PatrykKuniczak:
@aklinker1 This one also, needs your review

packages/wxt/src/core/builders/vite/index.ts
```diff
@@ -382,6 +386,59 @@ export async function createViteBuilder(
   };
 }
 
+export function getRunnerProfileWatchIgnores(
+  wxtConfig: ResolvedConfig,
+): string[] {
+  const root = normalizePath(wxtConfig.root);
+  const chromiumArgProfiles = extractPathArgs(
+    wxtConfig.runnerConfig.config?.chromiumArgs,
```
aklinker1, line 413: Hmm, I'm working on adding an alternative runner in the next major version and hardcoding access to one of the runner's config will prevent me from abstracting it.

It works for now like this, but we'll likely have to add a hook for this in the future...

Just thinking out loud, no changes here.

packages/wxt/src/core/builders/vite/index.ts
aklinker1: @PatrykKuniczak `for` loops are always faster than `.forEach()` loops. We should always use either `for (let i; ...)` or `for (... each ...)`.

packages/wxt/src/core/create-server.ts
```diff
@@ -1,35 +1,15 @@
-import { debounce } from 'perfect-debounce';
```
aklinker1: Can we remove this dependency?

packages/wxt/src/core/utils/create-file-reloader.ts
aklinker1, line 1: I'm going to move over the original code into it's own file in a different PR. That way I can see what your changes actually are here.

There's several conflicts this is also ignoring.

packages/wxt/src/core/utils/__tests__/create-file-reloader.test.ts
aklinker1, line 1: Should this be a test for `create-file-reloader.ts`? The current filename and directory is wrong.

ZerGo0:
@aklinker1 I can do a clean PR if you want?

(This issue is super frustrating tbh, 1/20 file saves trigger a rebuild/reload for me right now for me. It's usually faster to just stop and start the dev command ^^)

aklinker1:
Yeah, resolve the conflicts and I'll finish reviewing it.

packages/wxt/src/core/utils/__tests__/create-file-reloader.test.ts
```diff
@@ -0,0 +1,145 @@
...
+    expect(
+      allEntrypoints.some((entry) => entry.inputPath === relevantFile),
+    ).toBe(true);
+    expect(
+      rebuiltGroups.flat().some((entry) => entry.inputPath === relevantFile),
+    ).toBe(true);
+    expect(server.reloadExtension).toBeCalledTimes(1);
+  });
+
+  it('should rebuild and reload extension when a new entrypoint is added', async () => {
```
aklinker1, line 93: I noticed you resolved that TODO, thanks!

packages/wxt/src/core/utils/create-file-reloader.ts
```diff
@@ -48,46 +82,62 @@ export function createFileReloader(server: WxtDevServer) {
...
-              .map((entry) => entry.name);
-            wxt.logger.success(`Reloaded: ${getFilenameList(rebuiltNames)}`);
-            break;
+        const needsFullExtensionReload =
+          newEntrypointGroups.length ||
+          changes.type === 'extension-reload' ||
+          changes.type === 'no-change';
+        if (needsFullExtensionReload) {
+          server.reloadExtension();
+          wxt.logger.success(`Reloaded extension`);
```
aklinker1: Alright, changes look good. This is the only thing that stood out to me: why reload the extension when no changes were detected?

aklinker1 (changes requested):

packages/wxt/src/core/utils/create-file-reloader.ts
ZerGo0: I'm pretty sure that I added it as a temp fix for new entry points, but I removed it now since it's not needed anymore.

aklinker1: Now it goes through the else statement and reloads content scripts when there are no changes.

aklinker1 (changes requested):
Some type errors causing CI to fail, the no-reload case is still not handled properly, it runs the final `else` block and tries to reload content scripts

aklinker1 (approved):
Sorry for the delay on this, merging now!

Issue #604, opened 2024-04-06
Dotfiles cannot be included in the sources for Firefox


hikiko4ern:
### Describe the bug

Dotfiles cannot be included in a `sources.zip` using `includeSources` (added in #378) after the changes made in #501: `**/*` passed to `fast-glob` doesn't match dotfiles without `dot: true`, so `includeSources` doesn't actually affect anything
https://github.com/wxt-dev/wxt/blob/6b72f883b6c9407804240efe5b10a8967f62c753/src/core/zip.ts#L102

### To Reproduce

[wxt-zip-env.zip](https://github.com/wxt-dev/wxt/files/14894952/wxt-zip-env.zip)

1. Install dependencies: `pnpm i`
2. Create sources zip: `pnpm zip:firefox`
3. Check if `.env` file is included in `.output/wxt-starter-0.0.0-sources.zip`

If you downgrade `wxt` to `0.17.4`, `.env` file will be included in `sources.zip`

### Expected behavior

The dotfiles specified in `includeSources` must be included in `sources.zip`

### Environment

```
System:
  OS: Linux 5.15 Debian GNU/Linux trixie/sid
  CPU: (32) x64 AMD Ryzen 9 7950X3D 16-Core Processor
  Memory: 56.33 GB / 64.76 GB
  Container: Yes
  Shell: 5.9 - /bin/zsh
Binaries:
  Node: 20.12.1 - /mnt/wslg/runtime-dir/fnm_multishells/32306_1712432100305/bin/node
  Yarn: 4.1.1 - /mnt/wslg/runtime-dir/fnm_multishells/32306_1712432100305/bin/yarn
  npm: 10.5.0 - /mnt/wslg/runtime-dir/fnm_multishells/32306_1712432100305/bin/npm
  pnpm: 8.15.6 - /mnt/wslg/runtime-dir/fnm_multishells/32306_1712432100305/bin/pnpm
  bun: 1.1.2 - ~/.bun/bin/bun
Browsers:
  Chrome: 123.0.6312.86
  Chromium: 122.0.6261.57
npmPackages:
  wxt: ^0.17.12 => 0.17.12
```

### Additional context

Since `includeSources` are already globs, perhaps it is enough to just add them to the `glob` sources?

```ts
await glob(['**/*', ...(options?.include || [])], {
```
However, I'm not sure why `options.include` is passed to `zipDir`
[...]

PR #2114, opened 2026-02-11
fix(zip)!: make `includeSources`/`excludeSources` follow standard allowlist behavior

### LINKED ISSUE #2059: zip.includeSources does not behave intuitively, can lead to data leak

kaigritun:
When includeSources patterns are provided, only files matching those patterns should be included in the sources ZIP. Previously, includeSources patterns were added to the default `**/*` glob, which could leak sensitive files that weren't explicitly excluded.

**Changes per maintainer review (replaces #2111):**
- Move default `['**/*']` into `resolveZipConfig`
- Pass excludeSources directly to glob's `ignore` option
- Remove hardcoded `node_modules` glob (already in excludeSources default)
- Simplify `zipDir` logic
- Adds a `dotSources` option to allow configuring `tinyglobby`'s `dot` option, allowing for zipping hidden files

**BREAKING CHANGE:** Sources zip generation behavior has changed.

Fixes #2059
Supersedes #2111


packages/wxt/src/core/zip.ts
```diff
@@ -10,7 +10,6 @@ import { registerWxt, wxt } from './wxt';
 import JSZip from 'jszip';
 import glob from 'fast-glob';
 import { normalizePath } from './utils/paths';
-import { minimatchMultiple } from './utils/minimatch-multiple';
```
aklinker1: We can delete this file and it's test completely, this was the only place they were used.

aklinker1 (changes requested):
While reviewing, I remembered some past issues  related to zipping sources that we need to support:

- https://github.com/wxt-dev/wxt/issues/738
- https://github.com/wxt-dev/wxt/pull/902
- https://github.com/wxt-dev/wxt/discussions/373
- https://github.com/wxt-dev/wxt/issues/604

TLDR: We need to support including hidden files and files in hidden directories. I think this requirement was why the logic was so complex.

Right now, I don't think this PR supports adding hidden files at all.

I ran some tests to refresh my knowledge of `fast-glob`, verifying the three cases we care about (regular files, hidden files, regular files in hidden directories):

<details>

```ts
import { glob, Options } from "fast-glob";
import fs from "node:fs/promises";
import { inspect } from "node:util";

const test = async (sources: string[], options: Options) => {
  const res = await glob(sources, options);
  console.log(
    `
───

glob(
  ${inspect(sources, { colors: true })},
  ${inspect(options, { colors: true })}
)
= ${inspect(res, { colors: true })}`,
  );
};

const allFiles = await fs.readdir(".", {
  recursive: true,
});
console.log("All files (minus node_modules):");
console.log(
  allFiles.filter((file: string) => !file.startsWith("node_modules")),
);

await test(["**/*"], { ignore: ["**/node_modules/**"] });
await test(["**/.*"], { ignore: ["**/node_modules/**"] });
await test(["**/.*/**/*"], { ignore: ["**/node_modules/**"] });
```

and the output:

```
All files (minus node_modules):
[ ".wxt", "bun.lock", "package.json", ".env", "index.ts", ".wxt/test.txt" ]

───

glob(
  [ '**/*' ],
  { ignore: [ '**/node_modules/**' ] }
)
= [ 'bun.lock', 'package.json', 'index.ts' ]

───

glob(
  [ '**/.*' ],
  { ignore: [ '**/node_modules/**' ] }
)
= [ '.env' ]

───

glob(
  [ '**/.*/**/*' ],
  { ignore: [ '**/node_modules/**' ] }
)
= []
```

</details>

But it seems like `fast-glob` doesn't support including files from inside hidden directories (I did try the `dot: true` option as well, but it didn't make a difference)...

So I tried `glob` instead:

<details>

```ts
import { glob, GlobOptions } from "glob";
import fs from "node:fs/promises";
import { inspect } from "node:util";

const test = async (sources: string[], options: GlobOptions) => {
  const res = await glob(sources, options);
  console.log(
    `
───

glob(
  ${inspect(sources, { colors: true })},
  ${inspect(options, { colors: true })}
)
= ${inspect(res, { colors: true })}`,
  );
};

const allFiles = await fs.readdir(".", {
  recursive: true,
});
console.log("All files (minus node_modules):");
console.log(
  allFiles.filter((file: string) => !file.startsWith("node_modules")),
);

await test(["**/*"], { ignore: ["**/node_modules/**"] });
await test(["**/.*"], { ignore: ["**/node_modules/**"] });
await test(["**/.*/**/*"], { ignore: ["**/node_modules/**"] });
await test(["**/.*/**/*"], { ignore: ["**/node_modules/**"], nodir: true });
```

```
All files (minus node_modules):
[ ".wxt", "bun.lock", "package.json", ".env", "index.ts", ".wxt/test.txt" ]

───

glob(
  [ '**/*' ],
  { ignore: [ '**/node_modules/**' ] }
)
= [ 'index.ts', 'package.json', 'bun.lock' ]

───

glob(
  [ '**/.*' ],
  { ignore: [ '**/node_modules/**' ] }
)
= [ '.env', '.wxt' ]

───

glob(
  [ '**/.*/**/*' ],
  { ignore: [ '**/node_modules/**' ] }
)
= [ '.wxt/test.txt' ]

───

glob(
  [ '**/.*/**/*' ],
  { ignore: [ '**/node_modules/**' ], nodir: true }
)
= []
```

</details>

It works, but I can't get it to return just files. It does support returning with file-types, so we could filter out directories that way.

Unforunate since `glob` is 10x as large...

- https://pkg-size.dev/fast-glob - 515KB / 18 packages
- https://pkg-size.dev/glob - 5MB / 9 packages

Maybe there's another glob library that's lighter-weight and would allow devs to include hidden files and files in hidden directories?

Otherwise I think we have to switch from `fast-glob` to `glob`.

aklinker1:
Can you also update the docs, adding a section about this change to the docs here: https://github.com/wxt-dev/wxt/blob/19c1d0ab872c150f8e565b6147f2adabbcf4f1ab/docs/guide/resources/upgrading.md?plain=1#L66

Maybe wait to do that until all the changes are finalized through.

MichaelDeBoey:
I'm not making any claims about the quality of their work, but I wanted to let you know that @kaigritun is a fully-autonomous non-human actor
https://socket.dev/blog/ai-agent-lands-prs-in-major-oss-projects-targets-maintainers-via-cold-outreach

PatrykKuniczak:
@kaigritun Are you want to go on?

PatrykKuniczak:
@kaigritun One more poke :)

aklinker1 (approved):
I had to make some changes around how hidden files are included, but I'm happy with the result. I imagine for wxt 1.0, we may need to provide the full options for `tinyglobby`, not just `include`, `exclude`, and `dot`. But this is good enough to get feedback on.

PR #2117, opened 2026-02-12
chore: Change browser tests to fix deprecated warning

PatrykKuniczak:
### Overview

I've changed tests a little to fit new method instead of deprecated one

### Manual Testing

Let's see if CI pass

### Related Issue

This is part of #2022

### DIFF
--- a/packages/browser/src/__tests__/index.test.ts
+++ b/packages/browser/src/__tests__/index.test.ts
@@ -11,7 +11,7 @@ describe('browser', () => {
     });
 
     it('should provide values via the browser import', () => {
-      expectTypeOf<typeof browser.runtime.id>().toEqualTypeOf<string>();
+      expectTypeOf(browser.runtime.id).toEqualTypeOf<string>();
       expectTypeOf<
         typeof browser.storage.local
       >().toEqualTypeOf<Browser.storage.LocalStorageArea>();


packages/browser/src/__tests__/index.test.ts
```diff
@@ -5,19 +5,19 @@ import { browser, type Browser } from '../index';
...
 
     it('should provide values via the browser import', () => {
-      expectTypeOf(browser.runtime.id).toMatchTypeOf<string>();
-      expectTypeOf(
-        browser.storage.local,
-      ).toMatchTypeOf<Browser.storage.StorageArea>();
-      expectTypeOf(
-        browser.i18n.detectLanguage('Hello, world!'),
-      ).resolves.toMatchTypeOf<chrome.i18n.LanguageDetectionResult>();
+      expectTypeOf<typeof browser.runtime.id>().toEqualTypeOf<string>();
```
aklinker1:
```suggestion
      expectTypeOf(browser.runtime.id).toEqualTypeOf<string>();
```

These should be equivalent, just requires less code. If you have a value, pass it in as an arg. If you have a type, pass it in as a type param.

See official docs: https://github.com/mmkal/expect-type?tab=readme-ov-file#features

PatrykKuniczak: I've been thinking about it, but i want to follow the same structure for all, but if you want i can commit your suggestion, give reaction to that message :)

aklinker1: I want to use the library as intended, please use my recommendation.

PatrykKuniczak:
Time to sleep, enough for today :)

PR #2118, opened 2026-02-12
MISTAKE

PatrykKuniczak:
### Overview

I've created a couple new types, instead of non/any

### Manual Testing

Let's see if CI pass

### Related Issue

This is part of #2022


PatrykKuniczak:
fuNcccc. not that branch ...

PR #2119, opened 2026-02-12
chore: Created new types, instead of `any` for `analytics`

PatrykKuniczak:
### Overview

I've created a couple new types, instead of non/any

### Manual Testing

Let's see if CI pass

### Related Issue

This is part of #2022

### DIFF
--- a/packages/analytics/modules/analytics/client.ts
+++ b/packages/analytics/modules/analytics/client.ts
@@ -231,8 +231,8 @@ function createFrontendAnalytics(): Analytics {
     sessionId,
     timestamp: Date.now(),
     language: navigator.language,
-    referrer: globalThis.document.referrer || undefined,
-    screen: `${globalThis.window.screen.width}x${globalThis.window.screen.height}`,
+    referrer: document.referrer || undefined,
+    screen: `${window.screen.width}x${window.screen.height}`,
     url: location.href,
     title: document.title || undefined,
   });


packages/analytics/modules/analytics/types.ts
aklinker1, line 1: These are internal types, so I don't want to export them from `@wxt-dev/analytics/types`.

https://github.com/PatrykKuniczak/wxt/blob/b0cfc7c66bbc5afd9525222c874bad7590aa5ee4/packages/analytics/package.json#L25-L27

Either move them into `client.ts` (probably don't need to be exported from there) or to a different file that doesn't expose these APIs to package consumers.

packages/analytics/modules/analytics/client.ts
```diff
@@ -238,20 +238,20 @@ function createFrontendAnalytics(): Analytics {
...
         void analytics.track('click', {
           tagName: element.tagName?.toLowerCase(),
-          id: element.id || undefined,
-          className: element.className || undefined,
-          textContent: element.textContent?.substring(0, 50) || undefined, // Limit text content length
-          href: element.href,
+          id: element.id,
+          className: element.className,
+          textContent: element.textContent?.substring(0, 50), // Limit text content length
+          href: (element as HTMLAnchorElement).href,
```
aklinker1, line 266: I think the correct fix here is to modify `eventProperties?: Record<string, string>` to `eventProperties?: Record<string, string | undefined>` and keep the `|| undefined` for ignoring empty strings.

PatrykKuniczak:
@aklinker1 Should be good :)

PR #2121, opened 2026-02-13
chore!: Refactor and simplify i18n types to avoid `any`

PatrykKuniczak:
> [!CAUTION]
>
> **BREAKING CHANGE**
>
> If you use `createI18n` without type parameters, the return type has changed:
>
> ```diff
> - import { createI18n, type I18n, type DefaultI18nStructure } from '@wxt-dev/i18n'
> + import { createI18n, type UntypedI18n } from '@wxt-dev/i18n'
> 
> - const i18n: I18n<DefaultI18nStructure> = createI18n();
> + const i18n: UntypedI18n = createI18n();
> ```

### Overview

I've done my best, to type it, if anybody have better idea to do this, i'm open for edit suggestion.

### Manual Testing

Let's see if CI pass


packages/i18n/src/__tests__/types.test.ts
aklinker1, line 1: This file tests types by expecting type errors for invalid function calls. That means we need the `// @ts-expect-error` is present here. Revert it or find another way to test that type errors exist in these cases.

aklinker1 (commented):
I only looked a the test file. Revert those changes so the tests expect type errors for invalid function calls, then I'll look at the rest of this PR.

aklinker1 (changes requested):
Woops, meant to request changes.

PatrykKuniczak:
@aklinker1 Now it should be good

PatrykKuniczak:
@aklinker1 Conflicts resolved 😄

PatrykKuniczak:
@aklinker1 Again, let's check it :)

aklinker1:
@PatrykKuniczak What does this PR do? Why do your changes make the type-setup better? I'm not seeing anything immediately stand out after reviewing the changes. You made some good changes to make the internal code more type-safe, but I can't tell what your changes to the public APIs/types are doing

PatrykKuniczak:
> but I can't tell what your changes to the public APIs/types are doing

What do you mean?

packages/i18n/src/build.ts
aklinker1, line 1: I pulled out all the changes to this file to https://github.com/wxt-dev/wxt/pull/2201, they look good to me with a few tweaks.

packages/i18n/src/index.ts
```diff
@@ -1,16 +1,9 @@
...
-  Substitution,
-} from './types';
+import { I18nStructure, I18n, Substitution } from './types';
 import { browser } from '@wxt-dev/browser';
 
-export function createI18n<
-  T extends I18nStructure = DefaultI18nStructure,
->(): I18n<T> {
-  const t = (key: string, ...args: any[]) => {
+export function createI18n<T extends I18nStructure>(): I18n<T> {
```
aklinker1: OK, so we removed the default value from here... I'm not sure I like that. It is a breaking change for non-WXT projects. What do those projects use for the type param now?

packages/i18n/src/types.ts
```diff
@@ -7,13 +7,21 @@ export type I18nStructure = {
...
 };
 
+export interface I18n<T extends I18nStructure> {
+  t: string extends keyof T ? DefaultTFunction<keyof T & string> : TFunction<T>;
+}
+
 // prettier-ignore
 export type SubstitutionTuple<T extends SubstitutionCount> =
-    T extends 1 ? [$1: Substitution]
+    T extends 0 ? []
```
aklinker1: Why do we need to support 0 substitutions here?

packages/i18n/src/types.ts
```diff
@@ -7,13 +7,21 @@ export type I18nStructure = {
...
-  [K: string]: any;
+type DefaultTFunction<TKeys extends string> = {
+  (key: TKeys): string;
+  (key: TKeys, substitutions: string[]): string;
+  (key: TKeys, n: number): string;
+  (key: TKeys, n: number, substitutions: string[]): string;
 };
 
+export interface I18n<T extends I18nStructure> {
+  t: string extends keyof T ? DefaultTFunction<keyof T & string> : TFunction<T>;
```
aklinker1: OK, so we're splitting up the typing... We're using the new `DefaultTFunction` type when.... `T extends never`? Is that what `T` defaults to when not provided in `createI18n`?

packages/i18n/src/types.ts
```diff
@@ -31,10 +39,24 @@ export type TFunction<T extends I18nStructure> = {
...
     // prettier-ignore
-    key: K & { [P in keyof T]: T[P] extends { plural: false; substitutions: SubstitutionCount } ? P : never; }[keyof T],
+    key: K & { [P in keyof T]: T[P] extends { plural: false; substitutions: 2 } ? P : never; }[keyof T],
+    substitutions: SubstitutionTuple<2>,
+  ): string;
+
+  // Non-plural with 3+ substitutions
+  <K extends keyof T>(
+    // prettier-ignore
+    key: K & { [P in keyof T]: T[P] extends { plural: false; substitutions: 3 | 4 | 5 | 6 | 7 | 8 | 9 } ? P : never; }[keyof T],
```
aklinker1: What is the point of separating substitution counts of 1, 2, and 3-9 into separate overrides?

aklinker1:
> What do you mean?

What does this PR do? You haven't stated anywhere what the purpose of these changes are and it's hard for me to guess by just looking at the changes. Is the purpose just to remove `any` from the package?

If so, your PR title and description don't mention that at all.

PatrykKuniczak:
> > What do you mean?
> 
> What does this PR do? You haven't stated anywhere what the purpose of these changes are and it's hard for me to guess by just looking at the changes. Is the purpose just to remove `any` from the package?
> 
> If so, your PR title and description don't mention that at all.

Ooo yeah, i've wanted to avoid `any`, i'll se e your PR and the we'll be able to discuss rest changes from this one

packages/i18n/src/types.ts
PatrykKuniczak: I18nStructure is defined as { [K: string]: I18nFeatures } - a signature with string keys
When T = I18nStructure, keyof T evaluates to string = true
So the condition string extends keyof T evaluates to string extends string
Therefore it uses DefaultTFunction (the "no type-safety" version)

packages/i18n/src/types.ts
PatrykKuniczak: No, removed

packages/i18n/src/types.ts
PatrykKuniczak: I've missed it during review of generated code, sorry, i've fixed it already, it indeed can be simplified 😄

packages/i18n/src/index.ts
PatrykKuniczak: If they doesn't provide any type, I18nStructure will take default value:
```ts
export type I18nStructure = {
  [K: string]: {
    plural: boolean;
    substitutions: 0 ... 9;
};
```

I'm right?

PatrykKuniczak:
@aklinker1 I hope now, it's more readable 😄

PatrykKuniczak:
@aklinker1 It's good to go or you need sth extra?

packages/i18n/src/types.ts
```diff
@@ -7,10 +7,17 @@ export type I18nStructure = {
...
-  [K: string]: any;
+type DefaultTFunction<TKeys extends string> = {
+  (key: TKeys): string;
+  (key: TKeys, substitutions: string[]): string;
+  (key: TKeys, n: number): string;
+  (key: TKeys, n: number, substitutions: string[]): string;
 };
 
+export interface I18n<T extends I18nStructure> {
+  t: string extends keyof T ? DefaultTFunction<keyof T & string> : TFunction<T>;
```
aklinker1: OK, I had to go through and test what each of your changes do, and I think I get why most where made.

The main thing that sticks out to me was "wrong" is this line where you decide between using the typed and untyped versions of `t`. It seemed inefficient or overly complex.

If the main issue here is that we need to decide between returning a typed `I18n` object and an "untyped" version... let's just do that in an override that returns a separate type! That way there's no weird conditions like this and there's no extra type param when it's not needed. Much more readible when investigating types in your IDE.

Here's my proposed changes: https://github.com/wxt-dev/wxt/commit/66e5d090270af397fc83e4c4f36d609a232871d3

packages/i18n/src/types.ts
```diff
@@ -7,10 +7,17 @@ export type I18nStructure = {
   [K: string]: I18nFeatures;
 };
 
-export type DefaultI18nStructure = {
-  [K: string]: any;
```
aklinker1, line 11: Another concern is that we're removing a fairly important type. I'm tempted to make this PR a breaking change - we are changing the type-system quite drastically for people who don't use the typed `I18n` object. Some projects consider types breaking changes, some don't. WXT does.

So let's do that. Use `chore!: ...` for the PR title, we'll bump this from 0.2 &rarr; 0.3.

Ideally, I would not make any breaking changes, ever. We need good API design from the start. Is getting rid of `any` really worth forcing users to go through migrations? I always hate doing major version upgrades to libraries. Consider that before proposing a breaking change in the future.

packages/i18n/src/types.ts
```diff
@@ -22,7 +29,7 @@ export type SubstitutionTuple<T extends SubstitutionCount> =
   : T extends 7 ? [$1: Substitution, $2: Substitution, $3: Substitution, $4: Substitution, $5: Substitution, $6: Substitution, $7: Substitution]
   : T extends 8 ? [$1: Substitution, $2: Substitution, $3: Substitution, $4: Substitution, $5: Substitution, $6: Substitution, $7: Substitution, $8: Substitution]
   : T extends 9 ? [$1: Substitution, $2: Substitution, $3: Substitution, $4: Substitution, $5: Substitution, $6: Substitution, $7: Substitution, $8: Substitution, $9: Substitution]
-  : never
+  : []
```
aklinker1: This line and your `SubstitutionCount` &rarr; `Exclude<SubstitutionCount, 0>` changes I don't understand. Could you provide more context on why you made this change?


If I had to guess at this change, the `SubstitutionCount` includes 0, but there was no tuple for that number, so you added one? 

If I revert this change and remove the `Exclude` you added, everything still works and I think `SubstitutionTuple` type is more representative of how it's used - we only use this type when the count is guaranteed to be between 1-9, it doesn't need to support 0.

packages/i18n/src/types.ts
```diff
@@ -7,10 +7,17 @@ export type I18nStructure = {
   [K: string]: I18nFeatures;
 };
 
-export type DefaultI18nStructure = {
-  [K: string]: any;
+type DefaultTFunction<TKeys extends string> = {
+  (key: TKeys): string;
+  (key: TKeys, substitutions: string[]): string;
+  (key: TKeys, n: number): string;
+  (key: TKeys, n: number, substitutions: string[]): string;
```
aklinker1:
```suggestion
  (key: TKeys, substitutions: Substitution[]): string;
  (key: TKeys, n: number): string;
  (key: TKeys, n: number, substitutions: Substitution[]): string;
```

aklinker1 (changes requested):
LMK what you think of my proposed changes (first comment)

packages/i18n/src/types.ts
PatrykKuniczak: Ok. let's use your proposal

packages/i18n/src/types.ts
PatrykKuniczak: Yeah, breaking changes isn't comfy, but it happens from time to time for libs and IMO.
We shouldn't stop us from making better code, because of the people which'll need to bump WXT.

I hope this change improve our codebase :)
Anyway we have several breaking changes in the past, and we didn't gain a big lose of contributors and people which installs WXT, as i mentioned above, it happens from time to time

packages/i18n/src/types.ts
PatrykKuniczak: Yeah, i wantend to avoid returning other type that everything else for very last else statement.

Let's do with that whatever you want, if you want revert, let's do it.

PatrykKuniczak:
@aklinker1 Let's apply your proposal here and then, we'll wait for next major :)

aklinker1:
This isn't for the core `wxt` package, so it doesn't need to be queued up the same way. We can just merge it to `main` and release a new version. I don't expect there to be any more breaking changes to this package in the near future.

In the future, I'll use `major-wxt` as the branch name instead of just `major`

aklinker1 (approved):

PR #2122, opened 2026-02-13
chore: Simplify imports in `wxt/e2e`

PatrykKuniczak:


aklinker1 (approved):

PatrykKuniczak:
@aklinker1 About https://github.com/wxt-dev/wxt/pull/2122/changes/bebd618411eb79b3761b8910d875307b9bfc7f61
This should be configure in eslint config. because it's easy to miss, but with new option in eslint we can have better imports sort.

I was reading about it: https://github.com/wxt-dev/wxt/pull/2014#issuecomment-3679022213

PR #2123, opened 2026-02-13
chore: Fix typo of `manifest.test.ts`

PatrykKuniczak:
### Overview

I've removed it because `mv2Key` can be `undefined`

### DIFF
--- a/packages/wxt/src/core/utils/__tests__/manifest.test.ts
+++ b/packages/wxt/src/core/utils/__tests__/manifest.test.ts
@@ -37,7 +37,8 @@ describe('Manifest Utils', () => {
       const popupEntrypoint = (type?: ActionType) =>
         fakePopupEntrypoint({
           options: {
-            mv2Key: type,
+            // @ts-expect-error: Force this to be undefined instead of inheriting the random value
+            mv2Key: type ?? null,
             defaultIcon: {
               '16': '/icon/16.png',
             },


packages/wxt/src/core/utils/__tests__/manifest.test.ts
```diff
@@ -37,12 +37,11 @@ describe('Manifest Utils', () => {
       const popupEntrypoint = (type?: ActionType) =>
         fakePopupEntrypoint({
           options: {
-            // @ts-expect-error: Force this to be undefined instead of inheriting the random value
-            mv2Key: type ?? null,
+            mv2Key: type,
```
aklinker1: This makes the manifest tests flakey - run `pnpm test run manifest.test.ts` 10 or so times, you'll see it fail. 

Since `fakePopupEntrypoint` uses random data, and these overrides are merged by `lodash.merge`, we need to convert `undefined` to `null` because values that are `undefined` do not override lower priorty values. 

That's what the comment was explaining.


```suggestion
            // @ts-expect-error: Force this to be undefined instead of inheriting the random value
            mv2Key: type ?? null,
```

aklinker1 (changes requested):
Good find with the typo.

PatrykKuniczak:
@aklinker1 Ready to merge 😄

PR #2124, opened 2026-02-13
chore: Add type for ExampleModuleOptions of `wxt-demo`

PatrykKuniczak:
### Overview

I want to avoid `@ts-expect-error`, but if you @aklinker1 want it, feel free to reject and close this PR.


aklinker1:
This was intentional to confirm that all values, even untyped ones, show up in the result of `useAppConfig`, as described in the comment

PR #2125, opened 2026-02-13
chore: Remove unnecessary `// @ts-ignore` from `auto-import` output

PatrykKuniczak:
### Overview

I've removed it, because it's unnecessary, there's no error, but maybe anybody knows, who to do it in more pretty way, that's not bad, but MEH.


aklinker1 (commented):
If `unimport` is exporting types with `@ts-ignore`, I'm gonna trust in the Nuxt team's decision to use them.

https://github.com/unjs/unimport/blob/e908b928f122e79b1f175eedc484be2f0c17e836/src/utils.ts#L247-L266

Not sure if the reason they provide is good or not for the latest version of typescript, but I'll let them decide when it's OK to remove them or not.

PatrykKuniczak:
@aklinker1 No problem, we'll see: https://github.com/unjs/unimport/issues/499 :)

aklinker1:
Perfect, thanks for creating that!

PR #2126, opened 2026-02-13
chore: Remove `zipdir` of `packages/wxt`

PatrykKuniczak:
### Overview

I want to rename it, to match JS most common `cammelCase`.

I know this is breaking changes, we can't merge it now, but i think this can helps dev to have unified syntax for our methods, because most of that have `cammelCase`.

Migration will be veeery easy.

I hope this can go 😄

### DIFF
--- a/packages/wxt/src/@types/modules.d.ts
+++ b/packages/wxt/src/@types/modules.d.ts
@@ -1,22 +1,5 @@
 // Custom TS definitions for non-TS packages
 
-declare module 'zip-dir' {
-  // Represents the options object for zipDir function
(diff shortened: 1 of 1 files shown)


PatrykKuniczak:
@aklinker1 Done :)

PatrykKuniczak:
@aklinker1 I won't be active for a while, because i have been trying to push my product on the market.
I have some clients, which are open to test it and buy after a test.
That's not because of your PR rejection or sth like that, i simply need to focus on my goals, to be able(hopefuly) to leave my full time job and do my own stuff 😄

PR #2128, opened 2026-02-13
chore: Make type for `__ENTRYPOINT__` of vitest.globalSetup.ts

PatrykKuniczak:
### Overview

I've done it do avoid `@ts-expect-error`.

### DIFF
--- a/packages/wxt/globals.d.ts
+++ b/packages/wxt/globals.d.ts
@@ -1,3 +1,3 @@
-declare module globalThis {
+declare namespace globalThis {
   var __ENTRYPOINT__: string;
 }


PatrykKuniczak:
@aklinker1 That was last of my big refactor #2022 
I HOPE a couple of that changes will be helpful.

I don't want to overturn the table, but push project ahead.

I wasn't disgusted or sth like that, because of your review
That's a cold bucket poured on my head, but i need it, we need it, to be able to cooperate, thanks for all of your reviews, past and future.

I'll be less and less active for some time, but i'll be trying to support project in other way, than making a code, e.g support others in PR and doing reviews, but this refactor and some fixes are all of my to the nearest time, keep pushing, that's good product :)

aklinker1:
Sounds good, thanks for understanding! I'm reviewing all of them now.

aklinker1 (approved):

PR #2130, opened 2026-02-14
fix(storage): migrations incorrectly running

### LINKED ISSUE #1775: Meta is not set when storage item with migrations defined is initially created
### Describe the bug

Given that there's an item definition with migrations (https://wxt.dev/storage), for example 
```
const item = storage.defineItem<{
    value: string;
}>('local:data', {
    version: 2,
    migrations: {
        2: () => {
            // some code
        },
    },
});
```
Assume that user installed extension for the first time (i.e. there's no version 1 in their storage).

Setting any value to this item will store it without version metadata, which will lead to migration run when refreshing the page (if e.g. it's accessed in some content script). 

This is usually incorrect cause migration code assumes data has old format, which never existed and was never set for a fresh install.

A PR with a failing test case: https://github.com/wxt-dev/wxt/pull/

### Reproduction

A PR with a failing test case: https://github.com/wxt-dev/wxt/pull/1774

### Steps to reproduce

_No response_

### System Info

```shell
System:
    OS: Windows 11 10.0.26100
    CPU: (16) x64 Intel(R) Core(TM) i7-10700KF CPU @ 3.80GHz
    Memory: 34.40 GB / 63.87 GB
  Binaries:
    Node: 22.14.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.22 - C:\Program Files\nodejs\yarn.CMD
    npm: 11.4.2 - C:\Program Files\nodejs\npm.CMD
    pnpm: 9.1.4 - C:\Program Files\nodejs\pnpm.CMD
  Browsers:
    Edge: Chromium (133.0.3065.82)
    Internet Explorer: 11.0.26100.1882
```

### Used Package Manager

npm

### Validations

42willow:
* when items initiated at v >= 2
* fix #1775

follow up of #1996, which wasn't effective because of this commit https://github.com/wxt-dev/wxt/commit/c4577d72263a4b9e61605656be81d25df5a2a734 (sorry I didn't test this properly with my extension first...)

turns out there was a reported issue for this! #1775

### DIFF
--- a/docs/.vitepress/components/UsingWxtSection.vue
+++ b/docs/.vitepress/components/UsingWxtSection.vue
@@ -117,6 +117,7 @@ const chromeExtensionIds = [
   'jdcppdokgfbnhiacbeplahgnciahnhck', // Lofi BGM Player - Free lofi focus music for work & study
   'cgpmbiiagnehkikhcbnhiagfomajncpa', // Margin - Annotate and highlight any webpage, with your notes saved to the decentralized AT Protocol.
   'mfjdonmgmgcijagclnkfhmjiblbfjaid', // KeyFloat - Floating multilingual keyboard with native key mappings, drag, dark mode, sounds, and dynamic layouts for macOS & Windows
+  'dhiekgdaipindoapjmcnpompdknjeijf', // Glossy New Tab - Say Goodbye to Boring Tabs with live wallpapers
 ];
 
 const { data, err, isLoading } = useListExtensionDetails(chromeExtensionIds);


42willow:
@aklinker1 i think there was some confusion in the previous PR (https://github.com/wxt-dev/wxt/pull/1996#issuecomment-3658151972 and https://github.com/wxt-dev/wxt/pull/1996#discussion_r2784201403)

By "migrations incorrectly running on initialisation" I mean they are incorrectly running when a value is first set in local storage, not necessarily from the init function. I have moved the fix to the migration logic as `getOrInitValue` runs after migrations have completed which makes things confusing as the first time around the version is not set in meta, causing the test, specifically at [this line](https://github.com/42willow/wxt/blob/b84e030103a2cd180ace1b17a80b51a753238395/packages/storage/src/__tests__/index.test.ts#L719) to fail.

@absdjfh explains it well in https://github.com/wxt-dev/wxt/issues/1775

42willow:
I have been using this fix at https://github.com/schooltape/schooltape/commit/33c63391508ce0b75aee19b9d17e13af6a11d65d and will post updates here if I notice anything strange happening with the storage after this change

packages/storage/src/__tests__/index.test.ts
```diff
@@ -789,7 +791,7 @@ describe('Storage Utils', () => {
         const actualMeta = await item.getMeta();
 
         expect(actualValue).toEqual(0);
-        expect(actualMeta).toEqual({});
+        expect(actualMeta).toEqual({ v: 3 });
```
aklinker1, line 794: *Edit: Ignore this comment, I didn't read the new issue and missed a few comments on the PR. I understand the problem better now and am considering the PR again*

<details>

This test is was correct - `defaultValue` is not the same as `init`:

- Default values are returned when no value is found in storage, but ***it doesn't set a value in storage***.
- Items with `init` ***do set a value in storage when no value is found***

That means `init` should set the meta, since there's now a value to control the version of. But for default values, there is no value set in storage, nothing to control the version of.

Here's an example:

```ts
// v1
const blockedUrls = defineStorageItem<string[]>("local:blockedUrls", {
  defaultValue: [],
})

// v2
const blockedUrls = defineStorageItem<{ id: string, url: string }>("local:blockedUrls", {
  defaultValue: [],
  version: 2,
  migrations: {
    2: (v1) => v1.map((url, i) => ({ id: String(i), url })),
  },
})
```

If a user never updates this storage item, why do we need to track that a blank array is v1 or v2? Either way, it works. But not adding meta will mean storage is less poluted and takes up less space. Hence why we don't set the version for default values.

</details>

packages/storage/src/index.ts
```diff
@@ -415,7 +415,13 @@ function createStorage(): WxtStorage {
...
           driverMetaKey,
         ]);
-        if (value == null) return;
+        if (value == null) {
+          // Prevent running migrations if initiated at a version >= 2
+          if (targetVersion >= 2) {
+            await setMeta(driver, driverKey, { v: targetVersion });
+          }
+          return;
+        }
```
aklinker1, line 424: *Edit: Ignore this comment, I didn't read the new issue and missed a few comments on the PR. I understand the problem better now and am considering the PR again*

<details>

Yeah, this is just invalid. If there's no value in storage, there no need to track a version.

</details>

packages/storage/src/__tests__/index.test.ts
```diff
@@ -706,16 +706,18 @@ describe('Storage Utils', () => {
...
-          init: () => 1,
+        let item = storage.defineItem<number>('local:key', {
           version: 2,
           migrations: {
             2: migrate,
           },
         });
+        await item.setValue(1);
+        await item.migrate();
+
```
aklinker1, line 720: This test is still valid I think. Can you create a new test instead of editing this one?

Though I think the test name should have " with init defined".

packages/storage/src/__tests__/index.test.ts
aklinker1: OK, with more context, I still believe that this test was valid... I'd prefer to not add a version for a `null` value in storage.

That said, if there's not a better place to put the new logic, we may have to accept this behavioral change.

packages/storage/src/__tests__/index.test.ts
```diff
@@ -899,17 +901,19 @@ describe('Storage Utils', () => {
...
           migrations: {
             2: () => {
               throw cause;
             },
           },
         });
-        await fakeBrowser.storage.local.set({ key: 1, key$: { v: 1 } });
 
         await expect(item.migrate()).rejects.toThrow(expectedError);
       });
```
aklinker1, line 919: Once again, I beileve the previous test is still valid. Add a new one instead.

aklinker1 (commented):
> By "migrations incorrectly running on initialisation" I mean they are incorrectly running when a value is first set in local storage, not necessarily from the init function. I have moved the fix to the migration logic as getOrInitValue runs after migrations have completed which makes things confusing as the first time around the version is not set in meta, causing the test, specifically at [this line](https://github.com/42willow/wxt/blob/b84e030103a2cd180ace1b17a80b51a753238395/packages/storage/src/__tests__/index.test.ts#L719) to fail.
> 
> @absdjfh explains it well in https://github.com/wxt-dev/wxt/issues/1775

OK, yup, that clears up my confusion.

I've left some comments, I think the fix here is changing existing behavior too much - you changed some tests that should still be valid because, I assume, your changes broke them. That said, I'm still trying to decide the right place for this logic.

So hold off on any changes until I leave another review after I've dug around a bit more.

aklinker1:
I think there's a few options to fix this:

1. Set it during the migration like this pr
   - Downside: We set a version in storage when it's not needed
2. Set it when the value is set
   - Downside: Excessive calls to storage APIs to check the version each time

OR combine the above two options and track that the version needs set during the migration, but only actually set it when calling `setValue` for the first time. I think this is the right move.

- It doesn't set a version in storage when it's not needed
- It doesn't make excessive storage API calls

The only downside is we have to keep one more boolean in memory... which I think is worth it.

I created a PR with my fix (added both you guys as contributors on it), @42willow please review and test it with your extension: https://github.com/wxt-dev/wxt/pull/2139

PR #2136, opened 2026-02-16
feat: use Navigation API for location change detection with polling fallback

### LINKED ISSUE #1567: Improved `wxt:locationchange`
### Feature Request

Currently, this event is powered by polling the current URL every second. This means it's slow to react to URL changes.

Unfortunately, it's not as easy as monkeypatching the history APIs:

```ts
const ogPushState = history.pushState;
history.pushState = (...args) => {
  const result = ogPushState.apply(history, args);
  // do something
  return result;
};
```

This doesn't work in isolated content scripts. `history.pushState` in the content script is not the same function as `history.pushState` in the main world.

So that means we need to inject a script to monkey-patch these APIs.

Something like this:

```ts
// wxt-locationchange.ts
  const ogPushState = history.pushState;
  const ogReplaceState = history.replaceState;

  let prevUrl = location.href;

  function dispatchNavigationEvent() {
    const newUrl = location.href;
    if (newUrl !== prevUrl) {
      window.dispatchEvent(
        new CustomEvent('wxt:locationchange', { detail: { prevUrl, newUrl } }),
      );
    }
    prevUrl = newUrl;
  }

  // Monkeypatch SPA navigation in main world
  history.pushState = (...args) => {
    const result = ogPushState.apply(history, args);
    dispatchNavigationEvent();
    return result;
  };
  history.replaceState = (...args) => {
    const result = ogReplaceState.apply(history, args);
    dispatchNavigationEvent();
    return result;
  };

  // Listen to back/forward button clicks
  window.addEventListener('popstate', dispatchNavigationEvent);
});
```

```ts
// content.ts
  // ...
  main(ctx) {
    injectScript("/wxt-locationchange.ts");

    ctx.addListener(window, "wxt:locationchange", (event) => {
      console.log("Locaiton changed!", event.detail);
    });
  },
});
```

WXT should be able to generate this `wxt-locationchange.js` file for you, and include it in your build when you need it. Then when you call `ctx.addEventListener(window, "wxt:locationchange", () => { ... })`, WXT should inject the monkeypatching script instead of polling.

Downside to this approach is that we're modifying global variables... Which is frowned upon in general. Not really sure what else we can do about that 🤷 

https://discord.com/channels/1212416027611365476/1357508502804758649

None others that I'm aware of.

### Additional context

Related to #1029. Could simply be that when a content script has `spa: true` in it's definition that WXT adds this logic.

Aler1x:
### Overview

`wxt:locationchange` used to rely on polling `location.href` every second, so URL updates (e.g. SPA navigations) were detected with up to 1s delay. This change uses the [Navigation API](https://developer.mozilla.org/en-US/docs/Web/API/Navigation_API) when available: we subscribe to the navigate event so URL changes are handled immediately and no timer is used. In environments where the Navigation API is not available (e.g. older Firefox or Safari), behavior is unchanged and we keep the existing 1s polling. No new APIs for extension authors and no breaking changes.

### Manual Testing

1. Build WXT and the demo: `pnpm --filter wxt build` then `pnpm --filter wxt-demo build`.
2. Load the unpacked extension from `packages/wxt-demo/.output/chrome-mv3` (or equivalent for your build).
3. Open a site that uses client-side navigation (e.g. the demo’s `*://*.crunchyroll.com/*`), open DevTools → Console.
4. Navigate within the site (links that change the URL without full reload). You should see `Location changed: <newUrl> <oldUrl>` and `Watching for location change...`.
5. (Optional) In a Navigation API–capable browser the event fires immediately; with the polling fallback (e.g. force it by disabling the API in code or using an older browser) the same log appears within ~1s.

### Related Issue

This PR closes #1567

### DIFF
--- a/packages/wxt/src/utils/internal/location-watcher.ts
+++ b/packages/wxt/src/utils/internal/location-watcher.ts
@@ -2,8 +2,7 @@ import { ContentScriptContext } from '../content-script-context';
 import { WxtLocationChangeEvent } from './custom-events';
 
 const supportsNavigationApi =
-  (globalThis as any).navigation !== 'undefined' &&
-  (globalThis as any).navigation.addEventListener === 'function';
+  typeof (globalThis as any).navigation?.addEventListener === 'function';
 
 /**
  * Create a util that watches for URL changes, dispatching the custom event when detected. Stops


aklinker1 (approved):
Nice, looks like Firefox just added support for this event in January.

Was wondering why I didn't know about it? But it's been around for awhile in other browsers... Oh well, good to have it now!

Issue #544, opened 2024-03-12
Support more manifest fields in entrypoint options


aklinker1:
### Feature Request

When defining an entrypoint, there are additional settings/options that would be useful for managing the manfiest instead of adding those fields to `wxt.config.ts` in one place:

- [ ] `permissions` and `host_permissions` for all entrypoints
  - #379
- [ ] `web_accessible_resources` for unlisted pages/scripts
  - #536 
- [ ] ...

These options would be merged together in the final manifest.

### What are the alternatives?

Continue to maintain these options in the config file.

placoderm:
I would really love to see this. Here is a user story, not that you necessarily need one…

I just submitted to the Chrome web store and they suggested changing the permission from `tabs` to `activeTab`. My plugin has a large number of little features so it would have been helpful to know exactly which ones needed `tabs` and be able to test out a change.

Thanks for considering this improvement. Seems like a logical thing to offer in a framework.

aklinker1:
Closing as per reasons here: https://github.com/wxt-dev/wxt/pull/2142#issuecomment-3941656032

TLDR: Keep configuring global options globally in one place.

Issue #536, opened 2024-03-09
Mention the need to manually add unlisted script into `web_accessible_resources` if it's being injected.


KraXen72:
### Feature Request

After 2 hours of debugging, I've figured out that for an unlisted script to be injectable into the page it has to be declared in the manifest, like so:
```ts
// See https://wxt.dev/api/config.html
export default defineConfig({
	manifest: {
		web_accessible_resources: [
			{ resources: ["page.js"], matches: ["<all_urls>"] }
		]
	}
});
```
it would be good to mention this in the unlisted scripts docs or add it there automatically.

#### Is your feature request related to a bug?

N/A

### What are the alternatives?

N/A

### Additional context

This is my setup:
```ts
// page.ts
export default defineUnlistedScript(() => {
	console.log('unlisted script!')
})
```
  
```ts
// content.ts
export default defineContentScript({
	matches: ['<all_urls>'],
	main() {
		let pageScriptElem: HTMLScriptElement | null = null;
		
		browser.runtime.onMessage.addListener(
			async function (request, sender, sendResponse) {
				if (request.cmd && request.cmd === 'process') {
					if (pageScriptElem !== null) pageScriptElem.remove();

					pageScriptElem = document.createElement("script")
					pageScriptElem.src = browser.runtime.getURL("/page.js")
					pageScriptElem.type = 'text/javascript'
					document.head.appendChild(pageScriptElem)
				}
			}
		);
	}
});
```
Let me know if i'm misunderstanding how one should inject scripts like these

aklinker1:
Nope, this looks correct. I'll update the docs.

KraXen72:
it would also maybe be worthwhile to add something like an `accessible: boolean` flag to unlisted scripts, that automatically adds them to the web-accessible resources for some given urls. or some other way to automate it. but if it's too much work, a mention in the docs works too.

aklinker1:
> it would also maybe be worthwhile to add something like an accessible: boolean flag to unlisted scripts, that automatically adds them to the web-accessible resources for some given urls. or some other way to automate it. but if it's too much work, a mention in the docs works too.

Hmm, for now, I'm gonna pass on this. In addition to #379, I'll create a new issue to supporting adding more manifest related options to the entrypoints.

mundane799699:
> ### Feature Request
> After 2 hours of debugging, I've figured out that for an unlisted script to be injectable into the page it has to be declared in the manifest, like so:
> 
> ```ts
[...]

PR #2142, opened 2026-02-21
fix(types): Add type safety to `browser.runtime.executeScript` `files` option

nickbar01234:
### Overview

Previously, `browser.script.executeScript({ files: [] })` does not provide any type hint about what file paths are injectable. This PR augments the type definition to all `ScriptPublicPath`, generated from `path.d.ts`. 

A couple of notes
- I think `ScriptPublicPath` union is wider than what can be provided to `executeScript`. The worst case is we run into runtime errors, which is similar to the state before this change.
- This is a breaking change for downstream consumers that specifies a non-existent scripts 

### Manual Testing

- [X] Manual link with a local extension to confirm the build generates the right type definition

### Related Issue

### DIFF
--- a/packages/wxt-demo/src/entrypoints/background.ts
+++ b/packages/wxt-demo/src/entrypoints/background.ts
@@ -49,7 +49,7 @@ function _otherTypeChecksNotEvaluated() {
     target: { tabId: 1 },
     files: [
       '/background.js',
-      // @ts-expect-error: Should error out for unknown strings
+      // @ts-expect-error: Should error for non-existing paths
       '/other.js',
     ],
   });


packages/wxt/src/core/generate-wxt-dir.ts
```diff
@@ -190,6 +197,58 @@ declare module "wxt/browser" {
...
+  const text = `// Generated by wxt
+import "wxt/browser";
+import { WxtBrowser } from "wxt/browser";
+
+declare module "wxt/browser" {
+  type TypedScriptInjection = Parameters<
+      WxtBrowser["scripting"]["executeScript"]
+    >[0] & { files: Array<${paths.join(' | ')}>; func?: never | undefined };
+
+    type ReplaceFirstArg<Fnc, Arg> = Fnc extends {
```
nickbar01234: tbh: Not a big fan of this typescript magic, so open to other suggestions

nickbar01234: Alternatively, @aklinker1 curious if it's an option to update the backing `@wxt-dev/browser` package to have a generic `T` for file paths, which we can dynamically declare when generating the declaration file based on the filenames

aklinker1: @nickbar01234 Did you try how we do it here?

https://github.com/wxt-dev/wxt/blob/430a3acfeb25410d07e0ca48981a6f819762f29d/packages/wxt/src/core/generate-wxt-dir.ts#L89-L108

https://github.com/wxt-dev/wxt/blob/430a3acfeb25410d07e0ca48981a6f819762f29d/packages/wxt/src/browser.ts#L14-L31

We should be able to simplify this quite a bit. Unless there's something different about what you're doing here.

aklinker1: > if it's an option to update the backing @wxt-dev/browser package to have a generic T for file paths, which we can dynamically declare when generating the declaration file based on the filenames

I'm hesitant to make excessive changes to that package since it's generated based on `@types/chrome`. It may be possible to automate, but it would become more fragile. I'd prefer we use the existing patterns for augmenting the browser types

nickbar01234: Gotcha, I think I see what [wxt/packages/wxt/src/browser.ts](https://github.com/wxt-dev/wxt/blob/430a3acfeb25410d07e0ca48981a6f819762f29d/packages/wxt/src/browser.ts#L14-L31) is doing. Let me try refactoring this code to match the pattern

packages/wxt/src/browser.ts
```diff
@@ -23,9 +24,36 @@ export interface WxtRuntime {}
  */
 export interface WxtI18n {}
 
-export type WxtBrowser = Omit<typeof _browser, 'runtime' | 'i18n'> & {
+type ScriptInjection<Args extends any[], Result> =
+  Browser.scripting.ScriptInjection<Args, Result> extends infer T
+    ? T extends { files: string[] }
+      ? Omit<T, 'files'> & { files: ScriptPublicPath[] }
```
nickbar01234, line 30: API docs say that file injections can be either CSS or JS, which does not align with my understanding of `executeScript` API.

>  The path of the JS or CSS files to inject, relative to the extension's root directory. Exactly one of the files or func must be specified.


If this is the case, we'd need to add more union to `ScriptPublicPath`. The other concern is downstream consumers with wrong file name will break when they pick up this version. My thought is that it's ok for better type safety

nickbar01234:
@aklinker1 just updated the implementation whenever you have some downtime to review!

packages/wxt/src/browser.ts
```diff
@@ -23,9 +24,36 @@ export interface WxtRuntime {}
...
+    <Args extends any[], Result>(
+      injection: ScriptInjection<Args, Result>,
+    ): Promise<InjectionResult<Result>>;
+
+    <Args extends any[], Result>(
+      injection: ScriptInjection<Args, Result>,
+      callback: (results: InjectionResult<Result>) => void,
+    ): void;
+  };
+}
```
aklinker1: Huh, I guess we don't need to generate some of the overrides per-project, like the runtime overrides... I could just import the `PublicPath` type like you did here...

packages/wxt/src/browser.ts
```diff
@@ -23,9 +24,36 @@ export interface WxtRuntime {}
  */
 export interface WxtI18n {}
 
-export type WxtBrowser = Omit<typeof _browser, 'runtime' | 'i18n'> & {
+type ScriptInjection<Args extends any[], Result> =
+  Browser.scripting.ScriptInjection<Args, Result> extends infer T
```
aklinker1, line 28: I've never seen an this before... I thought you always had to extend something when using `extends`. Does this effectively create a variable called `T`?

aklinker1 (approved):
Perfect, thanks. Much simpler!

packages/wxt/src/browser.ts
nickbar01234: Yeah, but `T` is the union over all `Browser.scripting.ScriptInjection<Args, Result>`

packages/wxt/src/browser.ts
nickbar01234: not too familiar with the typescript build process itself, so I was mostly referencing https://github.com/wxt-dev/wxt/blob/main/packages/wxt/src/utils/inject-script.ts#L4-L8 that also uses the public type before they existed

packages/wxt/src/browser.ts
aklinker1: Interesting...

packages/wxt/src/browser.ts
aklinker1: I agree, it should be fine for now.

MDN doesn't include CSS files in their docs: https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/scripting/executeScript

I'd say it's fine for now, if someone opens an issue we'll know what needs added.

packages/wxt/src/browser.ts
aklinker1: Yeah, I think I'll move to this when possible. Modules like i18n will still need to generate it's own module augmentations, but the static stuff can be put here.

nickbar01234:
btw @aklinker1, I notice that we don't collect CSS files in public path and the scripts eligible for `injectScripts` should be declared `web_accessible_resources` right? Is it worth doing a cleanup here too?

aklinker1:
> scripts eligible for injectScripts should be declared web_accessible_resources right? Is it worth doing a cleanup here too?

@nickbar01234 I'm against making assumptions to generate `web_accessible_resources`. For unlisted scripts, they can also be used as web workers, which don't always need to be added to `web_accessible_resources`. 

That said, I could see a feature like this being useful. Feel free to implement if you'd like: https://github.com/wxt-dev/wxt/issues/536#issuecomment-1987307593

nickbar01234:
> > scripts eligible for injectScripts should be declared web_accessible_resources right? Is it worth doing a cleanup here too?
> 
> @nickbar01234 I'm against making assumptions to generate `web_accessible_resources`. For unlisted scripts, they can also be used as web workers, which don't always need to be added to `web_accessible_resources`.
> 
> That said, I could see a feature like this being useful. Feel free to implement if you'd like: [#536 (comment)](https://github.com/wxt-dev/wxt/issues/536#issuecomment-1987307593)

gotcha. Is anyone actively working on #544? The issue you referenced seems like it can be bundled as part of the issue

aklinker1:
Actually, nevermind. After looking at #544 again, I don't think I want to add these features. It might seem convenient at first, but managing all these global options (permissions, web accessible resources) in one place keeps things simple. If they're spread all over the place, you have to look everywhere for them - imagine removing a permission from one entrypoint, forgetting to remove it from another. Seems like it will cause bugs and tech debt... I think having them in one place is the right move.

And thinking about it more, a `accessible: true` option wouldn't be enough to build a MV3 web accessible resources entry.

So I'm actually gonna close #544 as not planned, sorry about that.

PR #2143, opened 2026-02-21
fix: escape virtual module import paths with apostrophes

### LINKED ISSUE #1697: Project name with `'` fails on `wxt` command
### Describe the bug

Project name with `'` fails on `wxt` command:

> ERROR  [vite:define] Transform failed with 1 error:
> virtual:wxt-background-entrypoint?existing-path-removed/hello'/entrypoints/background.ts:1:59:

> ERROR: Expected ";" but found "/"
> file: virtual:wxt-background-entrypoint?existing-path-removed/hello'/entrypoints/background.ts

### Reproduction

Reference the following section:

### Steps to reproduce

```shell
pnpm dlx wxt@latest init

# WXT 0.20.6
# ℹ Initializing new project
# √ Project Directory ... hello'
# √ Choose a template » vanilla
# √ Package Manager » pnpm
# ✔ Downloading template

# ✨ WXT project created with the vanilla template.

# Next steps:
#   1. cd hello' - does not work in PowerShell x Windows
#   2. pnpm install
```

```shell
cd "hello'" # requires "" around the path
pnpm i && pnpm dev

# WXT 0.20.6
# ✔ Started dev server @ http://localhost:3000
# ℹ Pre-rendering chrome-mv3 for development with Vite 6.3.5
# ✗ Build failed in 10ms

# ERROR  [vite:define] Transform failed with 1 error:
# virtual:wxt-background-entrypoint?existing-path-removed/hello'/entrypoints/background.ts:1:59:

# ERROR: Expected ";" but found "/"
# file: virtual:wxt-background-entrypoint?existing-path-removed/hello'/entrypoints/background.ts
```

### System Info

```shell
System:
  OS: Windows 11 10.0.26100
  CPU: (8) x64 Intel(R) Core(TM) Ultra 7 258V
  Memory: 8.21 GB / 31.51 GB
Binaries:
  Node: 24.1.0 - ~\AppData\Local\fnm_multishells\46852_1748823685102\node.EXE
  Yarn: 1.22.22 - ~\AppData\Local\fnm_multishells\46852_1748823685102\yarn.CMD
  npm: 11.3.0 - ~\AppData\Local\fnm_multishells\46852_1748823685102\npm.CMD
  pnpm: 10.11.0 - ~\AppData\Local\fnm_multishells\46852_1748823685102\pnpm.CMD
Browsers:
  Edge: Chromium (131.0.2903.112)
```

### Used Package Manager

pnpm

### Validations

Arktomson:
### Overview

  When a user's project path contains an apostrophe (e.g. /tmp/foo'bar/), building the extension fails because virtual module templates use a simple string replacement that doesn't account for quote characters in file paths. The path gets inserted directly into an import statement like import definition from '/tmp/foo'bar/background.ts', producing invalid JavaScript.

  This fix replaces the naive .replace() with a quote-aware approach: it matches the full quoted import specifier (including surrounding quotes) and replaces it with JSON.stringify(inputPath), which properly escapes any special characters in the path.

  A unit test is added to verify both single-quoted and double-quoted template variants handle paths with apostrophes correctly.

### Manual Testing

  1. Create a project directory with an apostrophe in the path, e.g. mkdir -p "/tmp/foo'bar" and scaffold a WXT project inside it
  2. Run wxt build — previously this would fail with a syntax error
  3. Verify the build now completes successfully

### Related Issue

  This PR closes #1697


packages/wxt/src/core/builders/vite/plugins/__tests__/resolveVirtualModules.test.ts
```diff
@@ -0,0 +1,45 @@
...
+      );
+
+      expect(plugin).toBeDefined();
+
+      const inputPath = `/tmp/foo'bar/background.ts`;
+      const code = await plugin!.load!(
+        '\0virtual:wxt-background-entrypoint?' + inputPath,
+      );
+
+      expect(code).toBe(`import definition from ${JSON.stringify(inputPath)};`);
```
aklinker1: We should hardcode the expected path here so it's clear what's being escaped and so the test doesn't rely on the same API as the runtime code.

Is this correct? Or should there be one more `\`? 

```suggestion
      expect(code).toBe(`import definition from "/tmp/foo\'bar/background.ts";`);
```

When I run it in the browser, I just get `"/tmp/foo'bar/background.ts"` without any escape characters...

```js
> console.log(JSON.stringify(`/tmp/foo'bar/background.ts`))
"/tmp/foo'bar/background.ts"
```

packages/wxt/src/core/builders/vite/plugins/__tests__/resolveVirtualModules.test.ts
```diff
@@ -0,0 +1,45 @@
...
+afterEach(async () => {
+  await Promise.all(tempDirs.splice(0).map((dir) => fs.remove(dir)));
+});
+
+describe('resolveVirtualModules', () => {
+  it.each([
+    `import definition from 'virtual:user-background-entrypoint';`,
+    `import definition from "virtual:user-background-entrypoint";`,
+  ])(
+    'should escape input paths when template contains %s',
```
aklinker1:
```suggestion
    'should escape input paths with apostrophes when encountering: %s',
```

Mention that we're testing apostrophes specifically in this test.

aklinker1 (changes requested):
Thanks for the PR. A few small changes to the tests, then we can merge!

packages/wxt/src/core/builders/vite/plugins/__tests__/resolveVirtualModules.test.ts
aklinker1: Oh wait :facepalm:  I see, the problem is just that the string was defined with single quotes, so having another one in it is invalid:

```ts
'/tmp/foo'bar/background.ts'
```

Can you also add an expected test for when a project path has a double quote `"` in it?

aklinker1:
I did a bit more research here. Should we use `pathToFileURL(absolutePath).href` from `node:url` instead of custom logic? I imagine there are more edge cases that the built-in function handles.

Arktomson:
> Thanks for the PR. A few small changes to the tests, then we can merge!

ok, let me continue editing.

packages/wxt/src/core/builders/vite/plugins/__tests__/resolveVirtualModules.test.ts
Arktomson: ok

aklinker1:
@Arktomson could you also look at https://github.com/wxt-dev/wxt/issues/2146? I think providing this cleanup logic could be a good util for WXT to export as well. But it needs to handle more special characters in file paths to be generic enough.

PatrykKuniczak:
@Arktomson Are you want to go on?

PatrykKuniczak:
@Arktomson One more poke :)

Issue #2020, opened 2025-12-24
Generated script constants


nickbar01234:
### Feature Request

A common pattern is to define a content script and dynamically injecting script at runtime via

```ts
// script.ts
export default defineUnlistedScript(() => { console.log("Hello" });

// background.ts
browser.scripting.executeScript({ file: ["/script.js"] })
```

If the filename were to change, the user would also need to update the path used in `executeScript`. Curious on whether it's feasible to generate a constant enum specifying the filename. The API can look something like

```ts
export default defineUnlistedScript({
  name: "MyDynamicScript"
  main: () => { ... }
}

// background.ts
browser.scripting.executeScript({ file: [Scripts["MyDynamicScript"]] });
```

the value would map to the actual path.

#### Is your feature request related to a bug?

N/A

aklinker1:
I like this! We don't need a separate `name` option in `defineUnlistedScript` or other entrypoints definition functions, WXT already has a name it assigned to entrypoints internally based on the file name.

If someone wants to contribute this feature, that would be great! This can be accomplished via an ["built-in WXT module"](https://github.com/wxt-dev/wxt/tree/main/packages%2Fwxt%2Fsrc%2Fbuiltin-modules) that generates a file in the `.wxt directory and adds an import alias and adds the constants to auto-imports as well.

Otherwise I'll prioritize working on it myself after v1.0 is out.

nickbar01234:
nice! I wasn't sure if it was accomplishable at first. Happy to take a stab at it!

aklinker1:
Sounds good, @ me if you have a question or get stuck. I can provide more direction as well if you don't know where to start.

nickbar01234:
Hey @aklinker1, started taking a look at this.

> We don't need a separate name option in defineUnlistedScript or other entrypoints definition functions, WXT already has a name it assigned to entrypoints internally based on the file name.

Can you clarify why the internal naming is helpful in this case? If I'm reading the code correctly, the internal name is dependent on the file name, so even if we add auto-import, subsequent file name changes would still break the build. 

```ts
export interface BaseEntrypoint {
  /**
   * The entrypoint's name. This is the filename or dirname without the type suffix.
   *
   * Examples:
   * - `popup.html` &rarr; `popup`
   * - `options/index.html` &rarr; `options`
   * - `named.sandbox.html` &rarr; `named`
   * - `named.sandbox/index.html` &rarr; `named`
[...]

PR #2149, opened 2026-02-21
feat(modules): Add support for augumenting entrypoint options

aklinker1:
### Overview

This allows modules to augment WXT's types and add their own config options to entrypoints. See the docs change for an example.

It also introduces the ability to find `<meta name="wxt.<name>" content="..." />` params in HTML entrypoints, adding them to the entrypoint's option internally.

### Manual Testing

See unit test changes.

### Related Issue

Related to #2020 - If the module discussed there needs to add custom options.

### DIFF
--- a/docs/guide/essentials/wxt-modules.md
+++ b/docs/guide/essentials/wxt-modules.md
@@ -137,6 +137,58 @@ console.log(config.myModule);
 
 This is very useful when [generating runtime code](#generate-runtime-module).
 
+#### Add custom entrypoint options
+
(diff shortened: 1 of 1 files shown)


packages/wxt/src/types.ts
```diff
@@ -684,6 +684,7 @@ export interface PopupEntrypointOptions extends BaseEntrypointOptions {
 }
 
 export interface OptionsEntrypointOptions extends BaseEntrypointOptions {
+  title?: string;
```
aklinker1, line 687: All HTML entrypoints get this, without it a test had a type error.

PR #2150, opened 2026-02-22
fix(types): include CSS entrypoints in PublicPath generation

dahomita:
### Overview

I picked this up as a follow-up to the discussion in #2142.

The goal is to ensure `PublicPath` typing matches actual generated outputs by including CSS entrypoints, so `browser.runtime.getURL` type checks are accurate for CSS assets as well as JS/HTML.

This updates path generation to map:
- `content-script-style` -> `.css`
- `unlisted-style` -> `.css`
- everything else unchanged

I also added an e2e test in `typescript-project.test.ts` to assert CSS paths are present in generated `.wxt/types/paths.d.ts` (`/plain.css` and `/content-scripts/overlay.css`).

### Manual Testing
- [x] `pnpm test`
- [x] I manually smoke-tested this in `templates/vanilla` by adding css entrypoints and confirming they were included in `PublicPath`

### Related Issue

### DIFF
--- a/.github/workflows/auto-label.yml
+++ b/.github/workflows/auto-label.yml
@@ -1,8 +1,8 @@
 name: ✨ Auto-label PR
 
 on:
-  pull_request:
-    types: [opened, synchronize]
+  pull_request_target:
+    types: [opened, synchronized, reopened]
 
 jobs:
   update-pr:


PatrykKuniczak:
@dahomita But that's good idea to include all `.css` files to be accessible inside content script?

dahomita:
> @dahomita But that's good idea to include all `.css` files to be accessible inside content script?

Good question @PatrykKuniczak ! This PR only updates TypeScript path generation (PublicPath) so valid CSS output paths are typed correctly for `browser.runtime.getURL.` It does not change `web_accessible_resource`s behavior

PatrykKuniczak:
@dahomita Yeah, i've forgot about `web_accessible_resource`

PR #2154, opened 2026-02-22
ci: Debug incorrect Linux test passes

aklinker1:
### Overview

Linux tests are passing when they shouldn't be. Windows tests has saved the PR checks by throwing errors correctly, but I need to investigate and fix linux checks.


aklinker1:
OK, if the failed test is in `is-background`, it fails checks correctly.

aklinker1:
Weird.

PR #2170, opened 2026-03-02
feat: add `vitePluginsBefore` option

mrtnbroder:
this allows plugins like `@tanstack/router-plugin` to be added before the react plugin to work correctly

### Overview

When using e.g. `pnpm install` together with `wxt` and `@tanstack/router-plugin` I got the following error:

```
 ERROR  Plugin order error: '@vitejs/plugin-react' is placed before '@tanstack/router-plugin'.
│ The TanStack Router plugin must come BEFORE JSX transformation plugins.
│ Please update your Vite config:
│   plugins: [
│     tanstackRouter(),
│     react(),
│   ]
```

To fix this issue, the easiest solution was to add a `vitePluginsBefore` field and add the plugin there.

### Manual Testing

Add the plugin to `vitePluginsBefore` and see no error raised.

### Related Issue

~~This PR closes #<issue_number>~~

### DIFF
--- a/packages/module-react/package.json
+++ b/packages/module-react/package.json
@@ -45,19 +45,20 @@
     "prepare": "buildc --deps-only -- wxt prepare"
   },
   "peerDependencies": {
-    "wxt": ">=0.19.16",
-    "vite": "^7.0.0"
+    "vite": "^5.4.19 || ^6.3.4 || ^7.0.0",
+    "wxt": ">=0.19.16"
   },
   "dependencies": {
     "@vitejs/plugin-react": "^4.4.1 || ^5.0.0"
   },
   "devDependencies": {
-    "@types/react": "^19.2.13",
     "@types/react-dom": "^19.2.3",
+    "@types/react": "^19.2.13",
     "publint": "^0.3.17",
-    "react": "^19.2.4",
     "react-dom": "^19.2.4",
+    "react": "^19.2.4",
     "typescript": "^5.9.3",
+    "vite": "^7.0.0",
     "wxt": "workspace:*"
   }
 }
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -277,6 +277,9 @@ importers:
       typescript:
         specifier: ^5.9.3
         version: 5.9.3
+      vite:
+        specifier: ^7.0.0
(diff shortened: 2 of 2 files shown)


packages/module-react/package.json
```diff
@@ -57,6 +57,7 @@
     "react": "^19.2.4",
     "react-dom": "^19.2.4",
     "typescript": "^5.9.3",
+    "vite": "^5.4.19 || ^6.3.4 || ^7.0.0",
```
aklinker1: Move this to `peerDependencies`, and just put `"vite": "^7.0.0",` in the dev dependencies.

mrtnbroder: Done! Btw this is **only** for the react-module for now, should I also add it to the others? (svelte, vue etc) so it stays consistent?

aklinker1: No, add this only for react.

In general, vite plugins shouldn't need to be ordered by hand, it's really weird that this one does. I would think they should just use the either [`enforce: "pre"`](https://vite.dev/guide/api-plugin#plugin-ordering) or [hook ordering](https://rollupjs.org/plugin-development/#build-hooks) instead of requiring devs to order their plugin list correctly.

But hey, maybe they tried those already and manual ordering is what worked for them 🤷

packages/module-react/package.json
```diff
@@ -45,7 +45,8 @@
     "prepare": "buildc --deps-only -- wxt prepare"
   },
   "peerDependencies": {
-    "wxt": ">=0.19.16"
+    "wxt": ">=0.19.16",
+    "vite": "^7.0.0"
```
aklinker1: Not quite. We need to add two vite entries in the package.json: 

1. `"vite": "^5.4.19 || ^6.3.4 || ^7.0.0"` in `peerDependencies` 
2. `"vite": "^7.0.0"` in `devDependencies`

This tells consumers of the package that they need Vite 5, 6, or 7 as a peer, but developers of the package use vite 7 to build the module. Does that make sense?

aklinker1 (changes requested):

packages/module-react/package.json
mrtnbroder: Well it is set as `pre`, but this module is added first and `plugin-react` also has `enforce: "pre"`, therefore the order is "wrong" :x

packages/module-react/package.json
mrtnbroder: Sure, fixed!

PR #2181, opened 2026-03-06
feat: replace `dotenv` with native `node:util` `parseEnv`

### LINKED ISSUE #2176: Replace `dotenv` with builtins
### Feature Request

See Vite's `loadEnv` imlementation - they use `dotenv-expand` and `node:util#parseEnv`.

https://github.com/vitejs/vite/blob/a704abc4e6f1528837030cf92cb7a2de65da54cc/packages/vite/src/node/env.ts

Added in 20.12:

https://nodejs.org/api/util.html#utilparseenvcontent

HarrysonLadines:
### Overview

This PR replaces the dotenv dependency with Node.js native node:util#parseEnv implementation.

- Dependency reduction: Removed dotenv from package.json, reducing the project's dependency footprint.
- Modern API integration: Refactored loadEnv to use node:fs and node:util, aligning WXT with the environment loading strategy used by Vite.
- Priority Handling: Maintained the existing priority logic where specific files (like .local) override general ones using a robust .reverse() and Object.fromEntries approach.
- Compatibility: Kept dotenv-expand to ensure support for variable expansion (e.g., VAR=$OTHER_VAR) remains functional for users.

### Manual Testing

I verified the implementation using a standalone TypeScript script within the repository to confirm the native `parseEnv` behavior:

1. Environmental Parsing: Verified that `node:util#parseEnv` correctly maps key-value pairs from `.env` files into a JavaScript object.
2. Priority Logic: Confirmed that reversing the `envFiles` array before calling `Object.fromEntries` ensures that specific files (like `.env.local`) correctly override base configurations.
3. Execution Environment: Successfully validated the logic using `bun` to ensure compatibility with the project's current development workflow.
4. Expansion Support: Confirmed that `dotenv-expand` remains functional for nested variables after the native parsing stage.

### Related Issue

This PR closes #2176

### DIFF
--- a/packages/wxt/src/core/utils/env.ts
+++ b/packages/wxt/src/core/utils/env.ts
@@ -6,36 +6,36 @@ import type { TargetBrowser } from '../../types';
 /** Load environment files based on the current mode and browser. */
 export function loadEnv(mode: string, browser: TargetBrowser) {
   const envFiles = [
-    `.env.${mode}.${browser}.local`,
-    `.env.${mode}.${browser}`,
-    `.env.${browser}.local`,
-    `.env.${browser}`,
-    `.env.${mode}.local`,
-    `.env.${mode}`,
-    `.env.local`,
+    // List is ordered with general files first, specific ones last, so the more
+    // specific files override the more general ones in the loop below.
     `.env`,
+    `.env.local`,
+    `.env.${mode}`,
+    `.env.${mode}.local`,
+    `.env.${browser}`,
+    `.env.${browser}.local`,
+    `.env.${mode}.${browser}`,
+    `.env.${mode}.${browser}.local`,
   ];
 
-  // We reversed the list so that more specific files take priority over more general ones.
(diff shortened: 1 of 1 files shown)


HarrysonLadines:
I've updated the pnpm-lock.yaml and amended the commit to sync dependencies. The previous deploy failure should be resolved now.

aklinker1 (approved):
Nice, looks good. Checks might fail due to formatting, run `pnpm prettier --write` if it does.

Edit: I had some issues with checks on `main`, they have been fixed. Looks like you have formatting and some type errors - run `pnpm check` to auto-fix them and see the type errors.

aklinker1:
~~Hey, as mentioned in #2185, this is only available in Node 22. Can you change this PR to be based off `major` instead so it is included as a breaking change?~~

Actually, double checked, this was added in 20.12, so it's fine:

image

https://nodejs.org/api/util.html#utilparseenvcontent

PR #2185, opened 2026-03-07
chore(deps): replace dotenv with Node.js builtins

### LINKED ISSUE #2176: Replace `dotenv` with builtins

Xavrir:
Drop `dotenv` in favor of Node's built-in `node:util#parseEnv` (22+), with a lightweight polyfill for older runtimes. Follows the same approach Vite uses in their `loadEnv`.

- New `parse-env.ts` utility: tries native `parseEnv` at load time, falls back to a polyfill that matches native semantics (`export` prefix, inline `#` comments, quote handling)
- `loadEnv` stays synchronous — no async infection
- Fixed env-file precedence: files are now loaded least-specific → most-specific so that `.env.production.chrome.local` correctly overrides `.env`
- Kept `dotenv-expand` for variable expansion

Ref: [Vite's loadEnv](https://github.com/vitejs/vite/blob/a704abc4e6f1528837030cf92cb7a2de65da54cc/packages/vite/src/node/env.ts)

Closes #2176

### DIFF
--- a/packages/wxt/src/core/utils/parse-env.ts
+++ b/packages/wxt/src/core/utils/parse-env.ts
@@ -23,37 +23,40 @@ function parseEnvPolyfill(input: string): Record<string, string> {
   const result: Record<string, string> = {};
 
   for (const line of input.split('\n')) {
-    const trimmed = line.trim();
-    if (!trimmed || trimmed.startsWith('#')) continue;
+    let entry = line.trim();
+    if (!entry || entry.startsWith('#')) continue;
 
-    const eqIndex = trimmed.indexOf('=');
-    if (eqIndex === -1) continue;
-
-    const key = trimmed.substring(0, eqIndex).trim();
-    if (!key) continue;
-
-    let value = trimmed.substring(eqIndex + 1).trim();
-
-    if (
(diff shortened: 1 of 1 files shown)


PatrykKuniczak:
Good work :)

Maybe we can enforce users to use 22+ here:
https://github.com/wxt-dev/wxt/blob/e4598a5e3789236f08120d445f46020947865c1e/package.json#L5

IMO it's good idea, because anyway the latest LTS is 24 and we're about to have 26 as newest LTS.

And we can avoid your fallback.
But @aklinker1 need to support that idea

aklinker1:
Ugh, so it's only node 22+... Can you change this PR to merge into the `major` branch? We'll have to do this in a breaking change.

aklinker1:
Actually, sorry, noticed this was a duplicate PR... Closing in favor of #2181

PR #2186, opened 2026-03-07
chore(deps): replace minimatch with picomatch

### LINKED ISSUE #2177: Replace `minimatch` with `picomatch`
### Feature Request

Tinyglob already relies on picomatch, so it makes sense to replace `minimatch` with it. Picomatch has zero deps.

Xavrir:
Swap `minimatch` for `picomatch` — zero deps and already in the tree via tinyglobby.

- Replace imports in `minimatch-multiple.ts` and `find-entrypoints.ts`
- Add `picomatch` as a direct dependency (was only transitive before)
- Drop `minimatch` from package.json and lockfile

Closes #2177

### DIFF
--- a/packages/wxt/package.json
+++ b/packages/wxt/package.json
@@ -44,6 +44,7 @@
     "linkedom": "^0.18.12",
     "magicast": "^0.5.2",
     "nano-spawn": "^2.0.0",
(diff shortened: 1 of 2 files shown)


aklinker1:
Closing in favor of #2188

PR #2188, opened 2026-03-08
chore: replace minimatch with picomatch

### LINKED ISSUE #2177: Replace `minimatch` with `picomatch`

Adiii-15:
### Overview

replaced refs of minimatch to picomatch. Only difference is minimatch had a native TS support while picomatch is JS-only here and doesn’t provide .d.ts. 

### Manual Testing

ran tests

### Related Issue

This PR closes #2177

### DIFF
--- a/packages/wxt/src/@types/modules.d.ts
+++ b/packages/wxt/src/@types/modules.d.ts
@@ -32,14 +32,3 @@ declare module 'web-ext-run/util/logger' {
   }
   export const consoleStream: IConsoleStream;
 }
-
-declare module 'picomatch' {
-  export interface PicomatchOptions {
-    [key: string]: unknown;
-  }
-
-  export default function picomatch(
-    glob: string | readonly string[],
-    options?: PicomatchOptions,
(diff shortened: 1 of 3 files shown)


packages/wxt/src/@types/modules.d.ts
```diff
@@ -32,3 +32,14 @@ declare module 'web-ext-run/util/logger' {
...
+declare module 'picomatch' {
+  export interface PicomatchOptions {
+    [key: string]: unknown;
+  }
+
+  export default function picomatch(
+    glob: string | readonly string[],
+    options?: PicomatchOptions,
+  ): (input: string) => boolean;
+}
```
aklinker1: Shouldn't we just install `@types/picomatch`?

aklinker1: I've installed it and pushed to this branch.

PR #2189, opened 2026-03-11
fix: allow user-defined sidebar_action to override generated fields

djyde:
## Summary
- Fixed an issue where the `sidebar_action` manifest field for Firefox was being completely overwritten by WXT, ignoring user-defined configuration from `wxt.config.ts`
- User-defined `sidebar_action` fields now properly override auto-generated values
- Added test case to verify the merge behavior

## Changes
- Modified `packages/wxt/src/core/utils/manifest.ts` to spread user-defined `sidebar_action` configuration after generated fields, allowing user values to take precedence
- Added test case in `packages/wxt/src/core/utils/__tests__/manifest.test.ts` to verify user configuration is properly merged

## Test Plan
- All existing tests pass (76 tests)
- New test case verifies that user-defined fields override generated values
- Custom fields (like `custom_field`) are preserved
- Auto-generated `default_panel` is still included


aklinker1:
You should be defining sidepanel options in the entrypoint, not the manifest, like how you define options for all other HTML entrypoints in WXT. WXT should automatically convert between side panel and sidebar actions when targeting firefox.

https://wxt.dev/guide/essentials/entrypoints.html#side-panel

I'm not going to merge this PR because I only want one way to customize the manifest options for sidebars. LMK if there was a bug this fixed that the existing way of configuring the side panel doesn't support. I'll gladly discuss the correct way to fix that problem.

PR #2201, opened 2026-03-18
chore: Remove `any` from `i18n`'s internal types

aklinker1:
### Overview

Pulled the internal changes out of #2121 so they can be merged while we discuss the other changes in that PR.

CC @PatrykKuniczak

### Manual Testing

N/A - See PR checks

### Related Issue

- #2121

### DIFF
--- a/packages/i18n/src/build.ts
+++ b/packages/i18n/src/build.ts
@@ -26,6 +26,15 @@ export interface ChromeMessage {
   placeholders?: Record<string, { content: string; example?: string }>;
 }
 
+export type Message =
+  | SimpleMessage
+  | PluralMessage
+  | ChromeMessage
+  | Message[]
+  | { [key: string]: Message };
+
+export type MessagesObject = Record<string, Message>;
+
 export interface ParsedBaseMessage {
   key: string[];
   substitutions: number;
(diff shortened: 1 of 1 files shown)


packages/i18n/src/build.ts
```diff
@@ -16,12 +16,25 @@ export { SUPPORTED_LOCALES } from './supported-locales';
...
   description?: string;
   placeholders?: Record<string, { content: string; example?: string }>;
 }
 
+export type Message =
+  | SimpleMessage
+  | PluralMessage
+  | ChromeMessage
+  | Message[]
+  | { [key: string]: Message };
```
aklinker1, line 34: This was your `MessageType` type, but it didn't cover all the different types of messages. Now it does.

I also remove the `Type` suffix since it's already a type.

packages/i18n/src/build.ts
```diff
@@ -16,12 +16,25 @@ export { SUPPORTED_LOCALES } from './supported-locales';
...
+
+export type MessagesObject = Record<string, Message>;
```
aklinker1, line 36: Created a separate type for the messages file contents, referred to as `MessagesObject` further down (see existing `parseMessagesObject` function). Couldn't think of a better name for this.

PR #2208, opened 2026-03-23
docs: Add Firefox Add-on support to homepage showcase

shadoath:
## Summary

- Add optional `firefoxSlug` to extension entries (Chrome-only, Firefox-only, or both)
- Fetch Chrome and Firefox data in a single GraphQL request using the unified `Extension` type
- Combine user counts into one number per card, with store links (Chrome, Firefox) in the footer

## Screenshot

Screenshot 2026-04-12 at 8 33 58 AM


InvictusNavarchus:
I support the idea, since I also publish on Firefox Addons, but having Firefox URL optional while Chrome Extension URL required is still biased toward Chrome Extensions. What if you only publish the extension(s) on Firefox? You can't showcase it, then.

Due to smaller user base of Firefox, it actually serves as a great real testing ground with real users before finally deploying it to Chrome. So some people, including me, may publish it there first.

shadoath:
I'll make that update. Also going to scrap the extensions that are already there and add the firefox slug if I can find it.

shadoath:
Only used "K" and not "m" to give the number some weight, when it was 2,000k, it seemed more impressive than 2m. Also, this change helped the stats sit on a single line, making the display cleaner.

shadoath:
@aklinker1 what do you think about firefox addition?

aklinker1:
Reviewing this soon! Firefox is my daily driver, so I'm down to add this for sure.

InvictusNavarchus:
> Only used "K" and not "m" to give the number some weight, when it was 2,000k, it seemed more impressive than 2m. Also, this change helped the stats sit on a single line, making the display cleaner.

Using the same logic, we can push it further by not doing any unit conversion at all, and let the raw count speak for itself (e.g., 200.000 users). This is currently the case for both Firefox Addons and Chrome Web Store. I'd suggest keeping it that way. 

Also, this is an entirely a separate change unrelated to the main topic of this PR. I think it's better to propose this in a different PR

aklinker1:
If we want to shorten the numbers, we should use `num.toLocaleString(undefined, { notation: "compact"})` and follow the locale's display preference. For me, that's `1234567` &rarr; `1.2M`. I'm good with that or the full number. If we have room, I'd probably prefer the full number.

Maybe we combine the firefox and chrome stats into a single number and just add links to each of the stores? That would simplify the UI a bit and give us more room to show the full number.

*Edit: Chrome reports a weekly user count vs Firefox's daily user count. The current PR ignores that, which is fine. It would also be fine to ignore the units when adding the two numbers together.*

docs/.vitepress/composables/useFirefoxAddonDetails.ts
aklinker1, line 1: The GraphQL API already supports getting firefox addons. I would recommend querying all of the chrome and firefox addons in a single request to the graphql API so we don't have to deal with multiple loading states. This would simplify the PR quite a bit.

Here's an example query:

```gql
{
  chromeExtensions(ids: ["mgmdkjcljneegjfajchedjpdhbadklcf"]) {
    ...ExtensionData
  }
  firefoxAddons(ids: ["anime-skip"]) {
    ...ExtensionData
  }
}

fragment ExtensionData on Extension {
  id
  name
  iconUrl
  shortDescription
  storeUrl
  rating
  users
}
```

> I just deployed a new version where the `ChromeExtension` and `FirefoxExtension` types implement a new `Extension` type that has a `users` field, instead of `weeklyActiveUsers` or `dailyActiveUsers`
>
> https://github.com/wxt-dev/queue/releases/tag/v0.4.4

We should be able to represent all results as a single type now too.

docs/.vitepress/components/UsingWxtSection.vue
```diff
@@ -1,174 +1,421 @@
...
 
 function getStoreUrl(extension: ChromeExtension) {
   const url = new URL(extension.storeUrl);
   url.searchParams.set('utm_source', 'wxt.dev');
   return url.href;
 }
+
+function primaryStoreUrl(extension: ListedExtension) {
+  return extension.chromeStoreUrl ?? extension.firefoxUrl ?? '#';
+}
```
aklinker1: Maybe we remove the link on the app name and only add links in the bottom row to each store?

aklinker1 (changes requested):
If you merge with main, the failing checks will pass and your PR will be deployed to preview.

Only real concern is around simplifying the data fetching by combining the firefox and chrome requests into a single graphql call. Then the whole UI could be simplified by combining the results into a single list of standardized objects before rendering them.

We should probably also add support for Edge, but I need to add that to the graphql API first as well. But ideally, this PR adds support for any number of extension stores in a generic way.

docs/.vitepress/composables/useFirefoxAddonDetails.ts
aklinker1: I've now added edge, but we can do that in a separate PR: 

```gql
{
  chromeExtensions(ids: ["mgmdkjcljneegjfajchedjpdhbadklcf"]) {
    ...Data
  }
  firefoxAddons(ids: ["anime-skip"]) {
    ...Data
  }
  edgeAddons(ids: ["aghmglalnfmjbipodadgbgpakneabgnp"]) {
    ...Data
  }
}

fragment Data on Extension {
  id
  name
  iconUrl
  shortDescription
  storeUrl
  rating
  users
}
```

docs/.vitepress/components/UsingWxtSection.vue
shadoath: I tried this out, but I think the link in the app name has a lot of value and is easy to see/use
Defaulted it to Chrome, with a fallback to Firefox. In the future, we could let authors choose whether they have both.

shadoath:
@InvictusNavarchus & @aklinker1. I took your comments into consideration and pushed up the simplified new version.

Let me know what you both think.

shadoath:
@aklinker1 can we add the Firefox slug field to the Extension type query?

aklinker1 (approved):

aklinker1:
Nice, much simpler, looks great too!

> @aklinker1 can we add the Firefox slug field to the Extension type query?

Sure, I can add that!

aklinker1:
Is this what you wanted? https://github.com/wxt-dev/queue/commit/3ce60f31678155b6580555257fa64030fed469eb#diff-70f27b1705e6c4424067f93d1d897f8f48711869f34688493bf0d16b1e8d0f1e

It's deployed

shadoath:
The slug option will clean up some logic here. If that is done first.
But we don’t need to wait that change can be another PR.

docs/.vitepress/components/UsingWxtSection.vue
```diff
@@ -1,174 +1,655 @@
...
+  if (!data.value) return [];
+
+  const chromeById = new Map(
+    (data.value.chrome ?? []).filter(Boolean).map((e) => [e.id, e]),
+  );
+  const firefoxBySlug = new Map(
+    (data.value.firefox ?? []).filter(Boolean).map((e) => {
+      const match = e.storeUrl.match(/\/addon\/([^/]+)/);
+      const slug = match ? match[1] : e.id;
+      return [slug, e];
```
aklinker1: I see, here? The changes I deployed should be able to clean this up, yes.

shadoath: Ops didn't see that it was already done. That was FAST!

shadoath: @aklinker1 pushed up the change to remove that regex

aklinker1: Oh sorry, I only added it to the `FirefoxAddon` type, you can't add it to the `Extension` fragment like that.

aklinker1:
Sweet, gonna merge, then move the list to a YML file outside the vue component lol. It's gotten quite large.

aklinker1:
Having problems with github actions... one sec...

PR #2212, opened 2026-03-25
fix: use git ls-files for sources zip to respect .gitignore

creeperkatze:
### Overview

Sources ZIP uses tinyglobby which traverses the full directory tree before filtering, causing it to hang or balloon when gitignored directories like Rust's target/ exist locally. Switch to git ls-files --recurse-submodules for the sources ZIP since it should only contain what git tracks anyway. Falls back to the existing glob behavior if git is unavailable.

### Manual Testing

1. Have a project with large gitignored directories (e.g. a built Rust submodule with a target/ dir)
2. Run wxt zip -b firefox
3. Before: Produces a bloated ZIP. After: Completes quickly with only git-tracked files.

### DIFF
--- a/packages/wxt/src/core/zip.ts
+++ b/packages/wxt/src/core/zip.ts
@@ -77,7 +77,6 @@ export async function zip(config?: InlineConfig): Promise<string[]> {
     await zipDir(wxt.config.zip.sourcesRoot, sourcesZipPath, {
       include: wxt.config.zip.includeSources,
       exclude: excludeSources,
-      useGit: true,
       transform(absolutePath, zipPath, content) {
         if (zipPath.endsWith('package.json')) {
           return addOverridesToPackageJson(absolutePath, content, overrides);
@@ -107,7 +106,6 @@ async function zipDir(
   options?: {
     include?: string[];
     exclude?: string[];
-    useGit?: boolean;
     transform?: (
       absolutePath: string,
       zipPath: string,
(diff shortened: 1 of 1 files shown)


PatrykKuniczak:
@aklinker1 Are you like it?

creeperkatze:
Will look into the failed tests, probably has something to do with it not using the fallback when git isnt used

aklinker1:
There are reasons to include got ignored files, so this isn't going to work.

There's a different PR (#2114) already merged into the `major` branch that changes the glob behavior. Right now it includes everything by default, causing the slowdown you mentioned here. In that PR, we switched it so you have to list directories to include manually, which fixes the performance issue.

creeperkatze:
Okay, tho wouldn't explicitly listing directories quickly get unmanageable with complex directory structures like git submodules?

creeperkatze:
For example i'm developing an extension that uses private packages from a github monorepo by including it as a submodule, since it uses ui components from that project.

aklinker1:
I don't think it would get out of control, no. At least no more than what the lists look like today. A few basic includes and some excludes, then you're good to go.

I think the ideal solution here is to actually use the build output metadata to detect all the files that are needed for the sources zip. See https://github.com/wxt-dev/wxt/issues/1505

aklinker1:
In the meantime, if `git ls` does work for you, you can put your logic in a wxt module and automatically configure those options rather than maintaining a list yourself.

Or just run the logic in the top-level of your wxt config file.

creeperkatze:
Mhm, gonna look into that. Im currently just using a patch.

PR #2229, opened 2026-04-03
Security: Build-time remote code download lacks integrity verification

tuanaiseo:
## Problem

The `url:` import mechanism downloads arbitrary remote JavaScript and bundles it into the extension build. This creates a supply-chain risk if the remote source is compromised or unexpectedly changed, since no checksum/signature verification is enforced before inclusion.

**Severity**: `medium`
**File**: `packages/wxt/src/core/builders/vite/plugins/download.ts`

## Solution

Require integrity metadata (e.g., SHA-256) for `url:` imports, verify content hash before caching/bundling, and optionally enforce an allowlist of approved domains.

## Changes

- `packages/wxt/src/core/builders/vite/plugins/download.ts` (modified)

## Testing

- [ ] Existing tests pass
- [ ] Manual review completed
- [ ] No new warnings/errors introduced

### DIFF
--- a/packages/wxt/src/core/builders/vite/plugins/download.ts
+++ b/packages/wxt/src/core/builders/vite/plugins/download.ts
@@ -1,3 +1,4 @@
+import { createHash } from 'node:crypto';
 import type { Plugin } from 'vite';
 import type { ResolvedConfig } from '../../../../types';
 import { fetchCached } from '../../../utils/network';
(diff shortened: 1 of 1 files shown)


PatrykKuniczak:
@tuanaiseo How it's possible to make supply-chain attack on this code, without sha?

aklinker1:
If someone wasn't using a fixed version, or the CDN was hacked, it could happen. I do think adding a hash could be useful.

That said, for google analytics for example, there is no version or way to create a hash. So we can't require it and there are valid cases where we shouldn't warn about it either. So I'm not sure we can merge this one.

PR #2232, opened 2026-04-05
fix: Scripts injection for content-scripts on Firefox MV2 target

### LINKED ISSUE #2183: `injectScript` never resolves in MV2
### Describe the bug

I used `injectScript` according to the documentation below. While it resolves correctly in MV3, it never resolves in MV2.

https://wxt.dev/guide/essentials/content-scripts.html#isolated-world-vs-main-world

### Reproduction

[reproduction.zip](https://github.com/user-attachments/files/25815812/reproduction.zip)

### Steps to reproduce

1. Run `pnpm install`
2. Run `pnpm dev:firefox`
3. Load the extension into Firefox.
4. Navigate to `https://www.google.com/`.
5. Open the developer tools console and verify that `Done!` is not logged.

### System Info

```shell
System:
    OS: Linux 6.6 Ubuntu 24.04.4 LTS 24.04.4 LTS (Noble Numbat)
    CPU: (24) x64 13th Gen Intel(R) Core(TM) i7-13700F
    Memory: 20.96 GB / 23.38 GB
    Container: Yes
    Shell: 5.2.21 - /bin/bash
  Binaries:
    Node: 24.14.0 - /home/nines/.volta/tools/image/node/24.14.0/bin/node
    npm: 11.11.0 - /home/nines/.volta/tools/image/npm/11.11.0/bin/npm
    pnpm: 10.30.3 - /home/nines/.volta/tools/image/pnpm/10.30.3/bin/pnpm
  npmPackages:
    wxt: ^0.20.18 => 0.20.18
```

### Used Package Manager

pnpm

### Validations

PatrykKuniczak:
### Overview

I've fixed issue with awaited injectScript for Firefox MV2, which isn't included in build files.

### Manual Testing

Let's try code from mentioned issue reproduction

This PR closes #2183

Code with example for that'll be merged in #2233


nines75:
@PatrykKuniczak My explanation at https://github.com/wxt-dev/wxt/issues/2183#issuecomment-4192691657 was insufficient. With the current code, `injectScript` will never resolve in MV3.
What I meant was to separate the creation of the Promise via `makeLoadedPromise` from the timing of using `await` on it, just like the `injectScript` code prior to this PR.

Current code in the PR:

```ts
if (isManifestV3) {
  // Event handlers (load, error) are registered before the script execution.
  // However, since `await` is used here, it will wait forever for the `load` event to fire.
  await makeLoadedPromise(script);
}

await options?.modifyScript?.(script);

(document.head ?? document.documentElement).append(script);

if (!options?.keepInDom) {
  script.remove();
}
```

Previous code (current main branch):

```ts
// Event handlers are registered before the script execution.
// Since `await` is not used, execution doesn't pause here.
const loadedPromise = makeLoadedPromise(script);

await options?.modifyScript?.(script);

// At this point, the event handlers are guaranteed to be registered.
(document.head ?? document.documentElement).append(script);

if (!options?.keepInDom) {
  script.remove();
}

// Wait for the `load` event to fire here.
// However, the `load` event doesn't fire in MV2.
await loadedPromise;
```

Based on the above, I think it would be better to implement the fix as follows:

```ts
// Register the event handlers only for MV3.
const loadedPromise = isManifestV3 ? makeLoadedPromise(script) : undefined;

await options?.modifyScript?.(script);

(document.head ?? document.documentElement).append(script);

if (!options?.keepInDom) {
  script.remove();
}

if (isManifestV3) {
  await loadedPromise;
}
// This alternative might be better:
// if (loadedPromise) {
//   await loadedPromise;
// }
```

I'm not entirely sure if this code is perfectly correct, but I have confirmed that it works properly on Firefox MV2 and Chrome MV3.

PatrykKuniczak:
@nines75 I was wondering, if this is necessary to do it that way, because it work's anyway.

But i've checked it, and
```ts
const loadedPromise = isManifestV3 ? makeLoadedPromise(script) : Promise.resolve();
```

works as well, we need to wait for @aklinker1 review, if we'll approve your suggestion, i'll do it, that's no problem :)

I'm not sure, my simplification is good approach 😆

packages/wxt/src/utils/inject-script.ts
```diff
@@ -26,15 +26,22 @@ export async function injectScript(
   const url = browser.runtime.getURL(path);
   const script = document.createElement('script');
 
-  if (browser.runtime.getManifest().manifest_version === 2) {
+  const isManifestV2 = browser.runtime.getManifest().manifest_version === 2;
+  const isManifestV3 = browser.runtime.getManifest().manifest_version === 3;
```
aklinker1:
```suggestion
  const manifest = browser.runtime.getManifest()
  const isManifestV2 = manifest.manifest_version === 2;
  const isManifestV3 = manifest.manifest_version === 3;
```

packages/wxt/src/utils/inject-script.ts
```diff
@@ -26,15 +26,22 @@ export async function injectScript(
...
     // MV3 requires using src
     script.src = url;
   }
 
-  const loadedPromise = makeLoadedPromise(script);
+  // For MV2: inline scripts execute synchronously when appended, so we don't need to wait
+  // For MV3: we need to wait for the load event
+  if (isManifestV3) {
+    await makeLoadedPromise(script);
+  }
```
aklinker1: We can't await this here. We need to create the promise before adding the script to the DOM, then await it after, like we were doing before.

```suggestion
  // For MV2: inline scripts execute synchronously when appended, so we don't need to wait
  // For MV3: we need to wait for the load event
  const loadedPromise = isManifestV3 ? makeLoadedPromise(script) : undefined;
```

Then add back the `await loadedPromise` at the bottom of the function, or wrap it in a similar if-statement.

Is this the discussion you were having above? You said this works, but that can't be true.

aklinker1 (changes requested):

packages/wxt/src/utils/inject-script.ts
aklinker1: Yeah, I just merged https://github.com/wxt-dev/wxt/pull/2233 and updated this branch to test against it. The current code awaits forever in MV3, doesn't work.

We need to keep awaiting it after the element is added to the DOM.

PatrykKuniczak: > Yeah, I just merged #2233 and updated this branch to test against it. The current code awaits forever in MV3, doesn't work.
> 
> We need to keep awaiting it after the element is added to the DOM.

No i see it, interesting because i've checked it before, maybe cache worked 🤔 
Thanks for second look, i'll fix it

PatrykKuniczak: @aklinker1 Should be good right now, let's check it one more time 😄

aklinker1: Why change it to `async` false?

aklinker1: Claude doesn't think setting `async` to false fixes the race condition around not adding the listeners before adding it to the dom.

aklinker1 (approved):
Confirmed its working with the latest changes.

packages/wxt/src/utils/inject-script.ts
aklinker1: I've made the change.

PatrykKuniczak:
@aklinker1 Yeah, that one also works.
I've previously `Promise.resolve()` on the `undefined` place, but i don't know why it wasn't work.
But it's good right now and let's leave it what it is 😄

PR #2233, opened 2026-04-05
chore: Create `content.ts` for `wxt-demo` as example

PatrykKuniczak:
### Overview

Added new file with tests, to show #2232 really fix the problem.

### Manual Testing

Let's check if everything is shown in console for each browser target in `packages/wxt-demo`

Blocked by #2232


PatrykKuniczak:
Really fu.. those tests, it's too complicated for that simply case, example code'll be enough.

It's doable, but anyway MV2 support will end soon also for Firefox and then we'll need to support only MV3.

aklinker1 (approved):

PR #2241, opened 2026-04-11
feat: type `/_favicon/` paths when `favicon` permission is declared

### LINKED ISSUE #1559: Automatically update `.wxt/types/paths.d.ts` and `web_accessible_resources` manifest for favicon
### Feature Request

For favicon feature: https://developer.chrome.com/docs/extensions/how-to/ui/favicons

Currently, if we want to use this, we have to do following steps:

1. Update permissions in wxt.config.ts to include `favicon`
2. Update `web_accessible_resources` in manifest to allow `_favicon/*`
3. We have to add `@ts-ignore` or `@ts-expect-error` when requesting favicon url (since it's not allowed by `.wxt/types/paths.d.ts`)

```
function getFaviconUrl(url: string, size = 16) {
  //@ts-expect-error not supported by wxt now
  const u = new URL(browser.runtime.getURL("/_favicon/"));
  u.searchParams.set("pageUrl", url);
  u.searchParams.set("size", size.toString());
  return u.toString();
}
```

I expect to use this feature as easily as just adding the `favicon` permission.

N/A

N/A

### Additional context

N/A

oferitz:
## Summary

- Adds a built-in WXT module (`wxt:built-in:favicon-permission`) that, when the `favicon` permission is declared in the manifest, splices a template-literal member `` `/_favicon/${string}` `` into the `PublicPath` union. Calls like `browser.runtime.getURL('/_favicon/?pageUrl=...')` now type-check without `@ts-expect-error`.
- The type is spliced as a **real** template literal (backticks), not a string literal, so arbitrary query strings are accepted while unrelated paths are still rejected.
- Intentionally does **not** touch `web_accessible_resources` — per the review feedback on #1570, only extensions that actually use favicons from a content script need that, and they can add their own entry.
- Adds a short docs page (`docs/guide/essentials/favicons.md`) covering both the type behavior and the content-script WAR opt-in.

Closes #1559

## Relationship to #1570

Revives the abandoned #1570 (inactive since Sept 2025) with two corrections to the review feedback that were never addressed:

1. The original pushed `'_favicon/?${string}'` into `prepare:publicPaths`, but the paths generator wraps every entry in `"..."` — producing a string-literal type `"/_favicon/?${string}"` that only matches those literal characters. Query strings would not have actually type-checked. This PR hooks `prepare:types` instead and splices backticks directly so the union member is a true template literal.
2. The original still added a `web_accessible_resources` entry despite the maintainer asking for it to be removed. This PR does not, and a negative assertion in the test locks that in.

I left a comment on #1570 before starting, offering to defer if the original author wanted to return to it.

## Test plan

- [x] New e2e test `should type \`/_favicon/\` paths when the \`favicon\` permission is declared` in `packages/wxt/e2e/tests/typescript-project.test.ts` — asserts (a) inline-snapshot of `.wxt/types/paths.d.ts` showing the template-literal member and (b) `manifest.web_accessible_resources` is `undefined` after build.
- [x] `pnpm vitest run` — 482 passed, 2 skipped (yarn tests skipped locally — unrelated, they require the `yarn` CLI).
- [x] `pnpm run check` — ESLint, Oxlint, Publint, TypeScript all green.
- [x] Existing `hooks.test.ts` and `typescript-project.test.ts` suites unaffected.


PatrykKuniczak (approved):
Works well 😄 

But i'm wondering if it's good idea to add `getFaviconUrl` and `getFaviconImg` (my suggestion) to exports of `wxt`, to helps WXT implementators to avoid need to create it by theyself.

I think it's good idea to add:
```ts
    function getFaviconUrl(pageUrl: string, size = 16) {
      const url = new URL(browser.runtime.getURL('/_favicon/'));
      url.searchParams.set('pageUrl', pageUrl);
      url.searchParams.set('size', String(size));
      return url.toString();
    }

    function getFaviconImg(
      pageUrl: string,
      alt: string,
      size = 16,
      classList = '',
    ) {
      const url = getFaviconUrl(pageUrl, size);

      const img = document.createElement('img');
      img.src = url.toString();
      img.alt = alt;
      img.classList.add(classList);

      return img;
    }
```

but let's wait for @aklinker1 approval.

PatrykKuniczak:
@oferitz Let's fix issue with `check`

PatrykKuniczak (changes requested):
OO yeah, let's avoid adding on Firefox, because it doesn't work here.
Cover that case

packages/wxt/e2e/tests/typescript-project.test.ts
```diff
@@ -462,6 +462,22 @@ describe('TypeScript Project', () => {
...
+    // path that 404s at runtime.
+    await project.build({
+      browser: 'firefox',
+      manifest: { permissions: ['favicon'] },
+    });
+
+    const pathsOutput = await project.serializeFile('.wxt/types/paths.d.ts');
+    expect(pathsOutput).not.toContain('/_favicon/');
+  });
+
```
aklinker1: This is going to lead to really inconsistent types - `wxt prepare` will generate types as chrome, then when you build for Firefox, it will remove the types and cause type errors in your IDE, a bad DX. I would prefer we always generate the type.

There has been lots of discussion around how to handle typing chrome or firefox-only APIs, and my personal take is that it's up to the developer to know when an API is available. TypeScript just isn't built for projects that need multiple type systems gated behind feature flags.

https://github.com/wxt-dev/wxt/issues/1652

docs/.vitepress/config.ts
```diff
@@ -219,6 +219,7 @@ export default defineConfig({
           ),
           menuItem('Extension APIs', 'extension-apis.md'),
           menuItem('Assets', 'assets.md'),
+          menuItem('Favicons', 'favicons.md'),
```
aklinker1: IMO, favicons are less important than targeting browsers, content scripts, and some other extension APIs. Let's move this down further, after I18n.

docs/guide/essentials/favicons.md
```diff
@@ -0,0 +1,53 @@
+# Favicons
+
+[Chrome Docs: How to fetch favicons](https://developer.chrome.com/docs/extensions/how-to/ui/favicons)
```
aklinker1:
```suggestion
[Chrome Docs](https://developer.chrome.com/docs/extensions/how-to/ui/favicons)
```

Keep it consistent with the other extension API docs.

docs/guide/essentials/favicons.md
```diff
@@ -0,0 +1,53 @@
...
+// Any entrypoint
+function getFaviconUrl(pageUrl: string, size = 16) {
+  const url = new URL(browser.runtime.getURL('/_favicon/'));
+  url.searchParams.set('pageUrl', pageUrl);
+  url.searchParams.set('size', String(size));
+  return url.toString();
+}
+```
+
+When `favicon` is in `permissions`, WXT automatically augments `browser.runtime.getURL` so that any `` `/_favicon/${string}` `` path type-checks — you no longer need a `@ts-expect-error`.
```
aklinker1:
```suggestion
```

This doesn't need to be documented if it just works for developers in their IDE.

packages/wxt/src/builtin-modules/favicon-permission.ts
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * @see https://developer.chrome.com/docs/extensions/how-to/ui/favicons
+ */
+export default defineWxtModule({
+  name: 'wxt:built-in:favicon-permission',
+  setup(wxt) {
+    wxt.hooks.hook('prepare:types', (_, entries) => {
+      // Firefox has no equivalent favicon API, so the `/_favicon/${string}`
+      // path would not resolve at runtime — don't suggest it via types.
+      if (wxt.config.browser === 'firefox') return;
```
aklinker1:
```suggestion
```

packages/wxt/src/builtin-modules/favicon-permission.ts
```diff
@@ -0,0 +1,48 @@
...
+      // The base generator wraps every public path in double quotes,
+      // producing string-literal union members. Favicon URLs need a
+      // template-literal type so arbitrary query strings type-check, so
+      // we splice a `\`/_favicon/\${string}\`` member into the union
+      // directly. The `HtmlPublicPath` line is a stable anchor emitted
+      // by generate-wxt-dir.ts:getPathsDeclarationEntry.
+      const anchor = '  type HtmlPublicPath';
+      pathsEntry.text = pathsEntry.text.replace(
+        anchor,
+        '    | `/_favicon/${string}`\n' + anchor,
```
aklinker1, line 37: 🤖 _This comment was generated by AI._

**Fragile string replacement against generated output.** This splices into the generated `paths.d.ts` text by finding the exact string `'  type HtmlPublicPath'` as an anchor. That anchor line is an internal implementation detail of `getPathsDeclarationEntry()` in `generate-wxt-dir.ts` — any future refactor that changes its indentation, renames the type, or restructures the template will silently break this replacement (`.replace()` returns the original string unchanged on no match, so the `/_favicon/` member would just disappear with no error).

I understand *why* `prepare:types` + string replacement was chosen: the existing `prepare:publicPaths` hook wraps every entry in double quotes (`"/${path}"`), producing a string-literal type that won't match arbitrary query strings. A template-literal type (`` `/_favicon/${string}` ``) is genuinely needed here.

However, I'd suggest considering a more robust approach — for example, extending `getPathsDeclarationEntry()` to support raw/template-literal union members natively. Something like allowing `prepare:publicPaths` to accept entries marked as "raw" that get emitted verbatim instead of wrapped in `"..."`. That would make this feature resilient to template reformatting while also benefiting future consumers that need the same capability.

At minimum, if the string-replacement approach stays, please add a defensive check:

```ts
const replaced = pathsEntry.text.replace(anchor, ...);
if (replaced === pathsEntry.text) {
  wxt.logger.warn('[wxt:built-in:favicon-permission] Could not splice /_favicon/ type — anchor not found in paths.d.ts. The internal template may have changed.');
}
pathsEntry.text = replaced;
```

This way a broken splice surfaces as a visible warning instead of silently degrading.

aklinker1: >  The original pushed `'_favicon/?${string}'` into `prepare:publicPaths`, but the paths generator wraps every entry in `"..."` — producing a string-literal type `"/_favicon/?${string}"` that only matches those literal characters. Query strings would not have actually type-checked. This PR hooks `prepare:types` instead and splices backticks directly so the union member is a true template literal.

Good find!

I don't like this either, but:

- Since it's internal to WXT, we can easily replace this with a better implementation once a feature is added to `prepare:publicPaths` hook,
- And you have tests, so if it breaks we'll know about it

So don't make any changes here, I'm just leaving a comment so I can open an issue based off it to track this improvement.

aklinker1 (changes requested):
Looks good overall, thanks for the PR. Some requested changes than we can get this merged.

> Let's avoid adding on Firefox, because it doesn't work here.

TypeScript types need to be stable and not be based on the target browser. See https://github.com/wxt-dev/wxt/pull/2241#discussion_r3068147712O for more details.

packages/wxt/src/builtin-modules/favicon-permission.ts
aklinker1: https://github.com/wxt-dev/wxt/issues/2245

aklinker1 (approved):
Nice, thanks for making the changes!

aklinker1:
> i'm wondering if it's good idea to add getFaviconUrl and getFaviconImg (my suggestion) to exports of wxt, to helps WXT implementators to avoid need to create it by theyself.

@PatrykKuniczak I'm not sold that every WXT project needs the favicon APIs. I would rather create a `@wxt-dev/favicon` package that supports both Chrome and other browsers bty falling back on public favicon services (like google, 1password, bitwarden, etc).

Would either you or @oferitz be interested in making that package?

PatrykKuniczak:
> > i'm wondering if it's good idea to add getFaviconUrl and getFaviconImg (my suggestion) to exports of wxt, to helps WXT implementators to avoid need to create it by theyself.
> 
> @PatrykKuniczak I'm not sold that every WXT project needs the favicon APIs. I would rather create a `@wxt-dev/favicon` package that supports both Chrome and other browsers bty falling back on public favicon services (like google, 1password, bitwarden, etc).
> 
> Would either you or @oferitz be interested in making that package?

I was wondering about that and i think it's good idea 😄

oferitz:
> > i'm wondering if it's good idea to add getFaviconUrl and getFaviconImg (my suggestion) to exports of wxt, to helps WXT implementators to avoid need to create it by theyself.
> 
> @PatrykKuniczak I'm not sold that every WXT project needs the favicon APIs. I would rather create a `@wxt-dev/favicon` package that supports both Chrome and other browsers bty falling back on public favicon services (like google, 1password, bitwarden, etc).
> 
> Would either you or @oferitz be interested in making that package?

I Can try. Are there any docs about creating a package or just follow the other packages patterns?

PatrykKuniczak (approved):
LGTM

PatrykKuniczak:
> > > i'm wondering if it's good idea to add getFaviconUrl and getFaviconImg (my suggestion) to exports of wxt, to helps WXT implementators to avoid need to create it by theyself.
> > 
> > 
> > @PatrykKuniczak I'm not sold that every WXT project needs the favicon APIs. I would rather create a `@wxt-dev/favicon` package that supports both Chrome and other browsers bty falling back on public favicon services (like google, 1password, bitwarden, etc).
> > Would either you or @oferitz be interested in making that package?
> 
> I Can try. Are there any docs about creating a package or just follow the other packages patterns?

You need to follow the other packages patterns, because there's no [Contributors guide](https://wxt.dev/guide/resources/how-wxt-works.html#how-wxt-works) for now.

Let's create it simple, maybe not that much as https://github.com/wxt-dev/wxt/tree/main/packages/is-background 😆 , but you can see how it's works on that package.

I've been wondering about, to add also `styles` to that img element, which i want to export from that new package, to covert most cases via argument of that func, last extra props each dev can add by accessing that new `img` after assing it to `const`.

aklinker1:
You can just copy the `packages/is-background` folder for a good starting point.

PR #2242, opened 2026-04-11
feat: auto-discover Firefox `theme_icons` from public assets

### LINKED ISSUE #1120: theme_icons support in WXT or @wxt-dev/auto-icons
### Feature Request

Firefox supports theme_icons field in browser_action (MV2) or action (MV3). It would be great to automatically discover (or generate with auto-icons) those.

It would be great to auto discover them i.e. in `icon/16-light.png` etc.

For auto-icons, IMHO best approach would be automatically check for `icon/light.(png|svg)` and `icon/dark.(png|svg)` and if present, generate proper sizes. 

Simply define them manually via `manifest.browser_action.theme_icons`.

### Additional context

[documentation](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/action)

oferitz:
## Summary

When building for Firefox, WXT now auto-discovers paired `-light`/`-dark` icons in the public assets and attaches them to the extension action's [`theme_icons`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/action#theme_icons) field, parallel to how regular icons are already discovered and put on the top-level `icons` field. Users can now drop both variants in `public/` and Firefox will swap the toolbar icon based on the browser theme — no manual manifest entry required.

```plaintext
public/
├─ icon-16.png          # regular (default_icon)
├─ icon-light-16.png    # Firefox light theme
├─ icon-dark-16.png     # Firefox dark theme
├─ icon-32.png
├─ icon-light-32.png
└─ icon-dark-32.png
```

### Behavior

- **Firefox-only.** The discovery call site in `generateManifest` is gated on `wxt.config.browser === 'firefox'`.
- **Attaches wherever the action already lives.** `manifest.action` (MV3), `manifest.browser_action` or `manifest.page_action` (MV2). The existing MV2 conversion (`convertActionToMv2`) handles `action` → `browser_action` copying downstream.
- **Skips when no action exists** — e.g. background-only extensions. Nothing to attach to.
- **Skips unpaired sizes.** A size is only emitted if both `light` and `dark` variants are present.
- **Respects explicit user config.** If `manifest.action.theme_icons` is already set (via `wxt.config.ts` or the popup entrypoint's `themeIcons` option), WXT does not overwrite it.
- **Reuses the existing `ThemeIcon` type** from `packages/wxt/src/types.ts` — no duplicate definition.

### Supported filename patterns

```ts
icon-light-16.png         icon-dark-16.png
icon-light-16x16.png      icon-dark-16x16.png
icon-16-light.png         icon-16-dark.png
icon-16x16-light.png      icon-16x16-dark.png
icon/light-16.png         icon/dark-16.png   (also icons/)
icon/light-16x16.png      icon/dark-16x16.png
icon/16-light.png         icon/16-dark.png
icon/16x16-light.png      icon/16x16-dark.png
```

## Files changed

- **`packages/wxt/src/core/utils/theme-icons.ts`** (new, 81 lines) — topic-specific module parallel to `content-scripts.ts` and `content-security-policy.ts`. Exports `addDiscoveredThemeIcons` and `discoverThemeIcons`.
- **`packages/wxt/src/core/utils/manifest.ts`** (+9 lines) — imports `addDiscoveredThemeIcons` and calls it after `addEntrypoints`, inside the `browser === 'firefox'` gate, before the `build:manifestGenerated` hook so user hooks can still override.
- **`packages/wxt/src/core/utils/__tests__/manifest.test.ts`** (+236 lines, +7 integration tests).
- **`docs/guide/essentials/config/manifest.md`** (+20 lines) — new `Firefox theme_icons` subsection with directory-layout example and a snippet include of the regex table.

Closes #1120

## Test plan

New integration tests in `packages/wxt/src/core/utils/__tests__/manifest.test.ts` under `describe('theme_icons auto-discovery (Firefox)')`:

- [x] MV3: discovers paired icons, attaches to `action.theme_icons`
- [x] MV2: attaches to `browser_action.theme_icons`
- [x] Supports all 8 naming patterns (suffix/prefix × folder/no-folder × `size`/`WxH`)
- [x] Skips sizes missing a light or dark pair
- [x] Does not auto-discover for non-Firefox browsers
- [x] Respects user-provided `theme_icons` (via popup `options.themeIcons`) and does not overwrite
- [x] Skips when no action exists (background-only extension)

- [x] `pnpm -F wxt run check` — ESLint, Oxlint, Publint, TypeScript all green
- [x] `pnpm -F wxt exec vitest run src/core/utils/__tests__/manifest` — 82 passed (75 existing + 7 new)

## Follow-ups not in this PR

The issue also mentions extending [`@wxt-dev/auto-icons`](https://github.com/wxt-dev/wxt/tree/main/packages/auto-icons) to auto-**generate** theme variants from a single `light.png` / `dark.png` source (via `sharp`, like the existing `sizes` generation). Happy to open a follow-up for that — kept out of this PR to keep the scope tight and reviewable.

### DIFF
--- a/docs/guide/essentials/config/manifest.md
+++ b/docs/guide/essentials/config/manifest.md
@@ -161,7 +161,9 @@ Alternatively, you can use [`@wxt-dev/auto-icons`](https://www.npmjs.com/package
 
 ### Firefox `theme_icons`
 
-Firefox supports a [`theme_icons`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/action#theme_icons) field on the toolbar action that swaps between a light and dark variant based on the current browser theme. When building for Firefox, WXT auto-discovers paired light/dark icons in the `public/` directory and attaches them to `action` (MV3) or `browser_action` (MV2). You only need to drop both variants next to your regular icons:
+Firefox supports a [`theme_icons`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/action#theme_icons) field on the toolbar action that swaps between a light and dark variant based on the current browser theme.
+
+When [targeting Firefox](/guide/essentials/target-different-browsers.md), WXT auto-discovers paired light/dark icons in the `public/` directory and attaches them to `action` (MV3) or `browser_action` (MV2). You only need to drop both variants next to your regular icons:
 
 ```plaintext
 public/
@@ -173,11 +175,15 @@ public/
 └─ icon-dark-32.png
 ```
 
-A size is only included in `theme_icons` if **both** a light and a dark file are present — unpaired files are silently skipped. The following filename patterns are discovered:
+A size is only included in `theme_icons` if **both** a light and a dark file are present. The following filename patterns are discovered:
 
 <<< @/../packages/wxt/src/core/utils/theme-icons.ts#snippet
 
-Auto-discovery is Firefox-only. If you set `manifest.action.theme_icons` (or `manifest.browser_action.theme_icons`) explicitly in `wxt.config.ts`, WXT will not overwrite it.
+If a size has a light file but no matching dark file (or vice versa), WXT logs a warning so you can fix the pair instead of silently dropping it. The same applies if two different naming patterns resolve to the same `(size, variant)` — WXT keeps the first match and warns about the collision.
+
+Only `.png` files are discovered today. Firefox's `theme_icons` also accepts SVG, but WXT doesn't yet support SVG icons for Firefox — follow [#1120](https://github.com/wxt-dev/wxt/issues/1120) for updates.
+
+If you set `manifest.action.theme_icons` (or `manifest.browser_action.theme_icons`) explicitly in `wxt.config.ts`, WXT will not overwrite it.
 
 ## Permissions
 
--- a/docs/guide/essentials/config/manifest.md
+++ b/docs/guide/essentials/config/manifest.md
@@ -179,9 +179,7 @@ A size is only included in `theme_icons` if **both** a light and a dark file are
 
 <<< @/../packages/wxt/src/core/utils/theme-icons.ts#snippet
 
-If a size has a light file but no matching dark file (or vice versa), WXT logs a warning so you can fix the pair instead of silently dropping it. The same applies if two different naming patterns resolve to the same `(size, variant)` — WXT keeps the first match and warns about the collision.
-
-Only `.png` files are discovered today. Firefox's `theme_icons` also accepts SVG, but WXT doesn't yet support SVG icons for Firefox — follow [#1120](https://github.com/wxt-dev/wxt/issues/1120) for updates.
+Only `.png` files are discovered today, even though Firefox supports `.svg` - follow [#1120](https://github.com/wxt-dev/wxt/issues/1120) for updates.
 
 If you set `manifest.action.theme_icons` (or `manifest.browser_action.theme_icons`) explicitly in `wxt.config.ts`, WXT will not overwrite it.
 
--- a/docs/guide/essentials/config/manifest.md
+++ b/docs/guide/essentials/config/manifest.md
@@ -159,6 +159,26 @@ export default defineConfig({
 
 Alternatively, you can use [`@wxt-dev/auto-icons`](https://www.npmjs.com/package/@wxt-dev/auto-icons) to let WXT generate your icon at the required sizes.
 
+### Firefox `theme_icons`
+
+Firefox supports a [`theme_icons`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/action#theme_icons) field on the toolbar action that swaps between a light and dark variant based on the current browser theme. When building for Firefox, WXT auto-discovers paired light/dark icons in the `public/` directory and attaches them to `action` (MV3) or `browser_action` (MV2). You only need to drop both variants next to your regular icons:
+
+```plaintext
+public/
+├─ icon-16.png         # regular (default_icon)
+├─ icon-light-16.png   # Firefox light theme
+├─ icon-dark-16.png    # Firefox dark theme
+├─ icon-32.png
+├─ icon-light-32.png
+└─ icon-dark-32.png
+```
+
+A size is only included in `theme_icons` if **both** a light and a dark file are present — unpaired files are silently skipped. The following filename patterns are discovered:
+
+<<< @/../packages/wxt/src/core/utils/theme-icons.ts#snippet
+
+Auto-discovery is Firefox-only. If you set `manifest.action.theme_icons` (or `manifest.browser_action.theme_icons`) explicitly in `wxt.config.ts`, WXT will not overwrite it.
+
 ## Permissions
 
 > [Chrome docs](https://developer.chrome.com/docs/extensions/reference/permissions/)
--- a/packages/wxt/src/core/utils/manifest.ts
+++ b/packages/wxt/src/core/utils/manifest.ts
@@ -20,6 +20,7 @@ import { normalizePath } from './paths';
 import { writeFileIfDifferent } from './fs';
 import defu from 'defu';
 import { wxt } from '../wxt';
+import { addDiscoveredThemeIcons } from './theme-icons';
 import { ManifestV3WebAccessibleResource } from './types';
 import type { Browser } from '@wxt-dev/browser';
 
@@ -115,6 +116,14 @@ export async function generateManifest(
       : versionName;
 
   addEntrypoints(manifest, entrypoints, buildOutput);
+
+  // Auto-discover Firefox theme_icons from paired `-light`/`-dark`
+  // files in the public assets. Runs after `addEntrypoints` so we can
+  // attach to whichever action the popup entrypoint set up, and before
+  // `build:manifestGenerated` so user hooks can still override.
+  if (wxt.config.browser === 'firefox') {
+    addDiscoveredThemeIcons(manifest, buildOutput);
+  }
 
   if (wxt.config.command === 'serve') addDevModeCsp(manifest);
   if (wxt.config.command === 'serve') addDevModePermissions(manifest);
--- a/packages/wxt/src/core/utils/manifest.ts
+++ b/packages/wxt/src/core/utils/manifest.ts
@@ -117,10 +117,6 @@ export async function generateManifest(
 
   addEntrypoints(manifest, entrypoints, buildOutput);
 
-  // Auto-discover Firefox theme_icons from paired `-light`/`-dark`
-  // files in the public assets. Runs after `addEntrypoints` so we can
-  // attach to whichever action the popup entrypoint set up, and before
-  // `build:manifestGenerated` so user hooks can still override.
   if (wxt.config.browser === 'firefox') {
     addDiscoveredThemeIcons(manifest, buildOutput);
   }
--- a/packages/wxt/src/core/utils/theme-icons.ts
+++ b/packages/wxt/src/core/utils/theme-icons.ts
@@ -1,0 +1,81 @@
+import type { Browser } from '@wxt-dev/browser';
+import type { BuildOutput, ThemeIcon } from '../../types';
+import { normalizePath } from './paths';
+
+/**
+ * If the manifest has an action (or browser/page_action) and the user has not
+ * already set `theme_icons` on it, discover light/dark icon pairs from the
+ * public assets and attach them.
+ *
+ * Firefox-only; the caller gates on `wxt.config.browser`.
+ */
+export function addDiscoveredThemeIcons(
+  manifest: Browser.runtime.Manifest,
+  buildOutput: Omit<BuildOutput, 'manifest'>,
+): void {
+  // The popup entrypoint sets `manifest.action` in MV3 and either
+  // `browser_action` or `page_action` in MV2 (see addEntrypoints in
+  // manifest.ts). Users can also declare any of these via wxt.config.ts.
+  // We attach to the one that already exists — the MV2 conversion step
+  // (convertActionToMv2) later copies `action` → `browser_action` if
+  // needed.
+  const action =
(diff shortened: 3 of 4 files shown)


packages/wxt/src/core/utils/manifest.ts
```diff
@@ -116,6 +117,14 @@ export async function generateManifest(
 
   addEntrypoints(manifest, entrypoints, buildOutput);
 
+  // Auto-discover Firefox theme_icons from paired `-light`/`-dark`
+  // files in the public assets. Runs after `addEntrypoints` so we can
+  // attach to whichever action the popup entrypoint set up, and before
+  // `build:manifestGenerated` so user hooks can still override.
```
aklinker1: Comment doesn't provide much value IMO. Most is already in the function's JSDoc.

```suggestion
```

packages/wxt/src/core/utils/theme-icons.ts
```diff
@@ -0,0 +1,81 @@
...
+    for (const regex of themeIconRegex) {
+      const match = asset.fileName.match(regex);
+      if (match?.groups == null) continue;
+      const size = Number(match.groups.size);
+      const variant = match.groups.variant as 'light' | 'dark';
+      const entry = bySize.get(size) ?? {};
+      entry[variant] = normalizePath(asset.fileName);
+      bySize.set(size, entry);
+      break;
+    }
```
aklinker1, line 60: 🤖 *This comment was generated by AI.*

**Edge case: mixed naming patterns for the same size.**

If a user happens to have both `icon-light-16.png` and `icon/light-16.png` in their public assets, both would match different regexes but resolve to the same size (16) and variant (`light`). The second match would silently overwrite the first in the `bySize` map. In a worst case, you could end up with a `light` path from one naming convention and a `dark` path from another for the same size entry.

This is unlikely in practice, but you could either:
- Log a warning when a `(size, variant)` pair is about to be overwritten, or
- Only accept the first match per `(size, variant)` by checking `if (entry[variant] == null)` before assigning.

packages/wxt/src/core/utils/theme-icons.ts
```diff
@@ -0,0 +1,81 @@
...
+  /^icon-(?<variant>light|dark)-(?<size>[0-9]+)\.png$/,              // icon-light-16.png
+  /^icon-(?<variant>light|dark)-(?<size>[0-9]+)x[0-9]+\.png$/,       // icon-light-16x16.png
+  /^icon-(?<size>[0-9]+)-(?<variant>light|dark)\.png$/,              // icon-16-light.png
+  /^icon-(?<size>[0-9]+)x[0-9]+-(?<variant>light|dark)\.png$/,       // icon-16x16-light.png
+  /^icons?[/\\](?<variant>light|dark)-(?<size>[0-9]+)\.png$/,        // icon/light-16.png | icons/light-16.png
+  /^icons?[/\\](?<variant>light|dark)-(?<size>[0-9]+)x[0-9]+\.png$/, // icon/light-16x16.png
+  /^icons?[/\\](?<size>[0-9]+)-(?<variant>light|dark)\.png$/,        // icon/16-light.png
+  /^icons?[/\\](?<size>[0-9]+)x[0-9]+-(?<variant>light|dark)\.png$/, // icon/16x16-light.png
+];
+// #endregion snippet
```
aklinker1, line 93: 🤖 *This comment was generated by AI.*

**Only `.png` is matched — Firefox `theme_icons` also supports `.svg`.**

The regexes all end with `\.png$`. This is consistent with the existing `iconRegex` for the top-level `icons` field, but Firefox's [`theme_icons`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/action#theme_icons) documentation explicitly mentions SVG as a supported format, and SVGs are arguably the most natural fit for theme-variant icons since they scale to any size.

Consider broadening the extension match to `\.(png|svg)$`, or at least calling it out as a known limitation / follow-up. If you do add SVG support, note that SVGs don't have an intrinsic pixel size, so you'd likely need to require the size in the filename (which is already the case here).

packages/wxt/src/core/utils/theme-icons.ts
```diff
@@ -0,0 +1,81 @@
...
+export function addDiscoveredThemeIcons(
+  manifest: Browser.runtime.Manifest,
+  buildOutput: Omit<BuildOutput, 'manifest'>,
+): void {
+  // The popup entrypoint sets `manifest.action` in MV3 and either
+  // `browser_action` or `page_action` in MV2 (see addEntrypoints in
+  // manifest.ts). Users can also declare any of these via wxt.config.ts.
+  // We attach to the one that already exists — the MV2 conversion step
+  // (convertActionToMv2) later copies `action` → `browser_action` if
+  // needed.
```
aklinker1: Comment doesn't add much value, the if-statements below and JSDoc already cover most of what this has to say.

```suggestion
```

packages/wxt/src/core/utils/theme-icons.ts
aklinker1: Let's log a warning in this case.

We should also log a warning instead of silently ignoring sizes that don't have both a light and dark icon.

packages/wxt/src/core/utils/theme-icons.ts
aklinker1: Let's just add a comment to the docs that SVGs aren't supported for now. Should probably support it in a future, will create an issue to track.

https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/icons#svg

WXT doesn't really support SVGs for firefox in general, might want to add support in-general for SVGs as well.

packages/wxt/src/core/utils/theme-icons.ts
```diff
@@ -0,0 +1,81 @@
...
+  const action =
+    manifest.action ?? manifest.browser_action ?? manifest.page_action;
```
aklinker1: Page actions don't support `theme_icons`, should it be listed as a fallback here?

https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/page_action#browser_compatibility

In the docs, you didn't mention `page_action` either.

docs/guide/essentials/config/manifest.md
```diff
@@ -159,6 +159,26 @@ export default defineConfig({
...
+├─ icon-32.png
+├─ icon-light-32.png
+└─ icon-dark-32.png
+```
+
+A size is only included in `theme_icons` if **both** a light and a dark file are present — unpaired files are silently skipped. The following filename patterns are discovered:
+
+<<< @/../packages/wxt/src/core/utils/theme-icons.ts#snippet
+
+Auto-discovery is Firefox-only. If you set `manifest.action.theme_icons` (or `manifest.browser_action.theme_icons`) explicitly in `wxt.config.ts`, WXT will not overwrite it.
```
aklinker1:
```suggestion
If you set `manifest.action.theme_icons` (or `manifest.browser_action.theme_icons`) explicitly in `wxt.config.ts`, WXT will not overwrite it.
```

docs/guide/essentials/config/manifest.md
```diff
@@ -159,6 +159,26 @@ export default defineConfig({
 
 Alternatively, you can use [`@wxt-dev/auto-icons`](https://www.npmjs.com/package/@wxt-dev/auto-icons) to let WXT generate your icon at the required sizes.
 
+### Firefox `theme_icons`
+
+Firefox supports a [`theme_icons`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/action#theme_icons) field on the toolbar action that swaps between a light and dark variant based on the current browser theme. When building for Firefox, WXT auto-discovers paired light/dark icons in the `public/` directory and attaches them to `action` (MV3) or `browser_action` (MV2). You only need to drop both variants next to your regular icons:
```
aklinker1:
```suggestion
Firefox supports a [`theme_icons`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/action#theme_icons) field on the toolbar action that swaps between a light and dark variant based on the current browser theme.

When [targeting Firefox](/guide/essentials/target-different-browsers.md), WXT auto-discovers paired light/dark icons in the `public/` directory and attaches them to `action` (MV3) or `browser_action` (MV2). You only need to drop both variants next to your regular icons:
```

docs/guide/essentials/config/manifest.md
```diff
@@ -159,6 +159,26 @@ export default defineConfig({
...
+public/
+├─ icon-16.png         # regular (default_icon)
+├─ icon-light-16.png   # Firefox light theme
+├─ icon-dark-16.png    # Firefox dark theme
+├─ icon-32.png
+├─ icon-light-32.png
+└─ icon-dark-32.png
+```
+
+A size is only included in `theme_icons` if **both** a light and a dark file are present — unpaired files are silently skipped. The following filename patterns are discovered:
```
aklinker1: We can remove this part if we log warnings instead of silently ignoring them.

```suggestion
A size is only included in `theme_icons` if **both** a light and a dark file are present. The following filename patterns are discovered:
```

aklinker1 (changes requested):
Thanks for adding support for this! Couple of comments.

> The issue also mentions extending [@wxt-dev/auto-icons](https://github.com/wxt-dev/wxt/tree/main/packages/auto-icons) to auto-generate theme variants from a single light.png / dark.png source (via sharp, like the existing sizes generation). Happy to open a follow-up for that — kept out of this PR to keep the scope tight and reviewable.

A future PR would be perfect!

packages/wxt/src/core/utils/theme-icons.ts
```diff
@@ -0,0 +1,103 @@
...
+      if (match?.groups == null) continue;
+      const size = Number(match.groups.size);
+      const variant = match.groups.variant as 'light' | 'dark';
+      const entry = bySize.get(size) ?? {};
+      const incoming = normalizePath(asset.fileName);
+      const existing = entry[variant];
+      if (existing != null && existing !== incoming) {
+        // Multiple files resolved to the same (size, variant) via different
+        // naming conventions — keep the first match and warn so the user
+        // can consolidate to a single pattern.
```
aklinker1: Tell you AI to leave less comments.

```suggestion
```

packages/wxt/src/core/utils/theme-icons.ts
```diff
@@ -0,0 +1,103 @@
...
+import { normalizePath } from './paths';
+import { wxt } from '../wxt';
+
+/**
+ * If the manifest has an `action` (MV3) or `browser_action` (MV2) and the user
+ * has not already set `theme_icons` on it, discover light/dark icon pairs from
+ * the public assets and attach them.
+ *
+ * `page_action` is intentionally excluded: Firefox does not honor `theme_icons`
+ * on page actions.
```
aklinker1:
```suggestion
```

packages/wxt/src/core/utils/theme-icons.ts
```diff
@@ -0,0 +1,103 @@
...
+/**
+ * Scan `publicAssets` for paired `-light`/`-dark` icon files and return the
+ * sizes where both variants exist, sorted ascending. Returns `undefined` if no
+ * complete pairs are found so callers can short-circuit.
+ *
+ * Warnings are logged for mixed-naming collisions and for sizes that are
+ * missing a light or dark pair, so users notice misnamed files rather than
+ * having them silently dropped.
+ */
+export function discoverThemeIcons(
```
aklinker1, line 34: Why is this exported? Doesn't seem to be used outside this module?

packages/wxt/src/core/utils/theme-icons.ts
```diff
@@ -0,0 +1,103 @@
...
+}
+
+/**
+ * Scan `publicAssets` for paired `-light`/`-dark` icon files and return the
+ * sizes where both variants exist, sorted ascending. Returns `undefined` if no
+ * complete pairs are found so callers can short-circuit.
+ *
+ * Warnings are logged for mixed-naming collisions and for sizes that are
+ * missing a light or dark pair, so users notice misnamed files rather than
+ * having them silently dropped.
```
aklinker1:
```suggestion
```

packages/wxt/src/core/utils/theme-icons.ts
```diff
@@ -0,0 +1,103 @@
...
+    );
+  }
+  pairs.sort((a, b) => a.size - b.size);
+
+  return pairs.length > 0 ? pairs : undefined;
+}
+
+// SVG is not supported yet — Firefox's `theme_icons` accepts SVG, but WXT
+// does not currently discover SVG icons for Firefox in general. Tracked as
+// a follow-up; widen the extension match once SVG icons are supported.
```
aklinker1:
```suggestion
```

aklinker1 (changes requested):

aklinker1:
@PatrykKuniczak another reason to not force push... you reverted some of the changes that were made based on my comments... now I have to re-review to decide what was reverted...

~If you have the old commit in your reflog, I would appreciate you restoring this branch to what it was before.~

Actually don't. You probably didn't have them locally, I'll just add the changes back. Use `--force-with-lease` instead of `--force` to not force push over commits on the remote you don't have locally.

PatrykKuniczak:
> @PatrykKuniczak another reason to not force push... you reverted some of the changes that were made based on my comments... now I have to re-review to decide what was reverted...
> 
> ~If you have the old commit in your reflog, I would appreciate you restoring this branch to what it was before.~
> 
> Actually don't. You probably didn't have them locally, I'll just add the changes back. Use `--force-with-lease` instead of `--force` to not force push over commits on the remote you don't have locally.

Ooo, i've though that uses `--force-with-lease` by default, i'll be updating with merge commit, it doesn't have impact, because after all we doing squash merge.

aklinker1:
Thankfully not a big deal this time, thanks for switching!

aklinker1 (approved):

PR #2248, opened 2026-04-12
fix: Support template literals in prepare:publicPaths

### LINKED ISSUE #2245: Support adding template literals in `prepare:publicPath` hook
> >  The original pushed `'_favicon/?${string}'` into `prepare:publicPaths`, but the paths generator wraps every entry in `"..."` — producing a string-literal type `"/_favicon/?${string}"` that only matches those literal characters. Query strings would not have actually type-checked. This PR hooks `prepare:types` instead and splices backticks directly so the union member is a true template literal.
> 
> Good find!
> 
> I don't like this either, but:
> 
> - Since it's internal to WXT, we can easily replace this with a better implementation once a feature is added to `prepare:publicPaths` hook,
> - And you have tests, so if it breaks we'll know about it
> 
> So don't make any changes here, I'm just leaving a comment so I can open an issue based off it to track this improvement.

_Originally posted by @aklinker1 in https://github.com/wxt-dev/wxt/pull/2241#discussion_r3068164716_

eupthere:
### Overview
- add backward-compatible `prepare:publicPaths` support for both raw `string` entries and explicit `{ type, path }` entries
- generate `PublicPath` union members as either quoted string literals or backtick template literal types based on the entry kind
- migrate the built-in favicon permission module to register its `/_favicon/?${string}` path through `prepare:publicPaths` instead of patching generated type text

### Manual Testing
- `pnpm --filter wxt test -- typescript-project.test.ts`
- `pnpm --filter wxt check`

### Related Issue
This PR closes #2245


aklinker1 (approved):
Nice, thanks for the quick follow up!

PR #2249, opened 2026-04-12
refactor: Rename internal popup `mv2Key` option to `actionType`

aklinker1:
### Overview

Adds `actionType` as a properly-named replacement for `mv2Key` on `PopupEntrypointOptions`. The old `mv2Key` name was misleading because it also applies to Firefox MV3. Both properties are kept in sync via getter/setter so modules reading or writing either one see the same value. `mv2Key` is marked `@deprecated`.

### Manual Testing

`pnpm --filter wxt vitest run src/core/utils/__tests__/manifest.test.ts` — all 83 tests pass.

### Related Issue

Follow-up from https://github.com/wxt-dev/wxt/pull/2200#discussion_r3069671824


packages/wxt/src/core/utils/building/find-entrypoints.ts
```diff
@@ -274,23 +274,40 @@ async function getPopupEntrypoint(
...
   // Extract non-per-browser options
-  const { themeIcons, title, type, ...perBrowserOptions } = options;
+  const { themeIcons, title, type, actionType, ...perBrowserOptions } = options;
 
   const strictOptions: PopupEntrypoint['options'] = resolvePerBrowserOptions(
     {
       ...perBrowserOptions,
       defaultTitle: title,
-      mv2Key: type,
+      actionType: type ?? actionType,
```
aklinker1: Using `type` from options is still expected, we don't need to add support for `actionType` there as well.

PR #2250, opened 2026-04-12
ci: Don't post thank you message on closed PRs for maintainers and bots

aklinker1:
### Overview

These are causing lots of extra notifications for me that I don't care about seeing. @PatrykKuniczak you want added to this list?


PatrykKuniczak:
@aklinker1 Yeah, let's add me please 😄

PatrykKuniczak:
@aklinker1 Can we also avoid sending info from `codecov` and `pkg-pr-new`?
There's a looot of emails on my box, i'm sure you have x10 more.

aklinker1:
I don't think that's a thing... I could disable the comments, but both are useful. If you use emails, you could setup a filter to delete any emails from those two apps.

I use the web notification center, not emails, and I don't think there's anything I can do about them :/

PR #2255, opened 2026-04-13
docs: Move homepage extensions to separate YAML file

aklinker1:
### Overview

Vue file was getting large, annoying to modify and annoying for contributors to scroll down to find the bottom of the list. So I moved it to a YAML file.

Wrote my own plugin to minimize dependencies, all the vite plugins that add yaml support are 3+ dependencies... With bun it's zero.

### Manual Testing

```sh
bun docs:dev
```

And make sure the extensions still load.

Or checkout the deploy preview.

### Related Issue

Follow up to #2208


package.json
```diff
@@ -13,9 +13,9 @@
     "test:coverage": "bun run --workspaces test:coverage",
     "prepare": "simple-git-hooks",
     "docs:gen": "typedoc --options docs/typedoc.json",
-    "docs:dev": "bun run --silent docs:gen && bun run vitepress dev docs",
-    "docs:build": "bun run --silent docs:gen && bun run vitepress build docs",
-    "docs:preview": "bun run --silent docs:gen && bun run vitepress preview docs"
+    "docs:dev": "bun run --silent docs:gen && bun run --bun vitepress dev docs",
+    "docs:build": "bun run --silent docs:gen && bun run --bun vitepress build docs",
+    "docs:preview": "bun run --silent docs:gen && bun run --bun vitepress preview docs"
```
aklinker1, line 18: Without `--bun`, vitepress runs in node, not bun, and the `Bun` global is not defined.

PR #2268, opened 2026-04-15
fix: deduplicate content script CSS in build output

### LINKED ISSUE #1987: Content Script CSS is added twice to build output
### Describe the bug

When adding a content css file to the entrypoints dir and importing it in content.ts, it get's included twice in the build output and builded zip. Once under .output/chrome-mv3/content-scripts/content.css and .output/chrome-mv3/assets/example.css. Both files have the same content.

### Reproduction

[Archive.zip](https://github.com/user-attachments/files/23981105/Archive.zip)

### Steps to reproduce

Build the project included in the attached zip file. node_modules is not included requiring `npm i` to install the dependencies. Look at the .output directory.

Alternatively, create a new project with the vanilla template. Add a file named example.css in the entrypoints directory and import it in the content.ts file as follows: `import "./example.css"`.
Execute `npm run build` or `npm run zip` and inspect the output folder.
Both files .output/chrome-mv3/content-scripts/content.css and .output/chrome-mv3/assets/example.css have the same contents as entrypoints/example.css

### System Info

```shell
System:
    OS: Linux 6.17 Arch Linux
    CPU: (12) x64 AMD Ryzen 5 3600 6-Core Processor
    Memory: 9.88 GB / 15.56 GB
    Container: Yes
    Shell: 4.2.1 - /usr/bin/fish
  Binaries:
    Node: 25.2.1 - /usr/bin/node
    Yarn: 1.22.22 - /usr/bin/yarn
    npm: 11.6.4 - /usr/bin/npm
    Deno: 2.5.6 - /usr/bin/deno
  Browsers:
    Chromium: 142.0.7444.175
    Firefox: 145.0.2
    Firefox Developer Edition: 145.0.2
  npmPackages:
    wxt: ^0.20.6 => 0.20.11
```

### Used Package Manager

npm

### Validations

oc-wonton:
## Description

When content scripts import CSS files, Vite outputs the CSS in **two** locations:
1. `content-scripts/{name}.css` (referenced in manifest)
2. `assets/{name}.css` (duplicate, not needed)

This adds a post-build deduplication step that removes identical CSS files from `assets/` when they already exist in `content-scripts/` with the same base name and content.

## Related Issues

Fixes #1987

## Type of Change

- [x] Bug fix

## Implementation

- New `deduplicateCss()` utility in `packages/wxt/src/core/utils/building/deduplicate-css.ts`
- Called after entrypoint builds merge but before manifest generation (in `rebuild.ts`)
- Conservative matching: only removes files with **same base name AND identical content**
- Preserves legitimate CSS files in `assets/` (e.g., unlisted-style entrypoints)

## Testing

- 8 unit tests covering: duplicate removal, different content preservation, different name preservation, multiple files, empty cases, missing files, multi-step builds
- 1 e2e test verifying CSS only appears in `content-scripts/` and not duplicated in `assets/`

```
✓ src/core/utils/building/__tests__/deduplicate-css.test.ts (8 tests) 71ms
  Test Files  1 passed (1)
       Tests  8 passed (8)
```

## Additional Notes

Based on maintainer guidance in #1987: each content script is built in a separate Vite build, causing CSS to appear in both locations. This fix deduplicates at the output level rather than changing the build pipeline.


oyzamil:
@oc-wonton When using tailwindcss in content-script and other pages like popup, options etc.

The CSS file is created 2 times. Will this pull also fix it or not?

See the 2 files. I am also curious why the CSS file is named as my component name, Button-id.css; it should be styles.css

image

oyzamil:
@aklinker1 Please review this.

packages/wxt/src/core/utils/building/__tests__/deduplicate-css.test.ts
```diff
@@ -0,0 +1,361 @@
...
+        outDir: testOutDir,
+      },
+    });
+  });
+
+  afterEach(async () => {
+    // Cleanup
+    await rm(testOutDir, { recursive: true, force: true });
+  });
+
```
PatrykKuniczak, line 30: That isn't necessary, because you're doing it in `beforeEach`

packages/wxt/src/core/utils/building/__tests__/deduplicate-css.test.ts
```diff
@@ -0,0 +1,361 @@
...
+    const csPath = resolve(testOutDir, 'content-scripts/content.css');
+    const assetPath = resolve(testOutDir, 'assets/content.css');
+
+    await writeFile(csPath, cssContent, 'utf-8');
+    await writeFile(assetPath, cssContent, 'utf-8');
+
+    const output: Omit<BuildOutput, 'manifest'> = {
+      steps: [
+        {
+          entrypoints: [] as any,
```
PatrykKuniczak, line 43:
```suggestion
          entrypoints: [],
```

This isn't working?

PatrykKuniczak (changes requested):

packages/wxt/src/core/utils/building/deduplicate-css.ts
PatrykKuniczak, line 1: It isn't possible to avoid creating duplicated css, instead of removing it after create?

packages/wxt/e2e/tests/output-structure.test.ts
```diff
@@ -291,6 +291,41 @@ describe('Output Directory Structure', () => {
       `);
   });
 
+  it('should not duplicate content script CSS in assets/ directory', async () => {
```
aklinker1, line 294: Ran this test on `main`, it already passes

packages/wxt/src/core/utils/building/__tests__/deduplicate-css.test.ts
```diff
@@ -0,0 +1,361 @@
...
+      'content-scripts/content.css',
+    );
+  });
+
+  it('should not remove non-duplicate CSS files from assets/', async () => {
+    // Setup: Create different CSS files with different names
+    const csContent = '.content { color: red; }';
+    const assetContent = '.other { color: blue; }';
+    const csPath = resolve(testOutDir, 'content-scripts/content.css');
+    const assetPath = resolve(testOutDir, 'assets/other.css');
```
aklinker1, line 77: I'm really confused about why you keep calling out "content-scripts" vs "assets". They're the same thing, they can both be deduplicated.

packages/wxt/src/core/utils/building/__tests__/deduplicate-css.test.ts
aklinker1, line 1: Move this to an E2E test and use the existing `TestProject` util. Will reduce all the code duplication here, there's a lot of setup because you're recreating the entire FS setup/cleanup.

I would rather have these be E2E tests too, there seems to be too much setup and specific mocking that could be fragile.

packages/wxt/src/core/utils/building/deduplicate-css.ts
aklinker1: Yes, it's possible. Rollup's [`generateBundle`](https://rollupjs.org/plugin-development/#generatebundle) hook lets a plugin mutate (or delete entries from) the bundle object before Vite writes it to disk. We already use this pattern in [`cssEntrypoints.ts`](https://github.com/wxt-dev/wxt/blob/main/packages/wxt/src/core/builders/vite/plugins/cssEntrypoints.ts#L34-L38) to drop the unwanted JS output of CSS-only lib-mode builds.

So a cleaner fix would be a small Vite plugin applied in `getLibModeConfig` (in `packages/wxt/src/core/builders/vite/index.ts`) for `content-script` entrypoints, which tracks CSS outputs throughout the build and deduplicates them before writing them to the file. Would be easier to test too (no temp dirs / FS mocking).

aklinker1 (changes requested):
Couple of comments.

Also, we should add an `experimental` for this. We don't have any others right now, but we shouldn't enable this by default for all projects without further testing.

https://github.com/wxt-dev/wxt/blob/d9c140114a643371be7a0b95ad79f85372463e21/packages/wxt/src/types.ts#L366

PR #2285, opened 2026-04-18
docs: base icon path configuration for auto-icons

Sbrjt:
### Overview

Added instructions for configuring base icon path in wxt.config.ts.

```ts
import { resolve } from 'node:path'
import { defineConfig } from 'wxt'

export default defineConfig({
  modules: ['@wxt-dev/module-react', '@wxt-dev/auto-icons'],
  autoIcons: {
    baseIconPath: resolve('public/icon.svg'),
  },
}
```

### Manual Testing

None

### Related Issue

None

### DIFF
--- a/packages/auto-icons/README.md
+++ b/packages/auto-icons/README.md
@@ -29,6 +29,19 @@ export default defineConfig({
 
 And finally, save the base icon to `<srcDir>/assets/icon.png`.
 
+To use another path, configure `autoIcons.baseIconPath` in your `wxt.config.ts`:
+
+```ts
+import { resolve } from 'node:path'
+
+export default defineConfig({
+  ...,
+  autoIcons: {
+    baseIconPath: resolve('public/icon.svg'),
+  },
+}
+```
+
 ## Configuration
 
 The module can be configured via the `autoIcons` config:


PatrykKuniczak (changes requested):
Let's merge it with:
https://github.com/wxt-dev/wxt/pull/2285/changes#diff-e6554a3c82501208ee1d70b8db5ca82b288ac8dfde1be3c11887cccf81ab74b0R32-R44

To simplify, because it seems redundant.

Sbrjt:
I'm sorry. I didn't understand.

PatrykKuniczak:
@Sbrjt Sorry my bad, because that's over changes scope, and link doesn't work well, i need to tell you.

Let's merge your documentation with current from lines `45 - 55`

You need to only put 
```ts
  autoIcons: {
    baseIconPath: resolve('public/icon.svg'),
  },
  ```
  
inside line 53 and adjust description from line 32 a little, to match current docs, make it all consistent

aklinker1 (dismissed):

aklinker1:
What @PatrykKuniczak is trying to say is we already have a section dedicated to config: 

https://github.com/Sbrjt/wxt/blob/da8479181a0247131ca4fb83862a29b389a5d2d0/packages/auto-icons/README.md?plain=1#L45-L58

It doesn't list them, but it does point you to the code where you can see all the options.

But rather than adding anything, I think we should start generating an API reference for packages other than WXT and use a link to the type there instead of telling people to look at the code or provide more details in the README.

aklinker1:
Or better, we start publishing packages to JSR so it hosts the API reference. We can't publish `wxt` there, it's incompatible, but we can publish all the other packages there 🤔

aklinker1:
Or we just use this link for now:

https://npmx.dev/package-docs/@wxt-dev/auto-icons/v/latest#interface-AutoIconsOptions

PR #2288, opened 2026-04-21
feat: add content script registration `optional`

### LINKED ISSUE #2239: Support new content script registration `optional`
### Feature Request

Add a new `registration: 'optional'` option for content scripts that moves their `matches` into `optional_host_permissions` instead of `host_permissions`, and registers them dynamically via `browser.scripting.registerContentScripts()` at runtime.

Currently, `registration: 'runtime'` moves matches to `host_permissions`, which still triggers Chrome's permission escalation dialog when new hosts are added on extension update (disabling the extension until the user re-accepts). Extensions that want to support new hosts without disabling need to use `optional_host_permissions` instead.

**Proposed API:**

```ts
    matches: ['https://my-app.com/*'],
    registration: 'optional',
    // ...
});
```

**Behavior:**
- Strip the content script's `matches` from `manifest.json` `content_scripts`
- Add them to `optional_host_permissions` instead of `host_permissions`
    - Would be nice if this was deduped if user already had  urls in `optional_host_permissions`. example if I had `*.app.com` in `optional_host_permissions` theres no need to put `auth.app.com` because its covered by the existing one
- Nice to have: Register dynamically via `browser.scripting.registerContentScripts()` in the background script
- Scripts run only after the user grants the optional host permission

N/A — though this addresses a common pain point where adding new `content_scripts.matches` entries causes Chrome to disable the extension on update due to permission escalation, even when the hosts are already covered by `optional_host_permissions`.

We built a custom WXT module that:
1. Hooks into `entrypoints:resolved` to identify content scripts targeting optional hosts
2. Generates a virtual module with the optional content script data via `prepare:types`
3. Strips them from the manifest in `build:manifestGenerated`
4. Registers them dynamically in the background script using `browser.scripting.registerContentScripts()`

This works but requires significant custom infrastructure that could be a first-class WXT feature. The `registration: 'runtime'` codepath already handles most of this — the main difference is the target permission type (`optional_host_permissions` vs `host_permissions`).

### Additional context

- Chrome treats any new entry in `content_scripts.matches` as a permission escalation, regardless of whether the host is already in `optional_host_permissions`
- `optional_host_permissions` changes alone do **not** trigger the escalation dialog
- `browser.scripting.registerContentScripts()` registrations persist across service worker restarts and extension updates, so they only need to be set up once
- This is a common pattern for extensions that progressively support new domains without disrupting existing users

eupthere:
### Overview
Adds support for `registration: "optional"` for content scripts so host match patterns are treated as optional origins instead of required host permissions.

**Related Docs**:
[MDN optional_host_permissions](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/optional_host_permissions)
[Chrome Extensions Docs MV2 - Optional Permissions](https://developer.chrome.com/docs/extensions/mv2/reference/permissions#implementing_optional_permissions)

**What changed**:

- Extended content script registration types to include `"optional"`.
- Updated manifest generation to:
  - Exclude `registration: "optional"` scripts from `content_scripts` (same dynamic-registration model as runtime scripts).
  - Move their `matches` into `optional_host_permissions` instead of `host_permissions`.
- Added MV2 conversion support:
  - `optional_host_permissions` are merged into `optional_permissions` when targeting MV2.
- Updated dev reload handling so `"optional"` follows runtime-style content script reload flow.
- Added/updated tests for:
  - optional registration manifest output in MV3
  - optional host permission conversion in MV2
  - validation behavior around `matches`
- Updated docs to include the new `"optional"` registration mode and describe intended optional-host runtime usage.

**Please review this**:
```ts
function validateContentScriptEntrypoint(
  definition: ContentScriptEntrypoint,
): ValidationResult[] {
  const errors = validateBaseEntrypoint(definition);
  if (
    definition.options.registration !== 'runtime' &&
    definition.options.matches == null
  ) {
    errors.push({
      type: 'error',
      message:
        '`matches` is required for content scripts not registered at runtime',
      value: definition.options.matches,
      entrypoint: definition,
    });
  }
  return errors;
}
```

I intentionally kept `registration: 'runtime'` as the only mode that can omit matches, to preserve manual injection flows (for example `browser.scripting.executeScript({ target: { tabId } ... })`) where URL match patterns are not needed in entrypoint options.

`registration: "optional"` requires matches because this mode derives origin scope from matches to populate `optional_host_permissions` (and MV2 conversion to `optional_permissions`).  
 
Could you confirm whether this assumption about the original runtime exception is correct, and whether keeping optional stricter is the intended design?

### Manual Testing

**wxt core test pass**
`bun run --filter wxt test run`

**Generated Manifest**

```ts
// packages/wxt-demo/src/entrypoints/content.ts
  registration: 'optional',
  matches: ['https://example.com/*'],
  main() {
    console.log('optional test');
  },
});
```

```bash
cd packages/wxt-demo
npm run build:chrome-mv2
npm run build:chrome-mv3
```

<details>
<summary>MV2</summary>

```json
{
  "manifest_version": 2,
  "name": "wxt-demo",
  "version": "1.0.0",
  "icons": {
    "16": "icons/16.png",
    "32": "icons/32.png",
    "48": "icons/48.png",
    "128": "icons/128.png"
  },
  "permissions": ["storage"],
  "default_locale": "en",
  "web_accessible_resources": [
    "iframe-src.html",
    "unlisted.js",
    "content-scripts/ui.css"
  ],
  "background": { "scripts": ["background.js"] },
  "browser_action": { "default_title": "Popup", "default_popup": "popup.html" },
  "options_ui": { "open_in_tab": false, "page": "options.html" },
  "sandbox": { "pages": ["example.html", "sandbox.html"] },
  "content_scripts": [
    {
      "matches": ["https://*.duckduckgo.com/*"],
      "css": ["content-scripts/automount.css"],
      "js": ["content-scripts/automount.js", "content-scripts/ui.js"]
    },
    { "matches": ["<all_urls>"], "js": ["content-scripts/example-tsx.js"] },
    { "matches": ["*://*.google.com/*"], "js": ["content-scripts/iframe.js"] },
    {
      "matches": ["*://*.crunchyroll.com/*"],
      "js": ["content-scripts/location-change.js"]
    },
    {
      "matches": ["*://*/*"],
      "js": ["content-scripts/main-world.js"],
      "world": "MAIN"
    }
  ],
  "optional_permissions": ["https://example.com/*"]
}
```

</details>

<details>
<summary>MV3</summary>

```json
{
  "manifest_version": 3,
  "name": "wxt-demo",
  "version": "1.0.0",
  "icons": {
    "16": "icons/16.png",
    "32": "icons/32.png",
    "48": "icons/48.png",
    "128": "icons/128.png"
  },
  "permissions": ["storage", "sidePanel"],
  "default_locale": "en",
  "web_accessible_resources": [
    {
      "resources": ["iframe-src.html", "unlisted.js"],
      "matches": ["*://*.google.com/*", "*://*.example.com/*"]
    },
    {
      "resources": ["content-scripts/ui.css"],
      "use_dynamic_url": true,
      "matches": ["https://*.duckduckgo.com/*"]
    }
  ],
  "background": { "service_worker": "background.js" },
  "action": { "default_title": "Popup", "default_popup": "popup.html" },
  "options_ui": { "open_in_tab": false, "page": "options.html" },
  "sandbox": { "pages": ["example.html", "sandbox.html"] },
  "side_panel": { "default_path": "sidepanel.html" },
  "content_scripts": [
    {
      "matches": ["https://*.duckduckgo.com/*"],
      "css": ["content-scripts/automount.css"],
      "js": ["content-scripts/automount.js", "content-scripts/ui.js"]
    },
    { "matches": ["<all_urls>"], "js": ["content-scripts/example-tsx.js"] },
    { "matches": ["*://*.google.com/*"], "js": ["content-scripts/iframe.js"] },
    {
      "matches": ["*://*.crunchyroll.com/*"],
      "js": ["content-scripts/location-change.js"]
    },
    {
      "matches": ["*://*/*"],
      "js": ["content-scripts/main-world.js"],
      "world": "MAIN"
    }
  ],
  "optional_host_permissions": ["https://example.com/*"]
}
```

</details>

### Related Issue

This PR closes #2239


eupthere:
@aklinker1 Right now `optional_host_permissions` dedupe is exact-string only.  

> Would be nice if this was deduped if user already had urls in optional_host_permissions. example if I had *.app.com in optional_host_permissions theres no need to put auth.app.com because its covered by the existing one

The issue mentions deduping as a nice to have.
 
Since WXT already uses `@webext-core/match-patterns`, I checked whether it can help directly. It has URL checks (`includes(url)`) but I didn’t find a pattern-to-pattern coverage API (`covers(otherPattern)` / subset).  
 
I can either:
- implement a small coverage check in WXT now, or
- add coverage support to `@webext-core/match-patterns` first and then wire it in here.  
 
Happy to do whichever you prefer.

packages/wxt/src/core/utils/__tests__/manifest.test.ts
```diff
@@ -1870,6 +1911,63 @@ describe('Manifest Utils', () => {
       });
     });
 
+    describe('optional_host_permissions', () => {
+      it('should keep optional_host_permissions as-is for MV3', async () => {
+        const expectedOptionalHostPermissions = ['https://google.com/*'];
+        const expectedOptionalPermissions = ['cookies' as const];
```
PatrykKuniczak: this shouldn't be:
```suggestion
        const expectedOptionalPermissions = ['cookies'] as const;
```
??
'cookies' is literal anyway

or this isn't necessary at all

packages/wxt/src/core/utils/__tests__/validation.test.ts
```diff
@@ -117,5 +117,32 @@ describe('Validation Utils', () => {
 
       expect(actual).toEqual(expected);
     });
+
+    it('should return an error when "registration: optional" content scripts don\'t have matches', () => {
+      const entrypoint = fakeContentScriptEntrypoint({
+        options: {
+          registration: 'optional',
+          // @ts-expect-error
```
PatrykKuniczak:
```suggestion
          // @ts-expect-error Let's add a reason
```

packages/wxt/src/core/utils/__tests__/validation.test.ts
```diff
@@ -85,7 +85,7 @@ describe('Validation Utils', () => {
           {
             type: 'error',
             message:
-              '`matches` is required for manifest registered content scripts',
+              '`matches` is required for content scripts not registered at runtime',
```
PatrykKuniczak:
```suggestion
              '`matches` is required for content scripts, not registered at runtime',
```

packages/wxt/src/core/utils/__tests__/validation.test.ts
```diff
@@ -117,5 +117,32 @@ describe('Validation Utils', () => {
...
+          // @ts-expect-error
+          matches: null,
+        },
+      });
+      const expected = {
+        errors: [
+          {
+            type: 'error',
+            message:
+              '`matches` is required for content scripts not registered at runtime',
```
PatrykKuniczak:
```suggestion
              '`matches` is required for content scripts, not registered at runtime',
```

packages/wxt/src/core/utils/manifest.ts
```diff
@@ -616,6 +635,17 @@ function addPermission(
...
+  manifest: Browser.runtime.Manifest,
+  permission: string,
+): void {
+  manifest.optional_permissions ??= [];
+  // @ts-expect-error: Allow using strings for permissions for MV2 support
+  if (manifest.optional_permissions.includes(permission)) return;
+  // @ts-expect-error: Allow using strings for permissions for MV2 support
+  manifest.optional_permissions.push(permission);
+}
+
```
PatrykKuniczak, line 648: Maybe use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce instead. inline in https://github.com/wxt-dev/wxt/pull/2288/changes#diff-50df625c6f95cac7dcf47882681f0bc55fb27a66858fc50256f26fb541b25c43R450

packages/wxt/src/core/utils/manifest.ts
```diff
@@ -669,6 +708,17 @@ function moveHostPermissionsToPermissions(
   delete manifest.host_permissions;
 }
 
+function moveOptionalHostPermissionsToOptionalPermissions(
```
PatrykKuniczak, line 711: The same like for `addOptionalPermission`

packages/wxt/src/core/utils/validation.ts
```diff
@@ -36,7 +36,8 @@ function validateContentScriptEntrypoint(
   ) {
     errors.push({
       type: 'error',
-      message: '`matches` is required for manifest registered content scripts',
+      message:
+        '`matches` is required for content scripts not registered at runtime',
```
PatrykKuniczak:
```suggestion
        '`matches` is required for content scripts, not registered at runtime',
```

PatrykKuniczak (changes requested):

packages/wxt/src/core/utils/manifest.ts
```diff
@@ -395,13 +396,21 @@ function addEntrypoints(
     if (wxt.config.command === 'serve' && wxt.config.manifestVersion === 3) {
       contentScripts.forEach((script) => {
         script.options.matches?.forEach((matchPattern) => {
-          addHostPermission(manifest, matchPattern);
+          if (script.options.registration === 'optional') {
+            addOptionalHostPermission(manifest, matchPattern);
+          } else {
+            addHostPermission(manifest, matchPattern);
+          }
```
PatrykKuniczak, line 403: What if i want to use both in the same time?

packages/wxt/src/core/utils/__tests__/validation.test.ts
eupthere: Clarified the message at b72dfe63437652dd1b2c36de37f18f80090658d6

packages/wxt/src/core/utils/__tests__/validation.test.ts
eupthere: Clarified the message at b72dfe63437652dd1b2c36de37f18f80090658d6

packages/wxt/src/core/utils/validation.ts
eupthere: Clarified the message at 884f596f7afc39d14806d7e1802866f9b2626509

packages/wxt/src/core/utils/__tests__/validation.test.ts
eupthere: Added here 836004278fcb37e5e8886c33e77a27f17c08e11a

packages/wxt/src/core/utils/manifest.ts
eupthere: I’m not sure `reduce` makes sense in `addOptionalPermission`, since that helper is just ensuring the array exists, deduping, and mutating the manifest. Did you mean refactoring the `optionalContentScripts.forEach(...)` logic, rather than the `manifest.optional_permissions.push(permission)` part?

packages/wxt/src/core/utils/manifest.ts
eupthere: I think that’s a totally fair point, and I hadn’t considered that case. 

```ts
    matches: ['https://my-app.com/*'],
    registration: 'optional',
    // ...
});
```
The original issue’s proposed API treats registration as a single option for the whole content script, so it doesn’t currently allow choosing required vs optional behavior per match.  

If we want to support both without introducing breaking changes, maybe we could extend the API, for example by adding a new optional field to separate optional matches from required ones. That way the current API keeps working, while still leaving room for mixed-permission content scripts.

Maybe something like this..
```ts
  matches: ['https://required.com/*'],
  optionalMatches: ['https://optional.com/*'],
})
```

packages/wxt/src/core/utils/__tests__/manifest.test.ts
eupthere: Thanks for catching that. That was a type error suppression from LLM, and I should’ve caught it before opening the PR. 

I’ve replaced it with the proper type here: 8115eacbf78d1e7959309b1650073482353f292e

packages/wxt/src/core/utils/manifest.ts
PatrykKuniczak: Yeah i mean refactor `optionalContentScripts.forEach(...)` because you have nested forEach and it doesn't look good.

I think you can omit entire `addOptionalPermission`, but now i have better idea.

Let's use `.reduce` like it's in line 414 for `hostPermissions`, it'll be good enough for there.

I should add this comment above, but as i mentioned here, i want to do it in other way previously 😆

packages/wxt/src/core/utils/manifest.ts
PatrykKuniczak: > ```ts
> export default defineContentScript({
>   matches: ['https://required.com/*'],
>   optionalMatches: ['https://optional.com/*'],
> })
> ```

Yeah i was thinking about that, because we can't close our users to THIS or THIS approach, especially if on native approach, you can have both.

packages/wxt/src/core/utils/manifest.ts
eupthere: Could you suggest the change please? I can't think of how using `Array.prototype.reduce` can help with readability. The `optionalContentScript` is pretty much identical to the existing code:

```ts
  // Runtime content scripts
  const runtimeContentScripts = contentScripts.filter(
    (cs) => cs.options.registration === 'runtime',
  );
  runtimeContentScripts.forEach((script) => {
    script.options.matches?.forEach((matchPattern) => {
      addHostPermission(manifest, matchPattern);
    });
  });
```

packages/wxt/src/core/utils/manifest.ts
eupthere: @marcellino-ornelas @aklinker1 Any feedback on this approach?

docs/guide/essentials/scripting.md
```diff
@@ -27,3 +27,9 @@ export default defineContentScript({
   },
 });
 ```
+
+## Optional Host Registration
+
+When using `registration: 'optional'`, WXT adds the script's `matches` to
+`optional_host_permissions` instead of `host_permissions`. You must request host
+access before registering/executing the script at runtime.
```
aklinker1, line 35: This doesn't go here. Let's put it here instead: https://wxt.dev/guide/essentials/content-scripts.html

We should add an entire section that covers all three options.

I need to update the example from above to not use `defineContentScript`, you should use an unlisted script for that use-case.

packages/wxt/src/core/utils/manifest.ts
aklinker1: Hmm... I like `optionalMatches`. As I was reviewing this, I was starting to get confused how we would describe the difference between `runtime` and `optional`, having a separate property might make more sense.

packages/wxt/src/core/utils/manifest.ts
aklinker1: Yeah, I'm not sure what you're requesting here either @PatrykKuniczak

packages/wxt/src/core/utils/manifest.ts
aklinker1: Nice catch on this @PatrykKuniczak

packages/wxt/src/core/utils/manifest.ts
PatrykKuniczak: Ok guys, my bad `.reduce()` here, isn't good.
But IMO a little better is:
```ts
contentScripts
  .filter((cs) => cs.options.registration === 'optional')
  .flatMap(script => script.options.matches || [])
  .forEach((matchPattern) => {
    addOptionalHostPermission(manifest, matchPattern);
  });
  ```
  
  But if you think it isn't let's leave what it is.

File: packages/wxt/src/core/utils/manifest.ts (hunk 8 above)
Comment (marcellino-ornelas) [reply packages/wxt/src/core/utils/manifest.ts:403]: I like the optionalMatches here as well 🙌🏽 This is actually something were supporting right now in our custom module wasnt sure how to express this at the time of my issue tho soo left this out. Being able to support both like this is amazing and solves all of our usecases ❤️

Comment (marcellino-ornelas) [reply packages/wxt/src/core/utils/manifest.ts:403]: although how would the registration option work here now? you need `registration: 'runtime'` soo it doesn't hit the manifest and optional content scripts need the `registration: 'runtime'` but your matches may want to use `registration: 'manifest'`

Unless were saying here all `matches` goes to `manifest` and all `optionalMatches` goes to `runtime`. The only weird case would be if you wanted your `matches` configured to `runtime` and have them go to `host_permissions` which I haven't found a use case for that yet but maybe someone out there does?

**Comment by marcellino-ornelas** [comment]
The only other improvement here I would recommend is to dedupe `optional_host_permissions` 😃 heres what I mean.

**example**
In our `wxt.config.ts` file we define `optional_host_permissions` to something like this:

```ts
optional_host_permissions: [
    "https://*.internal-server.com"
]
```

This way one host gives you access to anything internal. But in our content scripts were more specific like  `https://app1.internal-server.com/some/path` because we only want it to run on certain websites.

It would be sweet if:
- If there is `optional_host_permissions` defined in wxt config already, append optionalMatches hosts instead of replace
- (this would be sweet) Only add `optionalMatches` hosts to `optional_host_permissions` if one doesnt match it already. Soo in my above example `https://app1.internal-server.com` wouldnt be added again because `https://*.internal-server.com` already covers that requirement.


No big deal if not we can easily change our custom module to remove the dups if not 😃 Although I think overall this behavior would be sweet for not only optional content scripts but also regular matches too

packages/wxt/src/core/utils/manifest.ts
```diff
@@ -625,6 +655,15 @@ function addHostPermission(
   manifest.host_permissions.push(hostPermission);
 }
 
+function addOptionalHostPermission(
+  manifest: Browser.runtime.Manifest,
+  hostPermission: string,
+): void {
+  manifest.optional_host_permissions ??= [];
+  if (manifest.optional_host_permissions.includes(hostPermission)) return;
```
marcellino-ornelas, line 663: Left this comment separately but you might want to consider using `MatchPatterns` here soo not only do you check whether its included but you can also check to see if a more wildcard host captures this host more broadly

see this comment for more details:
https://github.com/wxt-dev/wxt/pull/2288#issuecomment-4348314414

eupthere: Love the deduping idea, and thanks for the implementation tips!

I am considering using `@webext-core/match-patterns` as it's already in wxt core dependency. I just need some feedback on the pattern matching API : https://github.com/wxt-dev/wxt/pull/2288#issuecomment-4290122369

aklinker1: I'm down for deduping, but yeah, this isn't supported in that package right now. Maybe just checking if a pattern includes another pattern is enough?

packages/wxt/src/core/utils/manifest.ts
aklinker1: Hmm, yes, this does complicate that... for now, let's keep it simple. `registration` only effects how `matches` is added to the manifest. `optionalMatches` will always go to the host permissions.

packages/wxt/src/core/utils/manifest.ts
marcellino-ornelas: > I'm down for deduping, but yeah, this isn't supported in that package right now. Maybe just checking if a pattern includes another pattern is enough?

➕ we do something like this right now

```ts
/**
 * Check to see if `sub` is a subset of `primary` pattern.
 *
 * @example
 * includesPattern('https://*.myapp.com/*', 'https://logs.myapp.com/*') // true
 * includesPattern('https://logs.myapp.com/*', 'https://*.myapp.com/*') // false
 * includesPattern('https://github.com/*', 'https://github.com/*') // true
 * includesPattern('https://github.com/*', 'https://github.com/*') // false
 * includesPattern('https://github.com/*', 'https://github.com/*') // false
 * includesPattern('https://github.com/*', 'https://github.com/some/path/*') // true
 */
export const includesPattern = (primary: string, sub: string): boolean => {
    if (primary === sub) return true;

    try {
        return new MatchPattern(primary).includes(sub);
    } catch {
        return false;
    }
};

```

PR #1976, opened 2025-11-13
feat: add support for Firefox data collection permissions

suveshmoza:
### Overview

Adds support for Firefox's required `data_collection_permissions` field in `browser_specific_settings.gecko`. As of November 3rd, 2025, all new Firefox extensions must declare their data collection practices.
### Changes

- Add `FirefoxDataCollectionType` and `FirefoxDataCollectionPermissions` TypeScript types
- Update `UserManifest` to include `data_collection_permissions` in `browser_specific_settings.gecko`
- Add warning when building for Firefox without `data_collection_permissions`  #specified

### Manual Testing
Building the wxt-demo for Firefox will show a warning to add `data_collection_permissions` to manifest config.

### Related Issue

This PR closes #1975

PR #2293, opened 2026-04-23
feat: Add warning for missing Firefox extension ID in manifest

Sbrjt:
### Overview

This PR adds a warning when the Firefox extension ID is missing from the manifest to help prevent invalid builds.

https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/browser_specific_settings#description

### Manual Testing

Not sure

### Related Issue

None


packages/wxt/src/core/utils/manifest.ts
```diff
@@ -162,6 +172,8 @@ export async function generateManifest(
       "Manifest 'version' is missing. Either:\n1. Add a version in your <rootDir>/package.json\n2. Pass the version via the manifest option in your wxt.config.ts",
     );
   }
+
```
PatrykKuniczak: Let's use prettier here, i'll fix CI errors and anyway, that gap isn't necessary

PatrykKuniczak (changes requested):

Sbrjt:
Do you think we can add some default like `pkg.author.email` or ``${pkg.name}@${pkg.author.name}``?

smashedr:
> Do you think we can add some default like `pkg.author.email` or `${pkg.name}@${pkg.author.name}`?

This would be terrible, as it would clobber extension settings without the user knowing it.
This is how firefox determines what to link internal data to like the storage area.

Additionally, this field is not required for development, only for installing/signing/publishing.

It is also publicity visible and permanent once published. It can never be changed without deleting the extension and publishing an new one.

I would prefer publishing to fail if I have not set my permanent, public, extension ID yet, so that I can set a proper ID; vs having a random one published.

You can easily define browser specific settings in the config

```typescript
...(isFirefox
  ? {
      browser_specific_settings: {
        gecko: {
          id: 'auto-auth@cssnr.com',
          strict_min_version: '128.0', // webRequest.onAuthRequired asyncBlocking
          data_collection_permissions: { required: ['none'] },
        },
        gecko_android: { strict_min_version: '128.0' }, // webRequest.onAuthRequired asyncBlocking
      },
    }
  : { minimum_chrome_version: '127' }), // chrome.action.openPopup
```

But I support a warning.

PatrykKuniczak:
@smashedr Good catch, in that case we shouldn't add any default data.

Let's leave only warn

packages/wxt/src/core/utils/manifest.ts
```diff
@@ -129,6 +129,16 @@ export async function generateManifest(
     );
   }
 
+  if (
+    wxt.config.browser === "firefox" &&
+    !manifest.browser_specific_settings?.gecko?.id
+  ) {
+    wxt.logger.warn(
+      "Firefox requires extension ID for MV3 and recommends it for MV2.\n" +
+      "For more details, see: https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/browser_specific_settings"
```
aklinker1: Link to the ID section:

```suggestion
      "For more details, see: https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/browser_specific_settings#id"
```

aklinker1 (changes requested):
Looks good, but can we add a option to suppress this warning? For any MV2 extensions that don't care about it.

Similar to this PR: https://github.com/wxt-dev/wxt/pull/1976

PatrykKuniczak (approved):
LGTM

aklinker1 (approved):

aklinker1:
@Sbrjt tests are failing, I have a fix, but can't push to your PR. Can you apply this patch?

```diff
diff --git a/packages/wxt/src/core/utils/__tests__/manifest.test.ts b/packages/wxt/src/core/utils/__tests__/manifest.test.ts
index a629f26d..341f767d 100644
--- a/packages/wxt/src/core/utils/__tests__/manifest.test.ts
+++ b/packages/wxt/src/core/utils/__tests__/manifest.test.ts
@@ -1671,7 +1671,6 @@ describe('Manifest Utils', () => {

         expect(actual.version).toBe('0.0.0');
         expect(actual.version_name).toBeUndefined();
-        expect(wxt.logger.warn).toBeCalledTimes(1);
         expect(wxt.logger.warn).toBeCalledWith(
           expect.stringContaining('Extension version not found'),
         );
@@ -2103,7 +2102,6 @@ describe('Manifest Utils', () => {
         const { manifest } = await generateManifest([], buildOutput);

         expect(manifest.manifest_version).toBe(expectedVersion);
-        expect(wxt.logger.warn).toBeCalledTimes(1);
         expect(wxt.logger.warn).toBeCalledWith(
           expect.stringContaining(
             '`manifest.manifest_version` config was set, but ignored',
```

Just delete the two lines calling `expect(wxt.logger.warn).toBeCalledTimes(1)`.

Sbrjt:
@aklinker1 is it ok?

PR #2295, opened 2026-04-24
feat: Change behaviour of eslint config

PatrykKuniczak:
### Overview

This pull request updates the handling of ESLint auto-import config generation in the WXT project. The main changes are a simplification of the `enabled` option for ESLint config generation, deprecating the `'auto'` option in favor of `true`, clarifying documentation, and improving type safety by introducing a dedicated type for ESLint config versions. Test cases and internal logic are updated to reflect these changes.

**Type safety and configuration updates:**

* Introduced a new `EslintConfigVersion` type (`8 | 9`) in `types.ts` to clarify and enforce valid ESLint config version values throughout the codebase.
* Updated the `Eslintrc` and `ResolvedEslintrc` interfaces to use the new `EslintConfigVersion` type, and clarified the documentation for the `enabled` option, marking `'auto'` as deprecated and setting the default to `true`. [[1]](diffhunk://#diff-d28c2725d1b8ea5238aee0b3d8e6c8aabdee0772ea03a560fcfe685561727f0cR1685-R1708) [[2]](diffhunk://#diff-d28c2725d1b8ea5238aee0b3d8e6c8aabdee0772ea03a560fcfe685561727f0cL1715-R1717)

**Logic and default behavior improvements:**

* Changed the default for `imports.eslintrc.enabled` from `'auto'` to `true`, and updated the logic in `resolve-config.ts` to warn about and deprecate `'auto'` in favor of `true`.
* Refactored internal function signatures and usages (e.g., `getEslintConfigEntry` in `unimport.ts`) to use the new `EslintConfigVersion` type for clarity and type safety. [[1]](diffhunk://#diff-149ce6064834c07952a8a674a6a0086425d6ef2ab50e66654e8d3ab29b99e0b7R8) [[2]](diffhunk://#diff-149ce6064834c07952a8a674a6a0086425d6ef2ab50e66654e8d3ab29b99e0b7L106-R107) [[3]](diffhunk://#diff-149ce6064834c07952a8a674a6a0086425d6ef2ab50e66654e8d3ab29b99e0b7L118-R120)

**Test updates:**

* Updated and expanded ESLint config generation tests in `auto-imports.test.ts` to cover both `true` and `'auto'` values for the `enabled` option, and clarified test descriptions for ESLint 8 and 9 compatibility. [[1]](diffhunk://#diff-d7ea0c96d92d83d8a51d2d26990087bd6cf83e2e2aca4b50eea6afd86fea1355L219-R241) [[2]](diffhunk://#diff-d7ea0c96d92d83d8a51d2d26990087bd6cf83e2e2aca4b50eea6afd86fea1355L254-R258)
* Updated helper signatures and usages in tests to use `EslintConfigVersion`.

These changes improve type safety, clarify configuration options, and ensure future compatibility as ESLint evolves.


packages/wxt/src/types.ts
```diff
@@ -1585,10 +1587,11 @@ export interface Eslintrc {
    * - `true`: Same as `8`.
    * - `8`: Generate a config file compatible with ESLint 8.
    * - `9`: Generate a config file compatible with ESLint 9.
+   * - `10`: Generate a config file compatible with ESLint 10.
```
aklinker1: I don't want to have to update this with every version of eslint. Can we just update the docs isntead?

```suggestion
  * - `8`: Generate a config file compatible with ESLint < 9.
  * - `9`: Generate a flat config file compatible with ESLint &gte; 9.
```

Or should we deprecate `8 | 9` and use more descriptive strings like `"flat" | "old"`?

aklinker1 (changes requested):

packages/wxt/src/types.ts
PatrykKuniczak: That's good idea, i've been wondering about that.
Even if there'll be newer breaking version, we'll need also to change `getEslint9PlusConfigEntry` anyway.

Only thing that'll be breaking changes also for us, because we'll need to drop support for older versions, because "old" contains version <= 8, and after will be containing also v9 and v10, and because <= 8 doesn't support `flat`, we'll need to drop versions older than 8, orrrr make another flag for newer version e.g 13 as `new_breaking_name`.

I'll try to figure it out tomorrow, but let's give me your POV.

PatrykKuniczak:
@aklinker1 Should be good right now :)

aklinker1:
This PR is a breaking change as it stands now. 

I'm not going to merge it, we're making way to many breaking changes. Make it work for people who still use `8` and `9` literals.

PatrykKuniczak:
@aklinker1 Ok, i'll do a fallback and use `@deprecated` docs.

PatrykKuniczak:
@aklinker1 I've rethough and rereview it, and i won't adjust it, because it can go to `major` with #2296

Let's wait with those(3) PRs on that topic, till you'll be releasing next `major`.

It isn't that big breaking changes, ok yeah, user need to change params, but that's it.
#2296 it bigger thing and i think both of those PRs should go in one release and that'll have the biggest sense and impact.

I want to suggest to do it in that way.

aklinker1:
No, this does not require a breaking change. We cannot have a release with as many breaking changes as this upcoming one ever again. I was too lax about this, and I don't think it will go over well with developers.

Deprecate the options and continue to support them. We can remove them in the major version after this one. 

Removing deprecated APIs has a much smaller overhead for devs compared to hard cutting to a new behavior - people get warnings and can see the migration steps and take them ahead of time.

PatrykKuniczak:
@aklinker1 Yeah
All right

PatrykKuniczak:
@aklinker1 Should be good

packages/wxt/e2e/tests/auto-imports.test.ts
```diff
@@ -389,7 +450,7 @@ describe('Auto Imports', () => {
             }),
           );
 
-          await expect(runEslint(project, 8)).rejects.toMatchObject({
+          await expect(runEslint(project, 'old')).rejects.toMatchObject({
```
aklinker1: Looks like they refer to the old format as "eslintrc" and the new one as "flat" - can we update "old" -> "eslintrc"?

PatrykKuniczak:
@aklinker1 Ready to go :)

packages/wxt/e2e/tests/auto-imports.test.ts
```diff
@@ -268,6 +289,46 @@ describe('Auto Imports', () => {
...
+        },
+      });
+
+      expect(await project.pathExists('.wxt/eslint-auto-imports.mjs')).toBe(
+        true,
+      );
+      expect(await project.pathExists('.wxt/eslintrc-auto-import.json')).toBe(
+        false,
+      );
+    });
```
aklinker1: We shouldn't add 10 and deprecate it in the same PR.

packages/wxt/src/builtin-modules/unimport.ts
```diff
@@ -129,7 +130,7 @@ export function getEslint8ConfigEntry(
   };
 }
 
-export function getEslint9ConfigEntry(
+export function getEslint9PlusConfigEntry(
```
aklinker1: This can be `getEslintFlatConfigEntry` if we're moving that that naming convention.

packages/wxt/src/types.ts
```diff
@@ -1664,6 +1664,15 @@ export type EslintGlobalsPropValue =
   | 'writable'
   | 'writeable';
 
+export type EslintSupportedVersions = boolean | 'old' | 'flat';
```
aklinker1:
```suggestion
export type EslintSupportedVersions = boolean | 'eslintrc' | 'flat';
```

packages/wxt/src/types.ts
```diff
@@ -1664,6 +1664,15 @@ export type EslintGlobalsPropValue =
...
+
+/**
+ * @deprecated Use `EslintSupportedVersions` (`'old'`/`'flat'`) instead. Legacy
```
aklinker1:
```suggestion
 * @deprecated Use `EslintSupportedVersions` (`'eslintrc'`/`'flat'`) instead. Legacy
```

packages/wxt/src/types.ts
```diff
@@ -1664,6 +1664,15 @@ export type EslintGlobalsPropValue =
...
+ *   numeric values and 'auto' are supported for backwards compatibility and
+ *   will be removed in the next major version.
+ */
+export type EslintLegacySupportedVersions = 8 | 9 | 10 | 'auto';
```
aklinker1:
```suggestion
export type EslintLegacySupportedVersions = 8 | 9 | 'auto';
```

packages/wxt/src/types.ts
```diff
@@ -1673,19 +1682,24 @@ export interface Eslintrc {
    *   config file.
    * - `true`: Same as `'auto'`.
    * - `false`: Don't generate the file.
-   * - `8`: Generate a config file compatible with ESLint 8.
-   * - `9`: Generate a config file compatible with ESLint 9.
-   *
+   * - `old`: Generate a config file compatible with ESLint &lte; 8.
```
aklinker1:
```suggestion
   * - `eslintrc`: Generate a config file compatible with ESLint &lte; 8.
```

packages/wxt/src/types.ts
```diff
@@ -1673,19 +1682,24 @@ export interface Eslintrc {
...
+   * - `flat`: Generate a config file compatible with ESLint &gte; 9.
+   * - `8`: Same as `old`.
+   * - `9`: Same as `flat`.
+   * - `10`: Same as `flat`.
```
aklinker1:
```suggestion
```

packages/wxt/src/types.ts
```diff
@@ -1673,19 +1682,24 @@ export interface Eslintrc {
...
+   *
+   * @deprecated Values `8`, `9`, `10`, and `'auto'` are legacy aliases and will
+   *   be removed in the next major version.
```
aklinker1:
```suggestion
   * @deprecated Values `8`, `9`, and `'auto'` are legacy aliases and will
   *   be removed in the next major version.
```

aklinker1 (changes requested):
I'll go ahead and implement the changes for you, since it took so long to get to...

aklinker1:
OK, so after cleaning some things up, I'm not sure this PR is meaningful.

All we're doing is renaming `8 | 9` to `'old' | 'flat'` - but what if a new config file version is released in the future? What would we call it in v11? I would call it `'flat11'`... or just `11` because people don't need to know what the type of config file is called, just a compatible version. And then we're back to the numbers.

The PR description says it resolves  #1981, but that PR was about type errors, nothing to do with eslint 10 support.

Plus WXT already supports eslint 10 - https://github.com/wxt-dev/wxt/blob/caa3dceb91f67798081d9f846f348f11bcec7ba9/packages/wxt/src/builtin-modules/unimport.ts#L118-L119

So I would push to:

1. Revert the changes to string names, keep using the version numbers to represent the config file version compatibility
2. Update the JSDoc for the version numbers to include the format name
3. Maybe add a type called `EslintConfigFileVersions = 8 | 9` so it's clear in the code we're not talking about the version of eslint, but the config file format.
4. Keep the deprecation of `"auto"`, that's valid.
5. Keep the change for the default value to `true`.

aklinker1 (approved):
@PatrykKuniczak I've made lots of changes to this PR, see comments above. Feel free to merge if you are happy with the changes, or let's talk about it some more.

packages/wxt/e2e/tests/auto-imports.test.ts
PatrykKuniczak: Yeah, i've missed that on review, lol

PatrykKuniczak:
@aklinker1 I think that's not bad, but `9` for version 10, 11 and etc. isn't meaningful.
That's why i've used `old` and `flat`.

If 11 or 12 will be released and it won't be support 'flat' config, we can create new key for that or event move `flat` to `old ` and drop current `old`(<=8 for then) and add support for newest versions.

But maybe, OK let's leave only booleans and numbers there it'll be the most meaningful.

PatrykKuniczak:
@aklinker1 Seems, you've broken tests.

packages/wxt/src/types.ts
```diff
@@ -1682,28 +1682,30 @@ export type EslintGlobalsPropValue =
   | 'writable'
   | 'writeable';
 
+export type EslintConfigVersion = 8 | 9;
```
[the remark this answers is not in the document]
PatrykKuniczak: That's what i was thining about.

With approach `old` and `flat` we'll be able to cover couple of versions in one key, but with current approach we need to add `10` there, even if `9` works the same.

PatrykKuniczak:
@aklinker1 I've changed title and generate new description, i hope now it's more meaningful, previously it was bad indeed.

PatrykKuniczak:
@aklinker1 
image


Without `break`, `true` will go each time `auto` will be used, but i think it's expected, yeah?

packages/wxt/src/types.ts
```diff
@@ -1682,28 +1682,30 @@ export type EslintGlobalsPropValue =
...
+
 export interface Eslintrc {
   /**
-   * When true, generates a file that can be used by ESLint to know which
-   * variables are valid globals.
+   * Determines if and in what format a config file will be generated to inform
+   * ESLint of unimport globals.
```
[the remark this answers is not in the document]
aklinker1: Nope, it should be "unimport"

aklinker1:
> Without break, true will go each time auto will be used, but i think it's expected, yeah?

Correct, since they're equivalent.

packages/wxt/e2e/tests/auto-imports.test.ts
```diff
@@ -216,25 +217,28 @@ describe('Auto Imports', () => {
...
   });
 
-  describe('eslintrc', () => {
-    it('"enabled: true" should output a JSON config file compatible with ESlint of package.json', async () => {
-      const project = new TestProject();
-      project.addFile('entrypoints/popup.html', `<html></html>`);
+  describe('eslint config', () => {
+    it.each([true, 'auto'] as const)(
+      '"enabled: %s" should output a JSON config file compatible with ESlint of package.json',
+      async (enabled) => {
```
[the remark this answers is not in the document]
aklinker1: Will fix

packages/wxt/src/types.ts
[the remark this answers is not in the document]
PatrykKuniczak: @aklinker1 Response here and it should be ready to merge.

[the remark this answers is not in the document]
aklinker1: Looks like this ran before you updated the PR description?

> > OK, so after cleaning some things up, I'm not sure this PR is meaningful.
> > 
> > All we're doing is renaming 8 | 9 to 'old' | 'flat' - but what if a new config file version is released in the future? What would we call it in v11? I would call it 'flat11'... or just 11 because people don't need to know what the type of config file is called, just a compatible version. And then we're back to the numbers.
> 
> I think that's not bad, but `9` for version 10, 11 and etc. isn't meaningful.
> That's why i've used `old` and `flat`.
> 
> If 11 or 12 will be released and it won't be support 'flat' config, we can create new key for that or event move `flat` to `old ` and drop current `old`(<=8 for then) and add support for newest versions.
> 
> But maybe, OK let's leave only booleans and numbers there it'll be the most meaningful.

I don't know how eslint's config will change in the future, if they'll continue calling it "flat" or just introduce a breaking change to how globals are defined. I don't think enum values are any more sustainable in the future for this reason, so I don't see a big enough reason to change this option. The update JSDoc for the option make it quite clear that `9` supports flat config files for eslint 9 and above.

Every deprecation and future breaking change means people will have to make changes and I don't see enums being that much better in the long term to make people switch to them eventually.

PatrykKuniczak:
@aklinker1 All right, thanks :)

PR #2296, opened 2026-04-24
feat!: Change default behavior of `getUnimportEslintOptions` true flag

PatrykKuniczak:
### Overview

Changed behavior of `getUnimportEslintOptions` to be current version of `eslint`  from `package.json` on `true` flag.

I hope this change is good, i think it's more meaningful to have latest version on `true` most of packages works like this.
If you enabling sth, you expect the latest version of that package.
IMO it has sense.

### Manual Testing

Let's see if tests passed


aklinker1 (changes requested):
Fix the diffs then I'll re-review. I'm OK with this change.

PatrykKuniczak:
@aklinker1 There's no diffs, what i'm supposed to do?

aklinker1:
This is a breaking change, it needs to be merged into the `major` branch.

PR #2297, opened 2026-04-24
fix: eslint auto import type declaration

PatrykKuniczak:
### Overview

Fixed abandoned #1981

Blocked by #2296 #2295 (After those, i'll adjust code to version 10)

### Manual Testing

Let's see if tests pass

PR #2298, opened 2026-04-24
chore: Change `wxt-demo` eslint extension

PatrykKuniczak:
### Overview

Changed to show #2297 works well and have one more place with `.ts`, with that we can be closer to fully `.ts` support on project.

Blocked by #2297 

### Manual Testing

Let's check if tests pass and see in IDE if there's no error

### Related Issue
#1981


aklinker1:
How do I test this?

PatrykKuniczak:
@aklinker1 let's pull those changes and see there's error like in #1981 and then #2297 needs to be merged or you can cherry-pick those changes and then see, there's no error for `.ts` extension

PR #2299, opened 2026-04-24
fix: `@types/chrome` version of `browser`

PatrykKuniczak:
### Overview

Add `^` to `@chrome/types` of `browser` package.json, to match bun.lock file.
Or if you really want without `^` i need to run `bun i` and let's merge new version of that file.

### DIFF
--- a/packages/browser/package.json
+++ b/packages/browser/package.json
@@ -25,7 +25,7 @@
     "src"
   ],
   "devDependencies": {
-    "@types/chrome": "0.1.40",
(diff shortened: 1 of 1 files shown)


aklinker1:
No, I want it a specific version. We should run a command during the action that updates the package to run `bun i` or something and commit the lockfile change

aklinker1:
Yeah, that script was the problem. Closing in favor of https://github.com/wxt-dev/wxt/pull/2310

PR #2300, opened 2026-04-24
ci: use --frozen-lockfile in action.yml for bun

PatrykKuniczak:
To avoid conflicts between `.lock` and `.json` of package, we need to have that.

Now there's a problem on `main` because there's versions mismatch.

Blocked by #2299


PatrykKuniczak:
@aklinker1 We should increase timings for those tests which failed

aklinker1 (approved):

aklinker1:
Actually, this wasn't a meaningful change. We're using `bun ci`, not `bun install`. `bun ci` uses `--frozen-lockfile` by default.

PR #2308, opened 2026-04-27
fix(browser): add IOS to PlatformOs enum

rxliuli:
## Summary
- Add `IOS = "ios"` to `Runtime.PlatformOs` enum in `@wxt-dev/browser` type definitions
- Safari on iOS/iPadOS reports `runtime.getPlatformInfo().os` as `"ios"`, but the type was missing this value, causing TypeScript errors when comparing against it

## Reference
- [MDN: runtime.PlatformOs - ios](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/PlatformOs#ios)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/browser/src/gen/index.d.ts
+++ b/packages/browser/src/gen/index.d.ts
@@ -8888,6 +8888,8 @@ export namespace Browser {
             OPENBSD = "openbsd",
             /** Specifies the Fuchsia operating system. */
             FUCHSIA = "fuchsia",
+            /** Specifies the iOS operating system. */
+            IOS = "ios",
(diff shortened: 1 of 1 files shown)


aklinker1:
This package is generated based on `@types/chrome`, make your changes there instead.

rxliuli:
> This package is generated based on `@types/chrome`, make your changes there instead.

done. https://github.com/DefinitelyTyped/DefinitelyTyped/pull/74923

PR #2316, opened 2026-04-29
chore: add missing tsdown dependencies

howlettt:
### Overview

Possible fix for the failing Windows tests: a couple packages were using tsdown but didn't have it in their devDependencies


aklinker1 (approved):
Thanks, saw these windows failures, couldn't track down the issue so I was just ignoring it lol. Thought I double checked all these!

Issue #2286, opened 2026-04-19
`wxt submit init` not working


MrSquaare:
### Describe the bug

I want to use `wxt submit init` to setup my secrets

I am getting the following error:

```shell
npm run submit-init

> wxt-starter@0.0.0 submit-init
> wxt submit init

file:///Users/mrsquaare/Programming/Projects/MrSquaare/wxt-submit-init-playground/node_modules/cac/dist/index.js:394
                if (maximumArgsCount < this.cli.args.length) throw new CACError(`Unused args: ${this.cli.args.slice(maximumArgsCount).map((arg) => `\`${arg}\``).join(", ")}`);
                                                                   ^

CACError: Unused args: `init`
    at Command.checkUnusedArgs (file:///Users/mrsquaare/Programming/Projects/MrSquaare/wxt-submit-init-playground/node_modules/cac/dist/index.js:394:54)
    at CAC.runMatchedCommand (file:///Users/mrsquaare/Programming/Projects/MrSquaare/wxt-submit-init-playground/node_modules/cac/dist/index.js:619:11)
    at file:///Users/mrsquaare/Programming/Projects/MrSquaare/wxt-submit-init-playground/node_modules/wxt/dist/cli/index.mjs:11:11
    at ModuleJob.run (node:internal/modules/esm/module_job:430:25)
    at async onImport.tracePromise.__proto__ (node:internal/modules/esm/loader:661:26)
    at async asyncRunEntryPointWithESMLoader (node:internal/modules/run_main:101:5)

Node.js v24.14.1
```

### Reproduction

[wxt-submit-init-playground.zip](https://github.com/user-attachments/files/26866920/wxt-submit-init-playground.zip)

### Steps to reproduce

1. npm install
2. npm run submit-init

### System Info

```shell
System:
    OS: macOS 26.4.1
    CPU: (10) arm64 Apple M1 Max
    Memory: 4.36 GB / 64.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 24.14.1 - /Users/mrsquaare/.local/share/mise/installs/node/24/bin/node
    npm: 11.12.1 - /Users/mrsquaare/.local/share/mise/installs/node/24/bin/npm
    pnpm: 10.33.0 - /Users/mrsquaare/.local/share/mise/installs/node/24/bin/pnpm
  Browsers:
    Chrome: 147.0.7727.102
    Firefox: 149.0.2
    Safari: 26.4
  npmPackages:
[...]

Issue #2129, opened 2026-02-14
wxt submit init does not work


orangejuiceext:
### Describe the bug

```
wxt submit init
[ble: exit 1]
```

Nothing else. I did search around and saw that you're working on updating the dependency on the submit backend, but figured I'd file this cause this is a new level of just not working at all. Feel free to close if I'm in the wrong.

### Reproduction

```
git clone https://github.com/OrangeJuiceExtension/OrangeJuice
wxt submit init
```

### System Info

```shell
bunx envinfo --system --browsers --binaries --npmPackages wxt,vite

  System:
    OS: macOS 26.3
    CPU: (16) arm64 Apple M4 Max
    Memory: 110.41 MB / 64.00 GB
    Shell: 5.3.9 - /opt/homebrew/bin/bash
  Binaries:
    Node: 25.6.1 - /opt/homebrew/bin/node
    npm: 11.9.0 - /opt/homebrew/bin/npm
    bun: 1.3.9 - /opt/homebrew/bin/bun
    Deno: 2.6.9 - /opt/homebrew/bin/deno
  Browsers:
    Brave Browser: 144.1.86.148
    Chrome: 144.0.7559.134
    Firefox: 89.0
    Safari: 26.3
  npmPackages:
    wxt: ^0.20.17 => 0.20.17
```

### Used Package Manager

bun

### Validations

- [x] Read the [Contributing Guidelines](https://github.com/wxt-dev/wxt/blob/main/CONTRIBUTING.md).
- [x] Read the [docs](https://wxt.dev/guide/installation.html).
- [x] Check that there isn't [already an issue](https://github.com/wxt-dev/wxt/issues) that reports the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/wxt-dev/wxt/discussions) or join our [Discord Chat Server](https://discord.gg/ZFsZqGery9).
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

thcl-odoo:
I also have an issue with `bun wxt submit` since v0.20.17

I noticed the issue in a GitHub Action runner, but after seeing this ticket, I tried `wxt submit init` and the issue is the same.
I have a little more luck because I have logs, so here they are:

```
Run bun wxt submit \
  
[...]

PR #2317, opened 2026-04-30
fix: Modify command to support variadic positional args

laplace-anon:
Updated command registration to allow variadic positional arguments for subcommands.

### Overview

`wxt submit <subcommand>` (e.g. `wxt submit init`) currently fails with:

```
CACError: Unused args: init at Command.checkUnusedArgs (node_modules/cac/dist/index.js:394:54) at CAC.runMatchedCommand (node_modules/cac/dist/index.js:619:11) at file:.../node_modules/wxt/dist/cli/index.mjs:11:11
```

Ref: https://github.com/wxt-dev/wxt/issues/2286

`createAliasedCommand` registers the alias as `cli.command(name)` (no positional arg spec) and then forwards `process.argv` to the underlying CLI. `.allowUnknownOptions()` only relaxes _flag_ checks, not positional args. This worked by accident on `cac@6`, but `wxt`'s declared range `^6.7.14 || ^7.0.0` now resolves to `cac@7.0.0`, which added a strict `checkUnusedArgs` (cacjs/cac#135) that throws before the action ever runs.

The fix is one line: declare the alias as `` `${name} [...args]` `` so cac treats positional args as variadic and forwards them through.

### Manual Testing

Before:
```sh
$ bun wxt submit init
CACError: Unused args: `init`
```

After:

```
$ pnpm wxt submit init
publish-extension/4.0.5
Usage: $ publish-extension init
...
```

### Related Issue

This PR closes: 

- #2286
- #2129 (the wxt submit init failure that was previously misdiagnosed as a publish-browser-extension v3→v4 issue but reproduces cleanly with cac@7 + publish-browser-extension@4.0.4)
```
╰─>$ bun why publish-browser-extension
publish-browser-extension@4.0.5
  └─ wxt@0.20.25 (requires ^2.3.0 || ^3.0.2 || ^4.0.4)
     ├─ dev wxt-react-starter (requires ^0.20.25)
     ├─ peer @wxt-dev/auto-icons@1.1.1 (requires >=0.19.0)
     │  └─ dev wxt-react-starter (requires ^1.1.1)
     └─ peer @wxt-dev/module-react@1.2.2 (requires >=0.19.16)
        └─ dev wxt-react-starter (requires ^1.2.2)
```

### DIFF
--- a/cspell.yml
+++ b/cspell.yml
@@ -11,6 +11,7 @@ ignorePaths:
   - patches/**
   - '**/*.snap'
   - docs/assets/extension-showcase.yml
+  - packages/*/stats.html
 words:
   - Aabid
   - aabidk
@@ -21,6 +22,7 @@ words:
   - buildc
   - bunx
   - cachable
+  - cacjs
   - charmbracelet
   - chromedriver
   - Cira
--- a/packages/wxt/src/cli/cli-utils.ts
+++ b/packages/wxt/src/cli/cli-utils.ts
@@ -82,8 +82,12 @@ export function createAliasedCommand(
   bin: string,
   docsUrl: string,
 ) {
+  // Declare a variadic positional arg so cac forwards subcommands like `wxt
+  // submit init` instead of rejecting them as unused args. `.allowUnknownOptions`
+  // only relaxes flag checks, not positional args. Required since cac@7, which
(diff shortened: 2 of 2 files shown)


aklinker1 (approved):
Thanks, good find

PR #2343, opened 2026-05-09
fix: avoid errors when files are removed during build

### LINKED ISSUE #1735: `printFileList` causes the build to fail if files are deleted before it runs
### Describe the bug

I'm using `@sentry/vite-plugin` to upload the source maps during the build process.

Given that source maps considerably increase the bundle size, I want to use the plugin's [sourcemaps.filesToDeleteAfterUpload](https://www.npmjs.com/package/@sentry/vite-plugin#sourcemapsfilestodeleteafterupload) option to delete source maps after they've been uploaded to Sentry.

Enabling that causes my builds to fail with the following error: `ENOENT: no such file or directory, lstat '/Users/my-user/Developer/Project/.output/chrome-mv3/background.js.map'`.

At first, I thought the plugin was causing the issue, so I added a hook to delete the source maps after the build finished. However, this produced the same result.

I searched the WXT codebase for occurrences of `lstat` and found it is used by the [printFileList](https://github.com/wxt-dev/wxt/blob/1272e2ec742e0b357c4e5858e5324b4b428aebe3/packages/wxt/src/core/utils/log/printFileList.ts#L23) function. I suspect the `files` parameter is passed before the build finishes, so it contains elements that no longer exist, which causes `fs.lstat(file)` to throw an error, but I'm not sure how to fix it or disable it.

Any help would be appreciated!

```ts
hooks: {
  'build:done': ({ config }) => {
    if (config.mode === 'production') {
      rimrafSync('.output/**/*.js.map', { glob: true });
      console.log('Source maps deleted');
    }
  },
},
```

### Reproduction

[Build-Issue-Reproduction.zip](https://github.com/user-attachments/files/20717706/Build-Issue-Reproduction.zip)

### Steps to reproduce

1. Enable source map generation under `wxt.config.ts > defineConfig > vite > build`
2. Add a hook that deletes the source maps after the build
4. Run `pnpm build`
5. The build fails with the error described earlier

The reproduction I attached already has source maps generation enabled and the post-build hook that deletes them.

### System Info

```shell
System:
  OS: macOS 15.5
  CPU: (11) arm64 Apple M3 Pro
  Memory: 115.98 MB / 18.00 GB
  Shell: 5.9 - /bin/zsh
Binaries:
  Node: 22.16.0 - ~/.local/share/mise/installs/node/22.16.0/bin/node
  npm: 11.4.1 - ~/.local/share/mise/installs/node/22.16.0/bin/npm
  pnpm: 10.4.1 - ~/.local/share/mise/installs/node/22.16.0/bin/pnpm
  Watchman: 2025.05.26.00 - /opt/homebrew/bin/watchman
Browsers:
  Brave Browser: 137.1.79.123
  Safari: 18.5
npmPackages:
  wxt: ^0.20.6 => 0.20.7
```

### Used Package Manager

pnpm

### Validations

howlettt:
### Overview

Added a try catch around the `lstat` call in `printFileList` to avoid the build failing when files are deleted during the build

### Manual Testing

Tested a build of the new package using the `@sentry/vite-plugin` with the `filesToDeleteAfterUpload` option. You can see the new warnings & missing file sizes for the map files in the screenshot below

{3B7A6420-5ABA-49A3-90ED-951A31049DFE}

### Related Issue

This PR closes #1735


aklinker1 (approved):
Looks good, ~~looks like checks will fail, type error probably.~~ 

Edit: I'll resolve it.

PR #2380, opened 2026-05-15
docs: clarify SPA navigation setup with locationWatcher.run()

mrumerf:
Added the missing `ctx.locationWatcher.run()` call to the SPA navigation example.

The `wxt:locationchange` event only fires after the location watcher is started, which was not previously shown in the documentation.

### Overview

Adds the missing `ctx.locationWatcher.run()` call to the SPA navigation documentation and examples.

Previously, the docs showed how to listen for the `wxt:locationchange` event, but did not mention that the location watcher must first be started with `ctx.locationWatcher.run()`. Without this call, the event is never dispatched.

### Manual Testing

1. Follow the updated SPA documentation example.
2. Add a `console.log` inside the `wxt:locationchange` listener.
3. Navigate between routes in an SPA site.
4. Verify that the event fires correctly after calling `ctx.locationWatcher.run()`.

### Related Issue

N/A

### DIFF
--- a/docs/guide/essentials/content-scripts.md
+++ b/docs/guide/essentials/content-scripts.md
@@ -734,7 +734,7 @@ const watchPattern = new MatchPattern('*://*.youtube.com/watch*');
 export default defineContentScript({
   matches: ['*://*.youtube.com/*'],
   main(ctx) {
-    ctx.locationWatcher.run()
+    ctx.locationWatcher.run();
     ctx.addEventListener(window, 'wxt:locationchange', ({ newUrl }) => {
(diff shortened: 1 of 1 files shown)


mrumerf (commented):
Applied the required formatting fixes and updated the PR.

PatrykKuniczak:
> Applied the required formatting fixes and updated the PR.

Still doesn't work

mrumerf (commented):
just added ";" to pass prettier test

aklinker1 (approved):

PR #2403, opened 2026-06-04
docs: document PerBrowser options for entrypoints (#2304)

### LINKED ISSUE #2304: Document entrypoint `PerBrowser` options
This is an undocumented feature right now.

```ts
  matches: {
    chrome: ["..."],
    firefox: ["..."],
  },
  // ...
})
```

dashitongzhi:
Closes #2304.

Adds a 'Per-Browser Options' section to docs/guide/essentials/target-different-browsers.md
explaining `PerBrowserOption<T>` / `PerBrowserMap<T>` usage on entrypoint
config fields like `matches`, `runAt`, and `defaultArea`.

- Explains the type shape: `T | { [browser: TargetBrowser]: T }`
- Lists the entrypoint fields that accept it (matches, runAt, matchAboutBlank, excludeMatches, includeGlobs, excludeGlobs, allFrames, matchOriginAsFallback, cssInjectionMode, registration, defaultTitle, defaultArea, persistent)
- Shows both forms (single value, per-browser map) with concrete `defineContentScript` examples
- Notes the fallback behavior when a build target is omitted from the map
- Adds a tip about mixing both forms inside the same entrypoint
- Cross-references Manifest config for fields that don't fit the per-key shape (e.g. `default_icon`)

### DIFF
--- a/docs/guide/essentials/target-different-browsers.md
+++ b/docs/guide/essentials/target-different-browsers.md
@@ -81,49 +81,25 @@ Alternatively, you can use the [`filterEntrypoints` config](/api/reference/wxt/i
 
 ## Per-Browser Options
 
-Many entrypoint config options accept either a single value (applied to every browser build) or a per-browser map (a different value per build target). These are typed as `PerBrowserOption<T>` in the WXT source:
-
-```ts
-export type PerBrowserOption<T> = T | PerBrowserMap<T>;
-export type PerBrowserMap<T> = { [browser: TargetBrowser]: T };
-```
-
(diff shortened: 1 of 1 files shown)


PatrykKuniczak:
@dashitongzhi I think it's alright, but wait for Aaron approval.

But i think we should make `TargetBrowser` not `string` but put a literal with browsers name there:
https://github.com/wxt-dev/wxt/blob/7905cb379b1210fe01a456bfed35401a2996d598/packages/wxt/src/types.ts#L569

Feel free to open another PR.

aklinker1 (changes requested):
This PR is a bit more verbose than it needs to be. I think a single paragraph with one example code block is enough. We shouldn't be re-defining types and listing options either, just link to the API reference for those kinds of things, though I don't think they're needed here to understand the feature.

I like the changes from https://github.com/wxt-dev/wxt/pull/2427/changes, but they put the code in the wrong place. It should go where you put it, with a heading like yours, but simplified docs like their PR.

aklinker1 (approved):

PR #2414, opened 2026-06-13
feat(analytics): add posthog as an analytics provider

creeperkatze:
### Overview

Adds a PostHog provider to the analytics package. Uses PostHog's REST capture API (/i/v0/e/) directly via fetch instead of importing posthog-js, keeping it MV3 compliant with no remote code. Supports identify, page, and track events.

Also adds missing debug logs to the Google Analytics 4 provider and removes unnecessary optional chaining on config.

### Manual Testing

1. Install @wxt-dev/analytics and add the posthog provider to app.config.ts:
	````js
	import { posthog } from '@wxt-dev/analytics/providers/posthog';
	
	posthog({ apiKey: 'phc_...', apiHost: 'https://us.i.posthog.com' })
	````
2. Enable analytics via the enabled config or analytics.setEnabled(true).
3. Trigger events (e.g. via analytics.autoTrack(document)).
4. Confirm events appear in PostHog's Activity > Live Events dashboard.


aklinker1 (approved):
Nice, thanks for the PR!

PR #2415, opened 2026-06-14
fix: `wxt clean` removes `.wxt/` directory at project root

### LINKED ISSUE #2413: `wxt clean` does not delete `.wxt` folder when located in the root directory
### Describe the bug

Looking at the [source code](https://github.com/wxt-dev/wxt/blob/9a638449d587f48c2821ca32577c5758346c3290/packages/wxt/src/core/clean.ts#L40-L43), it seems that the intent is for the `.wxt` folder to be deleted even when located in the root directory of a project. However, `wxt clean` does not currently delete the folder. I tested placing a `.wxt` folder within another folder (e.g. `<root>/dummy/.wxt` and the deletion does go through.

### Reproduction

Run `wxt clean` in a project containing a `.wxt` folder in the project's root directory.

### Steps to reproduce

_No response_

### System Info

```shell
npx envinfo --system --browsers --binaries --npmPackages wxt,vite

  System:
    OS: Windows 11 10.0.26200
    CPU: (4) x64 11th Gen Intel(R) Core(TM) i3-1115G4 @ 3.00GHz
    Memory: 1.87 GB / 11.70 GB
  Binaries:
    Node: 22.14.0 - C:\Program Files\nodejs\node.EXE
    npm: 10.9.2 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Chrome: 128.0.6613.138
    Edge: Chromium (146.0.3856.84)
    Internet Explorer: 11.0.26100.8115
  npmPackages:
    vite: ^8.0.3 => 8.0.16 
    wxt: ^0.20.20 => 0.20.26
```

### Used Package Manager

npm

### Validations

LUIDevo:
### Overview
`wxt clean` never deleted a `.wxt` directory located at the project root. The `**/.wxt` line requires at least one parent path, so only nested `.wxt` directories matched. Replace it with `{,**/}.wxt`, which matches both root and nested `.wxt`. 

(The same result is achievable by adding `'.wxt'` alongside `'**/.wxt'`, but that takes an extra line.)

### Manual Testing

Create a directory '.wxt' in the project root. By running 'wxt clean', the .wxt directory gets deleted, along with any nested `.wxt` directory. Before this change, the root '.wxt' directory would not be deleted, and now it does along with the other files.

### Related Issue

This PR closes #2413

### DIFF
--- a/packages/wxt/src/core/clean.ts
+++ b/packages/wxt/src/core/clean.ts
@@ -39,7 +39,7 @@ export async function clean(config?: string | InlineConfig) {
   const tempDirs = [
     'node_modules/.vite',
     'node_modules/.cache',
-    '**/.wxt',
+    '{,**/}.wxt',
     `${path.relative(root, wxt.config.outBaseDir)}/*`,
   ];
   wxt.logger.debug(


aklinker1:
Interesting, it works for non-hidden directories, but fails for hidden ones!

```sh
$ bun repl
❯ const { glob } = await import("tinyglobby")
❯ await glob("**/node_modules", { onlyDirectories: true, expandDirectories: false })
["node_modules/"]
❯ await glob("**/.output", { onlyDirectories: true, expandDirectories: false })
[]
❯ await glob("{,**/}.output", { onlyDirectories: true, expandDirectories: false })
[ ".output/" ]
```

Your solution does work

aklinker1 (approved):

PR #2438, opened 2026-06-27
test: skip yarn tests when yarn is missing

### LINKED ISSUE #2371: 2 tests for yarn doesn't pass on windows `by default`
### Describe the bug

If somebody(like me) doesn't have `yarn` installed globally.

For `packages/wxt` those 2 tests doesn't pass:

Image

To fix it, i need to install yarn globally (Only sad reaction)

To make better DX, we can place `yarn` as a `devDeps` inside this package.
But the big problem is:

Image

And it seems not supported anymore:
Latest version have over 2 years:
https://www.npmjs.com/package/yarn

There's no move in repo:
https://github.com/yarnpkg/yarn/commits/master/

I think good idea is to get rid of this package and those tests.

PS: 
OKK yarn drops support for version 1 and changed repo:
https://github.com/yarnpkg/berry/releases

IT'S `berry` right now???

### Used Package Manager

yarn

### Validations

cookesan:
### Overview

Skips the Yarn package-manager tests when the `yarn` command is not available on `PATH`. The same tests still run anywhere Yarn is present, including CI environments that install it.

### Manual Testing

- `bun run --filter wxt test run yarn.test.ts`
- `bun run --filter wxt test run yarn.test.ts` with the Yarn install directory removed from `PATH`
- `bun run --filter wxt test run`

### Related Issue

Closes #2371

### DIFF
--- a/packages/wxt/src/core/package-managers/__tests__/yarn.test.ts
+++ b/packages/wxt/src/core/package-managers/__tests__/yarn.test.ts
@@ -1,12 +1,15 @@
 import { describe, expect, it } from 'vitest';
+import { spawnSync } from 'node:child_process';
 import path from 'node:path';
 import { yarn } from '../yarn';
+
+const itWithYarn = hasYarnInstalled() ? it : it.skip;
 
 describe('Yarn Package Management Utils', () => {
   describe('listDependencies', () => {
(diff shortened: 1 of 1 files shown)


aklinker1:
Duplicate of #2441

PR #2441, opened 2026-06-28
test: skip package manager tests when binaries is missing

### LINKED ISSUE #2371: 2 tests for yarn doesn't pass on windows `by default`

cookesan:
## Summary

- Skip the yarn package-manager test suite when the `yarn` binary is not available.
- Keep the existing assertions unchanged for environments with yarn installed.

Fixes #2371.

## Testing

- `bun run --filter wxt test src/core/package-managers/__tests__/yarn.test.ts`
- `PATH=/Users/terry/.bun/bin:/usr/bin:/bin bun run --filter wxt test src/core/package-managers/__tests__/yarn.test.ts`
- `bun run prettier --check packages/wxt/src/core/package-managers/__tests__/yarn.test.ts`
- `git diff --check`

### DIFF
--- a/packages/wxt/src/core/package-managers/__tests__/pnpm.test.ts
+++ b/packages/wxt/src/core/package-managers/__tests__/pnpm.test.ts
@@ -2,10 +2,11 @@ import spawn from 'nano-spawn';
 import path from 'node:path';
 import { beforeAll, describe, expect, it } from 'vitest';
 import { pnpm } from '../pnpm';
+import { describeWithBin } from '../../utils/testing/fixtures';
 
 process.env.WXT_PNPM_IGNORE_WORKSPACE = 'true';
 
-describe('PNPM Package Management Utils', () => {
+describeWithBin('pnpm', 'PNPM Package Management Utils', () => {
   describe('listDependencies', () => {
     const cwd = path.resolve(__dirname, 'fixtures/simple-pnpm-project');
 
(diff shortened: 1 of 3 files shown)


aklinker1 (approved):

PatrykKuniczak:
@cookesan Nice.

`Bun` shouldn't be omit because it's necessary right now, and it's enforced by `engines` in `package.json` 
But `npm` also can use that `describeWithBin`, because(but there's a small change) somebody don't use `node` at all, and use only `bun`.

It isn't the case for now, because Node is only option for much cases, e.g Angular(of course, there's workaround, but still).

But maybe someday we'll need to use your func also for node, but that's very edge case for now.

PR #2445, opened 2026-06-28
feat: add storage metadata watchers

### LINKED ISSUE #1988: [FEAT] support watch meta change in storage
### Feature Request

Now we only support to watch the value of the local storage. But can't watch the change of the metadata of local storage variable.

Now we have to add `$` at the end of the KEY `local:someKey$` to watch its metadata. But in the future, the underline implementation of the meta may change, and may should not add `$`.

This is the work around of my code

```ts
  // Add $ to the key to get the meta key for LAST_SYNCED_CONFIG_STORAGE_KEY
  const metaKey = `local:${LAST_SYNCED_CONFIG_STORAGE_KEY}$`

  const unwatch = storage.watch<LastSyncedConfigMeta>(metaKey, (_, newMeta) => {
    const meta = newMeta
    setAtom(meta?.lastSyncedAt ?? null)
  })
```

cookesan:
## Summary
- add `storage.watchMeta` and `item.watchMeta` for listening to metadata changes without using the internal metadata key
- normalize removed metadata to `{}` so watcher callbacks match `getMeta`
- document the new APIs and cover storage areas plus defined items in tests

## Tests
- `bun run --filter @wxt-dev/storage test index.test.ts`
- `bun run --filter @wxt-dev/storage check`
- `bun run --filter @wxt-dev/storage build`
- `bun run docs:build`
- `bun run check`
- `bunx prettier --check docs/storage.md packages/storage/src/index.ts packages/storage/src/__tests__/index.test.ts`
- `git diff --check`

Fixes #1988

### DIFF
--- a/packages/storage/src/index.ts
+++ b/packages/storage/src/index.ts
@@ -64,7 +64,9 @@ function createStorage(): WxtStorage {
     value ?? fallback ?? null;
 
   const getMetaValue = (properties: any) =>
-    typeof properties === 'object' && !Array.isArray(properties)
+    properties != null &&
+    typeof properties === 'object' &&
+    !Array.isArray(properties)
       ? properties
       : {};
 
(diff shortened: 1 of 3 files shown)


packages/storage/src/index.ts
```diff
@@ -64,7 +64,9 @@ function createStorage(): WxtStorage {
     value ?? fallback ?? null;
 
   const getMetaValue = (properties: any) =>
-    typeof properties === 'object' && !Array.isArray(properties)
+    properties != null &&
```
aklinker1: This a breaking change? We used to return `null` when meta was empty, but now we don't?

cookesan: Good catch. I restored the existing null handling in `getMetaValue` and now normalize `null` to `{}` only inside `watchMeta`, so the new watcher API does not change existing `getMeta` behavior. The storage tests and package checks pass.

PR #2450, opened 2026-06-28
feat: Add `prepare:tsconfig` hook for updating `.wxt/tsconfig.json` file

aklinker1:
### Overview

Provide a way to fully configure WXT's `.wxt/tsconfig.json` file via JS. See docs in changed files for example usage.

I also updated the base `excludes` to follow [TS's performance guide](https://github.com/microsoft/TypeScript/wiki/Performance#misconfigured-include-and-exclude).

### Manual Testing

See new E2E test.


packages/wxt/src/types.ts
```diff
@@ -1259,6 +1259,11 @@ export interface ServerInfo {
   origin: string;
 }
 
+export type PrepareTsconfigs = {
+  /** The JSON contents of the `.wxt/tsconfig.json` file. */
+  tsconfig: any;
```
PatrykKuniczak, line 1264: @aklinker1 There's no way to make a type?
I hope there's prebuild type for tsconfig, which you're able to use.
It can be better for intelisense hints.

PR #2425, opened 2026-06-18
docs: prioritize Bun in package manager options

ded-furby:
## Overview

This PR updates Bun to be the first package-manager option in WXT’s package-manager command tabs on the installation CLI `wxt init` package manager prompt.

- Updated CLI prompt order in `packages/wxt/src/core/initialize.ts` so Bun is listed first.

## Related Issue

Closes #2431

PR #2459, opened 2026-06-29
ci: Fix VHS action

aklinker1:
### Overview

Fixing action so it can generate new GIF after #2425


docs/assets/init-demo.gif
PatrykKuniczak, line 1: `Bun` is still on the very bottom:

Image

PatrykKuniczak (changes requested):

PatrykKuniczak (approved):
Good, you can place `pnpm` on the second place if you want to.
It's better option than `npm` 

We can have `bun` as preffered, `pnpm` as second good and `npm` as third, because only IMHO it's better than `yarn` i don't see almost anybody which use it on project.

docs/assets/init-demo.gif
aklinker1: Fixed, just had to merge with main and re-run the workflow

aklinker1:
NPM is still the go-to package manager for a lot of people, I don't want to change the order too much. Personally, I would have just moved bun up above yarn, making the order npm, pnpm, bun, yarn, but I don't have that strong of opinions so I didn't say anything on the other PR lol.

PR #2485, opened 2026-07-04
docs: use consola's fancy reporter in init demo GIF

### LINKED ISSUE #2461: Init demo GIF logging improvements
Noticed this in #2459, but consola is using it's basic formatter in the GIF... so it's logging `[log]` and `[info]` instead of no prefix and `i`, respectively.

https://github.com/wxt-dev/wxt/blob/e5ef936/docs/assets/init-demo.gif

This is probably because the GIF is generated by the github action, and consola sees it's a CI environment, so it doesn't use the pretty formatter, uses basic instead.

Not sure how to override this, but I'd prefer we use the pretty logger for the GIF

thribhuvan003:
## Problem

The `wxt init` demo GIF shows consola's basic reporter output (`[log]` / `[info]` prefixes) instead of the fancy reporter (no prefix / `ℹ`).

## Root cause

`initialize.ts` logs through consola's default instance, which picks its reporter at import time:

```
options.fancy ?? !(isCI || isTest) ? new FancyReporter() : new BasicReporter()
```

std-env reports `isCI: true` whenever `GITHUB_ACTIONS` is present with **any** value (CI provider scan), independent of `CI`. The vhs recording runs on the GitHub Actions runner, so the ttyd shell inherits `CI=true` and `GITHUB_ACTIONS=true` → BasicReporter.

## Fix

`unset CI GITHUB_ACTIONS` in the tape's hidden preparation block, before `bunx wxt init` runs. A shell-level `unset` (rather than vhs `Env`) reliably defeats both the `env.CI` check and the provider scan, and it never appears in the recording.

## Verification

Reproduced the reporter selection with consola 3.4.2 using the same calls `wxt init` makes:

| Environment | Output |
|---|---|
| `CI=true GITHUB_ACTIONS=true` | `[info] Initializing new project` (basic) |
| both unset | `ℹ Initializing new project` (fancy) |

Since `vhs.yml` triggers on tape changes (and supports `workflow_dispatch`), the GIF will regenerate on merge and can be re-checked there.

Fixes #2461

### DIFF
--- a/cspell.yml
+++ b/cspell.yml
@@ -29,6 +29,7 @@ words:
   - comctx
   - confbox
   - configfile
+  - consola
   - consolas
   - Cpath
   - crbug
--- a/docs/tapes/init-demo.tape
+++ b/docs/tapes/init-demo.tape
@@ -37,6 +37,9 @@ Hide
 # Create a temporary folder for demo
 Type 'mkdir packages/wxt/init-demo' Enter
 Type 'cd packages/wxt/init-demo' Enter
+# Unset CI variables inherited from the GitHub Actions runner so consola
+# picks its fancy reporter instead of the basic one (#2461)
+Type 'unset CI GITHUB_ACTIONS' Enter
 Type 'clear' Enter
 Show
 


PatrykKuniczak (changes requested):
On the preview there's still:
Image

PatrykKuniczak (approved):
WAIT WAIT
"the GIF will regenerate on merge" i see, maybe right, GH Actions need to run on target branch.

Let's try it, but you need to fix CSpell issue.

nishu-murmu:
@PatrykKuniczak how do we test this?
We have to merge it to check this right?

nishu-murmu (approved):

PatrykKuniczak:
@nishu-murmu Yeah, i hope.

This works only inside this `tapes`, we can safely check it on `main`, if it won't work, i'll revert it and @thribhuvan003 will be able to try again.

thribhuvan003:
added consola to the cspell words, should be green now. and yeah @nishu-murmu — the vhs workflow only runs when the tape changes on main, so the gif regenerates after merge. if it still looks off after that I'll follow up.

PatrykKuniczak:
And after last commit we need @aklinker1 Approval 🤣

PatrykKuniczak:
@nishu-murmu

aklinker1:
Ahh, yeah, can't run the action on fork manually :/. Let's just merge and see if it works.

aklinker1 (approved):

aklinker1:
Worked, but I'm not sure it looks any better :/

https://github.com/wxt-dev/wxt/commit/13be7c5bfdedc6660fc799ded6d868736dffcb86

https://github.com/wxt-dev/wxt/blob/13be7c5bfdedc6660fc799ded6d868736dffcb86/docs/assets/init-demo.gif

Now instead of `[log]` we see `[timestamp]` lol

PR #2506, opened 2026-07-26
chore(deps): Remove `esbuild` dependency

aklinker1:
### Overview

Was reviewing deps in #2447, noticed we don't need `esbuild` as a direct dependency anymore... I don't remember why it was added originally.


aklinker1:
Tests are failing on #2447, gonna merge if the same tests fail

PR #2510, opened 2026-07-26
fix(docs): use pretty logs in init demo GIF

### LINKED ISSUE #2461: Init demo GIF logging improvements

lntutor:
Fixes #2461.

The VHS recording subprocess now clears GitHub Actions CI markers before starting the demo, so Consola selects its compact pretty reporter only in the generated GIF. The asset was regenerated from that workflow command.

Validation: `bun run check`.


aklinker1:
Someone already opened a PR: https://github.com/wxt-dev/wxt/pull/2485

PR #2456, opened 2026-06-29
fix!: Remove `url:` import feature

nishu-murmu:
### Overview

Since there was the risk of supply chain attack it was better to remove this feature.

### Related Issue

This PR closes #2262

Issue #2262, opened 2026-04-13
Remove the `url:` import feature


aklinker1:
### Feature Request

Really, you should be using NPM packages whenever possible. There's also potential for supply chain attacks if the CDN get's taken over. I can't actually think of a real-world use case for this feature, I added it initially for feature parity with Plasma... But I don't think it's relevant anymore.

#### Is your feature request related to a bug?

See #2229

### What are the alternatives?

You can always just download the file yourself and host it in your repo. That way you can track if it changed and inspect for supply chain attacks 🤷 

### Additional context

N/A

PatrykKuniczak:
It also fix #2235

nishu-murmu:
Should we deprecate it or entirely remove it?

aklinker1:
Hmmmmmm..... we should probably deprecate it first.... :/

nishu-murmu:
I created a PR which entirely removes the feature. What should I do?
Should I update that PR to deprecate it?
@aklinker1

aklinker1:
Leave that PR as-is for now. I'm prepare the next major release, 0.28.0, we should include a deprecation in it, so let's open a separate PR adding a warning. Don't add a way to silence the warning.

After 0.28.0, we can merge #2456 into the next major branch, which will be for 0.29 or 1.0.

aklinker1:
I've decided to just drop it without a deprecation notice, it will be removed in v0.21

Issue #2235, opened 2026-04-05
Sometimes but pretty often jquery url isn't available


PatrykKuniczak:
### Describe the bug

<img width="1982" height="132" alt="Image" src="https://github.com/user-attachments/assets/456a9840-0afe-4e60-8524-5d07db5e9515" />

We can try to change this link for other remote and will see, if this'll fix issue

### Reproduction

1. Clone this project
2. Run `pnpm build` on `wxt-demo` and for time to time, there's a problem with that

### System Info

```shell
System:
    OS: Windows 11 10.0.26200
    CPU: (12) x64 AMD Ryzen 5 3600 6-Core Processor              
    Memory: 17.11 GB / 31.91 GB
  Binaries:
    Node: 25.6.1
    npm: 11.9.0
    pnpm: 10.29.2
    bun: 1.3.3
  Browsers:
    Chrome: 146.0.7680.178
    Edge: Chromium (140.0.3485.54)
    Firefox: 149.0
    Internet Explorer: 11.0.26100.7309
  npmPackages:
    wxt: workspace:* => 0.20.20
```

### Used Package Manager

pnpm

### Validations

- [x] Read the [Contributing Guidelines](https://github.com/wxt-dev/wxt/blob/main/CONTRIBUTING.md).
- [x] Read the [docs](https://wxt.dev/guide/installation.html).
- [x] Check that there isn't [already an issue](https://github.com/wxt-dev/wxt/issues) that reports the same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/wxt-dev/wxt/discussions) or join our [Discord Chat Server](https://discord.gg/ZFsZqGery9).
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

aklinker1:
I'd be down to just remove the `url:` import from the demo. I don't think it's a pattern most people should ever use.

Though I can't say I've every had this error. There's no way JQuery's own CDN goes down that often.

PatrykKuniczak:
closed by #2262

PR #2514, opened 2026-07-26
feat!: Restore url: imports, require an integrity hash

### LINKED ISSUE #1924: Remote code checksum
### Feature Request

A remote code checksum could be interesting. Something like...

```ts
import 'url#CE114E4501D2F4E2DCEA3E17B546F339:https://www.googletagmanager.com/gtag/js';
```

Why? Imagine if wxt added a command, `wxt remote-code-outdated`. You could run it on CI every night. If it fails, you'd know that you need to release a new version of the browser extension.

N/A

Implementing a custom process, but it'd be more code.

### Additional context

When remote assets go out of date you might not know, and things might break!

aeharding:
### Overview

#2456 removed `url:` imports over supply chain risk (#2229, #2262). The suggested alternative was to download the file and commit it to your repo. That works, but it has three problems:

1. **It goes stale silently.** Nothing tells you the upstream file changed after you vendored it.
2. **It isn't self-documenting.** A year later nobody knows where `vendor/gtag.min.js` came from. You can write it down, but docs drift.
3. **It's PR noise.** 70KB of minified JS in the diff, and again on every update.

This PR restores `url:` imports with a required integrity hash. That keeps the thing that made vendoring safe, knowing exactly which bytes ship, without any of the three problems:

```ts
import 'url#sha256-kmHvs0B+OpCW5GVHUNjv9rOmY0IvSIRcf7zGUDTDQM8=:https://code.jquery.com/jquery-3.7.1.slim.min.js';
```

sha256/384/512, in base64 (the SRI format CDNs already publish) or hex. **Every URL import must pin a hash, with no opt-out**, so the hole that motivated removal stays closed. A bare `url:` import is now a build error that prints the exact line to paste, so migrating is copy/paste. Content that fails verification is never bundled or cached, and the cache is verified on read, so a poisoned `.wxt/cache` can't slip through either.

That also covers #1924 without a new command: build on a schedule in CI and you find out when a remote dependency drifts.

### Manual Testing

`packages/wxt-demo` pins a jQuery URL, so `bun run build` exercises it. Then edit the import in `src/entrypoints/options/main.ts`: dropping the hash, corrupting the hash, or using `md5-` each fails with a specific error, and swapping base64 for hex still builds.

Unit tests cover parsing, hashing, verification, and the cache behavior. `e2e/tests/remote-code.test.ts` covers valid, hex, missing, and mismatched hashes.

### Related Issue

This PR closes #1924.

Also relevant: #2262, #2229, and #2456 (the removal this reverts). Targets `major`, since that's where #2456 landed and this is breaking.

### DIFF
--- a/packages/wxt-demo/src/entrypoints/options/main.ts
+++ b/packages/wxt-demo/src/entrypoints/options/main.ts
@@ -1,3 +1,6 @@
+// @ts-expect-error: URL imports not typed
+import 'url#sha256-kmHvs0B+OpCW5GVHUNjv9rOmY0IvSIRcf7zGUDTDQM8=:https://code.jquery.com/jquery-3.7.1.slim.min.js';
+
 console.log(browser.runtime.id);
 logId();
 console.log(2);
--- a/README.md
+++ b/README.md
@@ -53,6 +53,7 @@ Or see the [installation guide](https://wxt.dev/guide/installation.html) to get 
 - 📦 [Module system](https://wxt.dev/guide/essentials/wxt-modules.html#overview) for reusing code between extensions
 - 🖍️ Quickly bootstrap a new project
 - 📏 Bundle analysis
+- ⬇️ Download and bundle remote URL imports
 
 ## Sponsors
 
--- a/docs/guide/essentials/remote-code.md
+++ b/docs/guide/essentials/remote-code.md
@@ -1,0 +1,72 @@
+# Remote Code
+
+WXT will automatically download and bundle imports with the `url:` prefix so the extension does not depend on remote code, [a requirement from Google for MV3](https://developer.chrome.com/docs/extensions/migrating/improve-security/#remove-remote-code).
+
+## Integrity Hashes
+
+Every URL import must pin an integrity hash. Whatever you import is bundled into your extension and ships to your users, so an unpinned import means a compromised or hijacked CDN can put arbitrary code in your next release without you noticing.
+
+The hash goes between `url` and the URL, prefixed with `#`:
+
+```ts
+import 'url#sha256-kmHvs0B+OpCW5GVHUNjv9rOmY0IvSIRcf7zGUDTDQM8=:https://code.jquery.com/jquery-3.7.1.slim.min.js';
+```
+
+`sha256`, `sha384`, and `sha512` are supported, in either base64 (the same format as [Subresource Integrity](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity), which most CDNs publish) or hex.
+
+If you don't know the hash yet, import the URL without one. The build will fail and print the hash of the file currently being served:
+
+```
+URL imports must pin an integrity hash: "url:https://code.jquery.com/jquery-3.7.1.slim.min.js"
+
+  import 'url#sha256-kmHvs0B+OpCW5GVHUNjv9rOmY0IvSIRcf7zGUDTDQM8=:https://code.jquery.com/jquery-3.7.1.slim.min.js';
+```
+
+:::warning
+A hash locks in whatever is served today — it does not make untrusted code safe. Review the file before pinning it.
+:::
+
+When the remote file changes, the build fails instead of silently bundling the new version:
+
+```
+Integrity check failed for "https://code.jquery.com/jquery-3.7.1.slim.min.js".
+
+  Expected: sha256-kmHvs0B+OpCW5GVHUNjv9rOmY0IvSIRcf7zGUDTDQM8=
+  Received: sha256-b6WuS8lqRB1JlP0z1BjCXGKtIFgKPKcJPvXCLpxLxeM=
+
+The remote file changed since you added it. Review the new file, and if you trust it, update the hash in your import.
+```
+
+Run your build on a schedule in CI to find out when a remote dependency changes, rather than at release time.
+
+## Google Analytics
+
+For example, you can import Google Analytics:
+
+```ts
+// utils/google-analytics.ts
+import 'url#sha256-XXXXXX:https://www.googletagmanager.com/gtag/js?id=G-XXXXXX';
+
+window.dataLayer = window.dataLayer || [];
+// NOTE: This line is different from Google's documentation
+window.gtag = function () {
+  dataLayer.push(arguments);
+};
+gtag('js', new Date());
+gtag('config', 'G-XXXXXX');
+```
+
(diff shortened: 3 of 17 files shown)


packages/wxt-demo/src/entrypoints/options/main.ts
```diff
@@ -1,3 +1,6 @@
+// @ts-expect-error: URL imports not typed
+import 'url#sha256-kmHvs0B+OpCW5GVHUNjv9rOmY0IvSIRcf7zGUDTDQM8=:https://code.jquery.com/jquery-3.7.1.slim.min.js';
```
PatrykKuniczak, line 2: https://github.com/wxt-dev/wxt/issues/2235
There's a problem for time to time, are you know about it, and it's fixed in this PR?

PatrykKuniczak (changes requested):

packages/wxt-demo/src/entrypoints/options/main.ts
aeharding: I think I may have found the problem. The `fetch` could throw an error (like on network issues, ECONNRESET whatever) which would bypass the cache. Please see changes here: https://github.com/wxt-dev/wxt/pull/2514/changes#diff-7f5f341e02d05df8813c881b169aa49234938ff2877a50b7e02c0778d42258c7R79-R96

aklinker1:
Another reason this was removed - do you have an example of a real dependency that needs to downloaded via a CDN or URL? 

- Your PR mentions `jquery`, which has an NPM package
- google analytics, which I used in docs and examples of this feature, has never worked in MV3 extensions in any context (background, content script, ui), even when using this feature. I've never actually tested it in MV2 extensions either, but I would assume it has the same issues as MV3 - that the backend just ignores events from extension sources.

Also, instead of adding this back to the core package, what about making this a WXT module that people can install if they need this functionality (I'm still not convinced anyone does)

This is what we did when we removed the `webextension-polyfill` in the last major release - people could continue using it by installing the `@wxt-dev/webextension-polyfill` module.

aklinker1:
I will wait for this to be resolved before doing the major release.

aeharding:
> Another reason this was removed - do you have an example of a real dependency that needs to downloaded via a CDN or URL?

Yes - for us its a feedback assistant chat tool, called Maven (https://maven.com) and they only distribute the js via CDN.

> Also, instead of adding this back to the core package, what about making this a WXT module that people can install if they need this functionality (I'm still not convinced anyone does)

That's an option, I'll defer to you on it. :) FWIW, we definitely need this functionality

aklinker1:
Thanks for sharing!

Some other thoughts:

>**It goes stale silently**. Nothing tells you the upstream file changed after you vendored it.

This is true, you would want to add tooling to check for updates. A simple script that downloads and checks if there's a git diff would do it - that's a few lines of shell scripting.

> **It isn't self-documenting**. A year later nobody knows where vendor/gtag.min.js came from. You can write it down, but docs drift.

Yes, you should have a `vendor.sh` file that you can run to downloads the latest versions of your vendored dependencies.

> **It's PR noise**. 70KB of minified JS in the diff, and again on every update.

Yes, it would be noisy, but that's normal for managing vendored dependencies. 

> This PR restores url: imports with a required integrity hash. That keeps the thing that made vendoring safe, knowing exactly which bytes ship, without any of the three problems:
>
> ```ts
> import 'url#sha256-kmHvs0B+OpCW5GVHUNjv9rOmY0IvSIRcf7zGUDTDQM8=:https://code.jquery.com/jquery-3.7.1.slim.min.js';
> ```

So this works fine (most of the time) for URLs with a version identifier. But what about URLs without a version? Or what if an attacker overwrites a version, same URL, but different content? Or for your example, what happens if maven goes out of business and stops hosting the file?

In both cases, **it would be impossible to build your extension** from a fresh computer unless you've committed the library to your repo. Because the SHA would have changed and you might never be able to download the old file again - causing your build to fail when it downloads something with a different SHA.

But if you download it, commit it to the repo, it would still build, you'd have a runtime error instead. I think that's much better than depending on something that no longer exists. You have no way to verify the old behavior without the old file.

There is no way to "lock" down a version to a specific URL and SHA without downloading the file and committing it.

aeharding:
> it would be impossible to build your extension from a fresh computer unless you've committed the library to your repo

Thats true but its true of any dependency, including npmjs.org.

> Or what if an attacker overwrites a version, same URL, but different content?

Well thats what the integrity is for - build should fail in that case and tell you.

> But if you download it, commit it to the repo, it would still build, you'd have a runtime error instead. I think that's much better than depending on something that no longer exists. You have no way to verify the old behavior without the old file.
>
> There is no way to "lock" down a version to a specific URL and SHA without downloading the file and committing it.

I see your point, but in our case before we used WXT thats what we did, and the dependency would stop working randomly because we wouldn't be notified that the file changed, and the JS would stop working (couldn't communicate with their API, too old).

---

All that being said though, I appreciate all the time you dedicate to this  project, I'm happy to make this into a plugin if you'd like to keep it out of core, should be possible!

aklinker1:
> > it would be impossible to build your extension from a fresh computer unless you've committed the library to your repo
>
> Thats true but its true of any dependency, including npmjs.org.

Yeah, very true, very true.

---

I see where you're coming from, but let me be stubborn one more time... This is a 700 line PR that I'll have to maintain, and it seems like your use case could be implemented in less than 20 lines:

```sh
# vendor.sh
curl -o vendor/dep.js https://some.website.com/some-path
```

```sh
# vendor-check.sh
./vendor.sh
if [ -n "$(git status --porcelain)" ]; then
  echo "Versioned dependencies are out of date" >&2
  exit 1
fi
```

```yaml
# .github/workflows/vendor-check.yml
on:
  cron: @daily

jobs:
  check:
    runs-on: ubuntu-latest
    steps:
      - uses: action/checkout
      - run: ./vendor-check.sh
      - if: failure()
        run: echo "Message slack, open PR, etc"
```

At the cost of some ugly diffs. No need to verify hashes if you commit the file to the repo.

Even if we decide we don't want to merge this, either apart of the core `wxt` package or a new one - you can publish this code in your own NPM package as a WXT module. Then it's maintained by someone who actually uses it. I'd by happy to direct people towards your package if people need this feature.

aeharding:
Yeah I think I see what you're saying! And if its just me using this functionality then I agree, best to not maintain it here. I'll pull it into a separate plugin, or some one-off scripting for the use case I need.

Thanks!

PR #2534, opened 2026-07-30
fix(vite-builder): tolerate `ENOENT` when deleting empty directories

### LINKED ISSUE #2533: Race in removeEmptyDirs() causes intermittent ENOENT during build when a plugin writes into outDir concurrently
### Describe the bug

`wxt build` intermittently fails with `ENOENT: no such file or directory, stat '.../inject-scripts/<file>.js'` (the exact file varies between runs — I've seen both `recorder.js` and `fill-helper.js`) on a fresh, clean `.output/` directory. It is not consistent — most builds succeed, but it recurs across otherwise-identical repeated runs (roughly 1 in 4-6 in my testing).

### Root cause (traced in `wxt@0.20.13`)

`moveHtmlFiles()` in `packages/wxt/src/core/builders/vite/index.ts` calls `removeEmptyDirs(config.outDir)` after each entrypoint group finishes building:

```js
async function moveHtmlFiles(config, group, chunks) {
  ...
  removeEmptyDirs(config.outDir);
  return movedChunks;
}
export async function removeEmptyDirs(dir) {
  const files = await fs.readdir(dir);
  for (const file of files) {
    const filePath = join(dir, file);
    const stats = await fs.stat(filePath);   // <-- can throw ENOENT
    if (stats.isDirectory()) {
      await removeEmptyDirs(filePath);
    }
  }
  try {
    await fs.rmdir(dir);
  } catch { ... }
}
```

This does an unsynchronized `readdir()` → `stat()` walk of the **entire shared `outDir`**, called once per entrypoint group (entrypoint groups themselves build sequentially, so this alone isn't the race). The race is with any **plugin hooked at the same lifecycle point that also writes into `outDir`** — in my case, `vite-plugin-static-copy` configured with `hook: 'writeBundle'` to copy static assets (`inject-scripts/*.js`, `workers/*`, `_locales/**/*`) into the output directory:

```js
viteStaticCopy({
  targets: [
    { src: 'inject-scripts/*.js', dest: 'inject-scripts' },
    { src: ['workers/*'], dest: 'workers' },
    { src: '_locales/**/*', dest: '_locales' },
  ],
  hook: 'writeBundle',
})
```

Both `writeBundle` (the copy plugin) and wxt's own post-build step (`removeEmptyDirs`) run around the same point in the build lifecycle without any ordering guarantee between them. When `removeEmptyDirs`'s `readdir()` observes a directory (e.g. `inject-scripts/`) that the copy plugin is concurrently populating, and a file inside it gets renamed/moved/removed by the copy operation between the `readdir()` and the subsequent `fs.stat()` call on that specific entry, `fs.stat()` throws `ENOENT` and the whole build fails.

I confirmed this is not:
- cross-entrypoint-group parallelism — `build-entrypoints.mjs` builds groups with a plain sequential `for` loop with `await`, so only one group builds at a time.
- a straightforward "static-copy runs after wxt" ordering bug — switching `hook: 'writeBundle'` → `hook: 'closeBundle'` (a later hook) did **not** eliminate the race; it still reproduced.

Given `removeEmptyDirs` performs raw `readdir`+`stat` (not atomic) over a directory tree that other write-side plugins are permitted to mutate concurrently, this looks like a genuine TOCTOU race in wxt's own build-output cleanup step, not something fixable purely from the consuming project's Vite config.

### Reproduction

Unfortunately I don't have a minimal standalone repro repo yet — this was found investigating a flaky CI/build failure in a larger project ([mcp-chrome](https://github.com/hangwin/mcp-chrome), our fork tracked internally as `oleks/mcp-chrome#61`). The failure rate is low enough (roughly 1 in 4-6 clean builds) that a minimal repro would need a project with:
- multiple entrypoint groups (so at least one HTML/background/content-script build triggers `moveHtmlFiles` → `removeEmptyDirs`)
- a `vite-plugin-static-copy` (or any other plugin) writing into a subdirectory of `outDir` around the same build-hook lifecycle point

I'm happy to help narrow this down further or test a patch if it'd help — wanted to file the root-cause analysis now while it's fresh rather than sit on it.

### System Info

```
wxt: 0.20.13
node: v24.16.0
OS: Linux (NixOS)
```

### Validations

- [x] Read the [docs](https://wxt.dev).
- [x] Checked if an issue already exists for the bug.

retif:
### What

`removeEmptyDirs()` (in `packages/wxt/src/core/builders/vite/index.ts`) walks `outDir` with `readdir()` then `stat()`s each entry to find directories to recurse into. If another plugin writes into `outDir` around the same build-hook lifecycle point — in my case `vite-plugin-static-copy` on `hook: 'writeBundle'` — a file can be renamed or removed between the `readdir()` and the `stat()` call on that specific entry, throwing `ENOENT` and failing the entire build intermittently.

This fixes that by catching `ENOENT` from `stat()` and skipping that entry (it can't be a directory we need to recurse into if it no longer exists), rather than letting the error propagate.

### Context

Full root-cause writeup, including how I traced this and ruled out a couple of other candidate causes (cross-entrypoint-group parallelism, a `writeBundle` → `closeBundle` hook-ordering fix): #2533

### Testing

Added `packages/wxt/src/core/builders/vite/__tests__/remove-empty-dirs.test.ts` with two cases:
- baseline: nested empty directories still get removed
- the regression case: a file disappearing between `readdir()` and `stat()` no longer throws

I wasn't able to run the full local build/test suite for this change (constrained dev environment), so please let CI be the actual verification here — happy to iterate on review feedback.

### DIFF
--- a/packages/wxt/src/core/builders/vite/index.ts
+++ b/packages/wxt/src/core/builders/vite/index.ts
@@ -1,491 +1,491 @@
-import { Hookable } from "hookable";
-import { mkdir, readdir, rename, rmdir, stat } from "node:fs/promises";
-import { dirname, extname, join, relative, resolve } from "node:path";
-import type * as vite from "vite";
+import { Hookable } from 'hookable';
+import { mkdir, readdir, rename, rmdir, stat } from 'node:fs/promises';
+import { dirname, extname, join, relative, resolve } from 'node:path';
+import type * as vite from 'vite';
 import {
-	BuildStepOutput,
-	Entrypoint,
-	EntrypointGroup,
-	ResolvedConfig,
-	WxtBuilder,
-	WxtBuilderServer,
-	WxtDevServer,
-	WxtHooks,
-} from "../../../types";
-import { normalizePath } from "../../utils";
-import { toArray } from "../../utils/arrays";
+  BuildStepOutput,
+  Entrypoint,
+  EntrypointGroup,
+  ResolvedConfig,
+  WxtBuilder,
+  WxtBuilderServer,
+  WxtDevServer,
+  WxtHooks,
+} from '../../../types';
+import { normalizePath } from '../../utils';
+import { toArray } from '../../utils/arrays';
 import {
-	getEntrypointBundlePath,
-	isHtmlEntrypoint,
-} from "../../utils/entrypoints";
-import { createExtensionEnvironment } from "../../utils/environments";
-import { safeVarName } from "../../utils/strings";
+  getEntrypointBundlePath,
+  isHtmlEntrypoint,
+} from '../../utils/entrypoints';
+import { createExtensionEnvironment } from '../../utils/environments';
+import { safeVarName } from '../../utils/strings';
 import {
-	VirtualEntrypointType,
-	VirtualModuleId,
-} from "../../utils/virtual-modules";
-import * as wxtPlugins from "./plugins";
+  VirtualEntrypointType,
+  VirtualModuleId,
+} from '../../utils/virtual-modules';
+import * as wxtPlugins from './plugins';
 
 interface RollupAssetNameInfo {
-	name?: string;
-	names?: string[];
+  name?: string;
+  names?: string[];
 }
 
 export async function createViteBuilder(
-	wxtConfig: ResolvedConfig,
-	hooks: Hookable<WxtHooks>,
-	getWxtDevServer?: () => WxtDevServer | undefined,
+  wxtConfig: ResolvedConfig,
+  hooks: Hookable<WxtHooks>,
+  getWxtDevServer?: () => WxtDevServer | undefined,
 ): Promise<WxtBuilder> {
-	const vite = await import("vite");
+  const vite = await import('vite');
 
-	/**
-	 * Returns the base vite config shared by all builds based on the inline and
-	 * user config.
-	 */
-	const getBaseConfig = async (baseConfigOptions?: {
-		excludeAnalysisPlugin?: boolean;
-	}) => {
-		const config: vite.InlineConfig = await wxtConfig.vite(wxtConfig.env);
+  /**
+   * Returns the base vite config shared by all builds based on the inline and
+   * user config.
+   */
+  const getBaseConfig = async (baseConfigOptions?: {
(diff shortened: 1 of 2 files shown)


packages/wxt/src/core/builders/vite/index.ts
```diff
@@ -528,65 +528,76 @@ function getRollupEntry(entrypoint: Entrypoint): string {
...
+		const filePath = join(dir, file);
+		let stats;
+		try {
+			stats = await stat(filePath);
+		} catch (err: any) {
+			// Another plugin (eg. vite-plugin-static-copy) can be writing into
+			// outDir concurrently with this cleanup pass. A file that existed at
+			// readdir() time can be renamed/removed before we stat() it - that's
+			// not a directory we need to recurse into, so skip it instead of
+			// failing the whole build.
```
aklinker1:
```suggestion
			// Ignore files that were moved between readdir and stat due to race
			// conditions. For more details, see:
			// https://github.com/wxt-dev/wxt/issues/2533
```

aklinker1 (approved):
New tests look good, we can point people to your writeup instead of having a large comment.

Checks are probably going to fail because of formatting, run `bun check` in the project root and it will be fixed.

_Edit: yup, and some typos_

```
7 Problems: 
⚠ Not formatted.
  → ./packages/wxt/src/core/builders/vite/__tests__/remove-empty-dirs.test.ts
✗ Unknown word (oleks)
  → ./packages/wxt/src/core/builders/vite/__tests__/remove-empty-dirs.test.ts:9:32
✗ Unknown word (oleks)
  → ./packages/wxt/src/core/builders/vite/__tests__/remove-empty-dirs.test.ts:46:97
✗ Unknown word (racey)
  → ./packages/wxt/src/core/builders/vite/__tests__/remove-empty-dirs.test.ts:50:10
✗ Unknown word (racey)
  → ./packages/wxt/src/core/builders/vite/__tests__/remove-empty-dirs.test.ts:51:20
✗ Unknown word (racey)
  → ./packages/wxt/src/core/builders/vite/__tests__/remove-empty-dirs.test.ts:57:18
⚠ Not formatted.
  → ./packages/wxt/src/core/builders/vite/index.ts
```

PR #2536, opened 2026-07-30
feat: Log warnings once

### LINKED ISSUE #2507: Introduce `Logger#warnOnce`
### Feature Request

We have lots of warning logs for experimental features that run when config is resolved. They get logged multiple times, like on dev server restarts, unecessarily.

Here's an example: https://github.com/wxt-dev/wxt/blob/5bd36e90bfdcecfed5ad8a2bd6bac2fa42ae6846/packages/wxt/src/core/utils/manifest.ts#L79-L81

Not a bug, discovered while working on f2f1577b514049324000f48eaa4c4886d4cce84f

Continue printing the same warning over and over.

### Additional context

To prevent a breaking change, this function should be optional. When missing, callers should fallback on `warn`. Alternatively, `warnOnce` could be added inside WXT to all loggers, and the internal type for `wxt.logger` is updated.

fkatsuhiro:
### Overview

- Wrapped `wxt.logger` with `createWxtLogger` to add `warnOnce` (the Logger type itself stays unchanged)
- The dedup `Set` is created once in `registerWxt` and delegates through a closure that always reads the current `wxt.config.logger`
- State persists across `reloadConfig()`, so `manifest.ts` warnings no longer repeat on dev-server restarts

#### `warnOnce`
- Calling it repeatedly with the same arguments logs to the underlying `warn` only once, on the first call, with those exact arguments
- Calling it with distinct arguments logs each one independently, once per distinct message
- Regular warn calls are unaffected by dedup — they always pass through to the underlying warn every time
- If the underlying logger instance is swapped out mid-life (simulating wxt.reloadConfig() replacing wxt.config.logger), the same message still isn't re-logged, while a genuinely new message is correctly delivered to the new underlying logger
#### `delegation`
- `debug`/ `log` / `info` / `error` / `fatal` / `success` are all passed straight through to the underlying logger
- Reading and writing `level` is reflected on the underlying logger

image

### Related Issue
This PR closes #2507

### DIFF
--- a/packages/wxt/src/core/wxt.ts
+++ b/packages/wxt/src/core/wxt.ts
@@ -5,7 +5,6 @@ import { InlineConfig, Wxt, WxtCommand, WxtHooks, WxtModule } from '../types';
 import { createViteBuilder } from './builders/vite';
 import { createWxtPackageManager } from './package-managers';
 import { resolveConfig } from './resolve-config';
-import { createWxtLogger } from './utils/log/wxtLogger';
 
 /**
  * Global variable set once `createWxt` is called once. Since this variable is
@@ -34,15 +33,13 @@ export async function registerWxt(
     createWxtPackageManager(config.root),
   ]);
 
-  // Created once so `warnOnce`. Reads `wxt.config` so it always delegates to
-  // the latest resolved logger, even after the config is reloaded.
-  const logger = createWxtLogger(() => wxt.config.logger);
-
   wxt = {
     config,
     hooks,
     hook: hooks.hook.bind(hooks),
-    logger,
+    get logger() {
+      return config.logger;
+    },
     async reloadConfig() {
       // Prevent changing the server port when resolving config multiple times
       // get-port-please doesn't always return the same port if it was recently closed.
--- a/packages/wxt/src/core/wxt.ts
+++ b/packages/wxt/src/core/wxt.ts
@@ -5,6 +5,7 @@ import { InlineConfig, Wxt, WxtCommand, WxtHooks, WxtModule } from '../types';
 import { createViteBuilder } from './builders/vite';
 import { createWxtPackageManager } from './package-managers';
 import { resolveConfig } from './resolve-config';
+import { createWxtLogger } from './utils/log/wxtLogger';
 
 /**
  * Global variable set once `createWxt` is called once. Since this variable is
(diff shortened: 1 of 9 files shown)


packages/wxt/src/core/wxt.ts
```diff
@@ -33,13 +34,15 @@ export async function registerWxt(
     createWxtPackageManager(config.root),
   ]);
 
+  // Created once so `warnOnce`. Reads `wxt.config` so it always delegates to
+  // the latest resolved logger, even after the config is reloaded.
+  const logger = createWxtLogger(() => wxt.config.logger);
```
aklinker1: Rather than using a callback, we need to move this logic into `resolve-config.ts`, and `createWxtLogger` should accept a `Logger` as the argument instead of a `() => Logger`.

This line: https://github.com/wxt-dev/wxt/blob/c9520b688d49bd4296d89ff5dcc6cd7fc2a8c5f1/packages/wxt/src/core/resolve-config.ts#L70

should be

```ts
const logger = createWxtLogger(mergedConfig.logger ?? consola)
```

That way we only call the function once per config change instead of on every access of `wxt.logger`

Then we can also only warn once for config, by passing in a `WxtLogger` here:

https://github.com/wxt-dev/wxt/blob/c9520b688d49bd4296d89ff5dcc6cd7fc2a8c5f1/packages/wxt/src/core/resolve-config.ts#L504-L524

packages/wxt/src/core/wxt.ts
```diff
@@ -33,13 +34,15 @@ export async function registerWxt(
...
+
   wxt = {
     config,
     hooks,
     hook: hooks.hook.bind(hooks),
-    get logger() {
-      return config.logger;
-    },
+    logger,
```
aklinker1:
```suggestion
    get logger() {
      return config.logger;
    },
```

Revert this back after moving the logic to `resolve-config.ts`

aklinker1 (changes requested):
Looks good other than the callback setup you're using. It adds a bit of overhead every time `wxt.logger` is accessed - see comments for recommended changes

fkatsuhiro:
> Looks good other than the callback setup you're using. It adds a bit of overhead every time `wxt.logger` is accessed - see comments for recommended changes

Thank you for review this PR. Let me update it. Thank you.

aklinker1 (approved):
I cleaned up the tests a little - mocked dependencies for the `resolve-config.test.ts` (e2e tests setup directories, not unit), and used `mock` instead of fully defining a fake logger instance. I also changed the logic to share the warn set between all instances of `WxtLogger`, not per-instance.

Issue #2523, opened 2026-07-28
Drop, Replace, Inline, DIY, or Minimize Dependencies


aklinker1:
### Feature Request

I really want to decrease the size of WXT. As of 0.21.2, here's it's install size (excluding peers):

```
22MB / 156 packages
```

> - https://pkg-size.dev/wxt@%5E0.21?no-peers
> - https://npmgraph.js.org/?q=wxt@^0.21
> - https://node-modules.dev/grid/depth#install=wxt@^0.21

There's a couple options for cleaning these up:

1. Is the dependency actually needed anymore? No? Drop it
2. Can the dependency be replaced with a smaller alternative? Yes? Do it
3. Is the dependency small and not likely to have a vulnerability? Yes? Inline it
4. Is the dependency something that can be replaced by a single function? Yes? DIY it
5. After reviewing, could the package be cleaned up? Yes? Contribute to it

### Dependencies

This is a list of all of WXT's 37 direct dependencies, and what, if anything, we can do to reduce it's size:

#### High Priority

- [x] `import-meta-resolve` (79KB / 1 package, ESM) - Drop, not needed in Node 22
   - #2526 
- [x] `publish-browser-extension` (933KB / 7 packages, ESM)
   - Released v6 &rarr; [333KB / 4 packages](https://pkg-size.dev/publish-browser-extension@6)
   - Explored inlining final deps, but decided not to: https://github.com/aklinker1/publish-browser-extension/pull/87
- [ ] `@aklinker1/rollup-plugin-visualizer` (1.7MB / 23 packages, CJS) - Replace with <https://sonda.dev> or DIY it (how hard can it be? lol). What does nuxt's analysis tool and devtools use? OR trim down my fork to remove `yargs` and `open`
   - Issue for sonda already exists: https://github.com/filipsobol/sonda/issues/195
   - Rolldown provide's it's own bundle analysis tool, though it is experimental and this would lock us into Vite even more. We would just have to render a UI: https://rolldown.rs/builtin-plugins/bundle-analyzer
   - I got an extension working with vite's environment API, https://github.com/wxt-dev/wxt/issues/1046#issuecomment-5109933695, but `rollup-plugin-visualizer` didn't work when building in series, only in parallel, meaning it's getting lucky with a race condition. So my fork is still necessary

#### Medium Priority

- [x] `open` (83KB / 12 packages, ESM) - Replace with `tiny-open`?
   - #2527
[...]

PR #2538, opened 2026-07-30
chore(deps): Replace `jszip` with `@aklinker1/zero-zip`

aklinker1:
### Overview

Apart of #2523. Reduces WXT's install footprint: [20MB / 139 packages](https://pkg-size.dev/https:%2F%2Fpkg.pr.new%2Fwxt@d9653d7df643ac52629763fb267b8b2d54c0bb4b?no-peers) &rarr; [18MB / 127 packages](https://pkg-size.dev/https:%2F%2Fpkg.pr.new%2Fwxt@d6cdc45?no-peers)

That's a 10% reduction in bytes, 8.6% reduction in dependencies.

### Manual Testing

- [x] Does chrome accept these zips?
   - Yes, on Linux at least
- [x] Does it work with a compression level of zero?
   - Yes, demo project is `269.24 kB` in build output, zip is `273.1 kB`.
- [x] Are the zip sizes comparable?
   - `97.65 kB` on `main`, `97.1 kB` on this PR
- [x] What about time it takes to build the zip?
   - `53 ms` on `main`, `27 ms` on this PR
- [x] Memory used? `/usr/bin/time -v bun zip`
   - `415912` on `main`, `407784` on this PR


aklinker1:
Interestingly, we can't inline dependencies when using `--unbundle` mode... that's why checks are failing

PR #2540, opened 2026-07-31
chore: Use Vite 8 for WXT development

aklinker1:
### Overview

Was using vite 7 before. Pulled out of #2537

### DIFF
--- a/packages/wxt/e2e/tests/encoding.test.ts
+++ b/packages/wxt/e2e/tests/encoding.test.ts
@@ -3,7 +3,7 @@ import { describe, expect, it } from 'vitest';
 import { TestProject } from '../utils';
 
 describe('Encoding', () => {
-  it('should convert unicode characters to ascii escaped chars', async () => {
+  it.skip('should convert unicode characters to ascii escaped chars', async () => {
     // See more details about this test, see:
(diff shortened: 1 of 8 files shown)


packages/wxt/e2e/tests/encoding.test.ts
```diff
@@ -3,7 +3,7 @@ import { describe, expect, it } from 'vitest';
 import { TestProject } from '../utils';
 
 describe('Encoding', () => {
-  it('should convert unicode characters to ascii escaped chars', async () => {
+  it.skip('should convert unicode characters to ascii escaped chars', async () => {
```
aklinker1, line 6: This will be unskipped in #2537

packages/wxt/e2e/tests/output-structure.test.ts
```diff
@@ -99,15 +98,7 @@ describe('Output Directory Structure', () => {
       `body { color: purple }`,
     );
 
-    await project.build({
-      vite: () => ({
-        build: {
-          rollupOptions: {
-            strictDeprecations: true,
```
aklinker1, line 106: This is not available in rolldown, didn't seem to affect the test when removed.

PR #2552, opened 2026-07-31
chore: Upgrade to TS 7 for WXT development

aklinker1:
### Overview

`bun --cwd packages/wxt check` runs in:

- Old: 3.432s
- New: 0.424s (8.09x)

Pretty nice!


aklinker1:
Looks like we need to upgrade typedoc before we merge this...

PR #2554, opened 2026-08-01
chore(deps): Replace `normalize-path` with custom `normalize` function

suveshmoza:
### Overview

Replace `normalize-path`(9.2KB) with a local `normalize` function. 

### Manual Testing
All existing path tests pass:
`bun run --filter wxt test -- run src/core/utils/__tests__/paths.test.ts`

### Related Issue

#2523


aklinker1:
I think I would rather inline this package than re-implement it. I looked at it's implementation and it's a bit smaller than yours: https://npmx.dev/package-code/normalize-path/v/3.0.0/index.js

I don't think unbundle mode supports inlining, `tsdown` tries to point it to the path in `node_modules` when the package was built - but dependencies are stored in different places for different package managers, so it doesn't work.

https://github.com/wxt-dev/wxt/actions/runs/30576077054/job/90984572240

Could you look into a way to either download or inline the package instead?

aklinker1:
Actually, I just thought of something... maybe it's as simple as writing a rolldown plugin that looks at the `alwaysBundle` option and overrides how those modules are resolved - ~~leaving them as importing the package by name instead of using a relative path to the location in `node_modules`...~~ bundling them to the output and resolving to the output file

Not sure if tsdown will respect the rolldown hooks and plugins system though in unbundle mode...

suveshmoza:
~~Opened a test PR #2556 for the Rolldown plugin. It remaps the `alwaysBundle` deps to `dist/_vendor`. This was new to me, so I used some help from an LLM for the plugin part.~~

Closing #2556 , saw your implementation and it’s much simpler.

aklinker1:
Mine doesn't work quite yet... I'll check yours out to.

aklinker1:
OK, got it working. I'll add you as a contributor, we'll use my PR: https://github.com/wxt-dev/wxt/pull/2557

PR #2557, opened 2026-08-01
chore(deps): Inline `normalize-path`

aklinker1:
### Overview

Prebuild inline deps as minified ESM, then change where the WXT build resolves these depedencies from - from `node_modules` to the new `inline` directory.

Goal here is to strip unused CJS, docs, maps, and other files that are not necessary for WXT to function. Generally dependencies should only be inlined if they are small themselves, and this will just prevent package managers from having to resolve and download these packages.

A couple of potential issues with this:

- **CJS/ESM interop** - does rolldown output a ESM file in the same export format as we consume in the source code? Will consumers be able to import these files? Seems like it if the template tests are passing
- **Tree-shaking** - currently, we're pre-bundling the entire inline dep - but what if we only use one of the exports? Does the WXT build tree-shake out unused code automatically? If not, our install footprint will still be smaller, but not as small as it could be.
- **Inlining dependencies with dependencies** - this should work just fine.

Results from inlining `normalize-path`: [18MB / 125 packages](https://pkg-size.dev/https:%2F%2Fpkg.pr.new%2Fwxt@1153189082431a67f8f14c0ce5fdd48391047125?no-peers) &rarr; [18MB / 124 packages](https://pkg-size.dev/https:%2F%2Fpkg.pr.new%2Fwxt@0a408ed?no-peers)

WXT itself goes from `335KB` &rarr; `336KB` (+0.79 kB vs -9.2 kB from removing the package)

### Manual Testing

If the template jobs pass CI, that means the setup is working, and projects using WXT can import things correctly.

### Related Issue

Apart of #2523


aklinker1:
Tested in one of my extensions for real, works just fine.

PR #2559, opened 2026-08-01
chore(deps): replace `picomatch` with  `path`

suveshmoza:
### Overview

Replace `picomatch`(93KB) with Node's `path.matchesGlob`.

### Manual Testing

`bun run --filter wxt test -- run src/core/utils/building/__tests__/find-entrypoints.test.ts`

### Related Issue

#2523


packages/wxt/src/core/utils/building/find-entrypoints.ts
```diff
@@ -61,7 +61,7 @@ export async function findEntrypoints(): Promise<Entrypoint[]> {
       const inputPath = resolve(wxt.config.entrypointsDir, relativePath);
       const name = getEntrypointName(wxt.config.entrypointsDir, inputPath);
       const matchingGlob = pathGlobs.find((glob) =>
-        picomatch(glob)(relativePath),
+        path.matchesGlob(relativePath, glob),
```
aklinker1, line 64: Looks like this was introduced in v22.5.0... WXT's engine is set to `>=22`, so this is technically incompatible... 

https://nodejs.org/api/path.html#pathmatchesglobpath-pattern

What about using regex instead? The mapping is very simple, maybe we change the `PATH_GLOB_TO_TYPE_MAP` to an array of tuples?

```ts
const PATH_GLOB_TO_TYPE_MAP: Array<[pattern: string | RegExp, type: Entrypoint['type']]> = {
  ['sandbox.html', 'sandbox'],
  ['sandbox/index.html', 'sandbox'],
  [/.*\.sandbox\.html/, 'sandbox'],
  // ...
}
```

Or do we just yolo it and assume no one is using node 22.0-22.4?

suveshmoza: We can go yolo ;) considering that the LTS version of Node V22 is `v22.23.2`.  Regex approach seems doable

aklinker1:
Hmm, let's hold off on this one for now - I don't want to yolo it when there are 5 other packages using it. So it's still going to be installed until we get rid of the other dependencies. So this one can wait.

Screenshot 2026-08-08 at 11 43 47 AM

Engines:

```
@1natsu/wait-element                       4.1.2        -
@aklinker1/rollup-plugin-visualizer        5.12.0       >=14
@aklinker1/zero-zip                        1.0.1        -
@topcli/prompts                            4.0.0        >=22.0.0
@webext-core/fake-browser                  2.0.1        *
@webext-core/isolated-element              3.0.0        *
@webext-core/match-patterns                2.0.0        *
@wxt-dev/browser                           0.2.5        -
@wxt-dev/storage                           1.2.9        -
c12                                        3.3.4        -
cac                                        7.0.0        >=20.19.0
chokidar                                   5.0.0        >= 20.19.0
consola                                    3.4.2        ^14.18.0 || >=16.10.0
defu                                       6.1.7        -
dotenv-expand                              13.0.0       >=12
filesize                                   11.0.17      >= 10.8.0
get-port-please                            3.2.0        -
giget                                      3.2.0        -
hookable                                   6.1.0        -
json5                                      2.2.3        >=6
linkedom                                   0.18.12      >=16
magicast                                   0.5.2        -
nanospinner                                1.2.2        -
nypm                                       0.6.5        >=18
picomatch                                  4.0.5        >=12
publish-browser-extension                  5.1.0        >=18.0.0
superlock                                  1.3.2        >= 14
tiny-open                                  1.3.0        -
tinyexec                                   1.2.4        >=18
tinyglobby                                 0.2.17       >=12.0.0
unimport                                   5.6.0        >=18.12.0
```

suveshmoza:
Yeah, makes sense

PR #2581, opened 2026-08-06
fix: don't set use_dynamic_url on browsers that don't support it

rxliuli:
## Summary

Content scripts using `cssInjectionMode: 'ui'` unconditionally get `use_dynamic_url: true` on their generated `web_accessible_resources` entry, regardless of target browser.

`use_dynamic_url` is a Chrome-only MV3 field (obscures the resource URL behind a per-session token to prevent fingerprinting) — [MDN's browser-compatibility table](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/web_accessible_resources#browser_compatibility) lists it as unsupported in Firefox and Safari. Firefox silently ignores the unknown key, but Safari's web extension converter warns about it:

```
Warning: The following keys in your manifest.json are not supported by your current version of Safari...
	use_dynamic_url
```

(It's a warning, not a build failure — the converter still produces a working Xcode project. Reproduced by building a project with a `cssInjectionMode: 'ui'` content script and running `wxt zip -b safari`.)

## Changes

- `use_dynamic_url: true` is now only added when `wxt.config.browser` is neither `'firefox'` nor `'safari'`. The `resources`/`matches` entry itself is still added on every browser — only this one field is conditional.
- Pinned `browser: 'chrome'` in the three existing tests that asserted `use_dynamic_url: true` — they previously passed only because the test fixture's default browser is randomly `chrome` or `firefox` (`faker.helpers.arrayElement(['chrome', 'firefox'])`), so after this fix they'd become flaky without an explicit browser.
- Added a test asserting the field is omitted for both `firefox` and `safari`.

## Test plan

- [x] `vitest run src/core/utils/__tests__/manifest.test.ts` — 92 tests passing, run repeatedly across several random `faker` seeds
- [x] `tsc --noEmit` — no new errors

Related: #2580 (same file, unrelated root cause — split into two PRs per CONTRIBUTING.md's per-PR changelog convention).

### DIFF
--- a/packages/wxt/src/core/utils/manifest.ts
+++ b/packages/wxt/src/core/utils/manifest.ts
@@ -590,7 +590,12 @@ export function getContentScriptCssWebAccessibleResources(
 
     resources.push({
       resources: [cssFile],
-      use_dynamic_url: true,
+      // Chrome-only MV3 field (obscures the resource URL behind a per-session
+      // token); Firefox ignores it, and Safari's web extension converter
+      // rejects it as an unsupported key.
+      ...(wxt.config.browser !== 'firefox' && wxt.config.browser !== 'safari'
+        ? { use_dynamic_url: true }
+        : {}),
       matches:
         script.options.matches?.map((matchPattern) =>
           stripPathFromMatchPattern(matchPattern),
--- a/packages/wxt/src/core/utils/__tests__/manifest.test.ts
+++ b/packages/wxt/src/core/utils/__tests__/manifest.test.ts
@@ -1231,6 +1231,7 @@ describe('Manifest Utils', () => {
               outDir,
               command: 'build',
               manifestVersion: 3,
+              browser: 'chrome',
             },
           });
 
@@ -1315,6 +1316,7 @@ describe('Manifest Utils', () => {
               outDir,
               command: 'build',
               manifestVersion: 3,
+              browser: 'chrome',
             },
           });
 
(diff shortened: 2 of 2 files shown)


packages/wxt/src/core/utils/manifest.ts
```diff
@@ -590,7 +590,12 @@ export function getContentScriptCssWebAccessibleResources(
 
     resources.push({
       resources: [cssFile],
-      use_dynamic_url: true,
+      // Chrome-only MV3 field (obscures the resource URL behind a per-session
+      // token); Firefox ignores it, and Safari's web extension converter
+      // rejects it as an unsupported key.
+      ...(wxt.config.browser !== 'firefox' && wxt.config.browser !== 'safari'
```
aklinker1, line 596: This makes assumptions about the strings people will use - what about an 'ios' build?

I think there are patterns to better tell if a browser is chrome, I can't find them on mobile. I'll look this up later, probably Saturday

rxliuli: Re iOS: wxt.config.browser for iOS/macOS builds is 'safari' — they differ by runtime.getPlatformInfo().os, not by browser target (ref: [DefinitelyTyped#74923](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/74923)). So the current check already covers both.

Happy to switch to a whitelist. Since TargetBrowser is string, neither approach is fully future-proof — a whitelist like chrome/edge would miss other Chromium browsers (opera, brave, …), while the current blacklist would miss a hypothetical new non-Chromium target. But missing use_dynamic_url is harmless (just no URL obscuring), while including it on an unsupported browser causes warnings — so whitelist is the safer default. I'll switch to === 'chrome' || === 'edge' if that works for you.

aklinker1: Hmm, I could have sworn I solved this by having a config that specified if a browser target was chromium or not... but I can't find it, I must have implemented it in some other project.

Since this is just a warning, it'll be fine for most cases, no worse than it is today at least. Let's merge as-is

aklinker1: Gonna open an issue for that idea so we can implement something standard.

aklinker1: See 2588

aklinker1 (approved):

PR #2589, opened 2026-08-08
chore(deps): Replace `filesize` with DIY function

aklinker1:
### Overview

58 kB &rarr; 178 B

### Related Issue

Apart of #2523

### DIFF
--- a/packages/wxt/src/core/utils/fs.ts
+++ b/packages/wxt/src/core/utils/fs.ts
@@ -44,3 +44,9 @@ export async function getPublicFiles(): Promise<string[]> {
   });
   return files.map(unnormalizePath);
 }
+
+export function getBytesDisplay(bytes: number): string {
+  if (bytes < 1000) return `${bytes} B`;
+  if (bytes < 999995) return `${(bytes / 1e3).toFixed(2)} kB`;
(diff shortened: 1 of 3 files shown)


packages/wxt/src/core/utils/fs.ts
```diff
@@ -44,3 +44,9 @@ export async function getPublicFiles(): Promise<string[]> {
   });
   return files.map(unnormalizePath);
 }
+
+export function getBytesDisplay(bytes: number): string {
+  if (bytes < 1000) return `${bytes} B`;
+  if (bytes < 999995) return `${(bytes / 1e3).toFixed(2)} kB`;
```
aklinker1, line 50: Using standard scientific notation for file size, not the base 2 one. So 1 kilobytes (kB) = 1000 bytes, not 1 kibibyte (KiB) = 1024 bytes.

https://en.wikipedia.org/wiki/Kilobyte

aklinker1, line 50: Using an odd cutoff here so 999,999 bytes is displayed as `1 MB` instead of `1000 kB`, given how `toFixed` works. See tests for founding and unit behavior.

PR #1412, opened 2025-02-08
fix: auto-import does not work when using `@vitejs/plugin-vue` alone

aiktb:

PR #2606, opened 2026-08-29
fix: Generate high resolution sourcemaps when injecting auto-imports

### LINKED ISSUE #2604: Auto-imports emit a line-only sourcemap, silently collapsing code coverage (regression since v0.19.27, affects build and test)
### Describe the bug

In a project using `WxtVitest()`, `@vitest/coverage-v8` (and `-istanbul`) report near-total coverage collapse for every module that references an auto-imported global. The whole module body maps to a single statement, so it is neither counted as covered nor as uncovered — it vanishes from the denominator.

In my project this made a `coverage.thresholds` gate of 100% pass while the real figure was 90%: coverage measured **871 of 3,390 statements**, hiding ~300 genuinely uncovered lines. It fails open and silently, which is the dangerous part — a threshold gate reports green precisely because the code is invisible.

A 1,139-line component came out as:

```
SF:src/components/ConfigForm.tsx
DA:65,3
DA:89,3
LF:2
LH:2
```

Line 65 is a top-level `const`; line 89 is the component's declaration. Istanbul's statement map holds one statement spanning lines **89 → 1137**, with `column: null` throughout.

## Cause

`WxtVitest` registers the upstream unimport unplugin directly:

```js
// wxt/dist/testing/wxt-vitest-plugin.mjs
plugins.push(UnimportPlugin.vite(wxt.config.imports));
```

and `unimport/unplugin` generates its sourcemap with no options:

```js
// unimport@6.4.0/dist/unplugin.mjs
return {
  code: s.toString(),
  map: s.generateMap(),   // MagicString defaults to hires: false → no column mappings
};
```

The plugin is `enforce: 'post'`, so it runs after the TS/JSX transform and its line-only map is the last in the chain, discarding the column fidelity everything upstream produced. Modules where `s.hasChanged()` is false are untouched and measure correctly — which is exactly the observed split between files that use `browser`/`storage` and files that don't.

It is also not only the missing columns: `source` is never passed either, so the emitted map carries `sources: [""]`. Reproduced against `unimport` alone, with no bundler involved:

```
current (s.generateMap()):
  sources          : [""]
  segments per line: [0,1,1,1,1,1,1,0]

with hires: 'boundary' + source:
  sources          : ["/src/a.ts"]
  segments per line: [0,10,12,18,18,8,1,0]
```

## This is a regression

[#381](https://github.com/wxt-dev/wxt/pull/381) did add the fix — but to WXT's own `wxt:unimport` Vite plugin:

```js
map: injected.s.generateMap({ hires: 'boundary', source: id }),
```

[#1412](https://github.com/wxt-dev/wxt/pull/1412) then deleted that plugin and pointed **both** the build path and `WxtVitest()` at `UnimportPlugin.vite()`:

```diff
     addViteConfig(wxt, () => ({
-      plugins: [vitePlugin(unimport)],
+      plugins: [UnimportPlugin.vite(options)],
     }));
```

```diff
-    plugins.push(unimportPlugin(unimport));
+    plugins.push(UnimportPlugin.vite(wxt.config.imports));
```

The `hires`/`source` options went with it, and nothing has carried them since — `git log -S hires --all` returns only #381, #771 and #1412. First affected release is **v0.19.27** (2025-02-12).

So this is not a testing-path gap; both paths lost the same fix. Coverage is simply where it is loudest, since the same transform feeds Vite's sourcemap chain during builds.

## Upstream

I've filed the root cause at [unjs/unimport#562](https://github.com/unjs/unimport/issues/562) — the bare `s.generateMap()` at `src/unplugin.ts:64` is still there on `main`, so every unplugin consumer is affected, not just WXT. If unimport takes the default change, whatever WXT does here can be removed again later.

## Suggested fix

Restore the `wxt:unimport` plugin with #381's sourcemap options, keeping the two things #1412 actually needed: `enforce: 'post'` and a regex `include`/`exclude` filter (the `@vitejs/plugin-vue` case), with the defaults imported from `unimport/unplugin` rather than copied so they can't drift. One plugin, shared by the build path and `WxtVitest()`, so both are fixed by the same change.

### Reproduction

Any WXT project with `WxtVitest()` in `vitest.config.ts` and `coverage.include: ['src/**']`.
Such as https://github.com/zizzfizzix/scrape-similar

### Steps to reproduce

1. Have a module that references an auto-imported global (`browser`, `storage`, …) and a test that renders/exercises it.
2. `vitest --run --coverage` → that module reports ~1-3 measured statements at 100%.
3. Edit `node_modules/.pnpm/unimport@*/node_modules/unimport/dist/unplugin.mjs`, changing `map: s.generateMap()` to `map: s.generateMap({ hires: 'boundary', source: id })`.
4. Re-run → measured statements jump and real coverage appears.

Measured in my project:

| `map:` value | measured statements | reported coverage |
| --- | --- | --- |
| `s.generateMap()` (current) | 871 | 100% |
| `s.generateMap({ hires: true })` | 3,390 | 90.02% |
| `s.generateMap({ hires: 'boundary', source: id })` | 3,390 | 90.02% |

`hires: 'boundary'` matches `hires: true` exactly here and is the cheaper option — and it is already what [#381](https://github.com/wxt-dev/wxt/pull/381) chose.

Reduced to a four-branch module using one auto-imported global, with a test covering a single branch:

| | Statements | Branches |
| --- | --- | --- |
| today | **100% (0/0)** — file absent from the report | 100% (0/0) |
| with the fix | 37.5% (3/8) | 16.66% (1/6) |

### System Info

- wxt 0.21.4 (broken since 0.19.27)
- unimport 6.4.0 (transitive)
- vitest 4.1.11, @vitest/coverage-v8 4.1.11 (reproduced identically with `@vitest/coverage-istanbul`, so it is not provider-specific)
- vite 8.2.2

### Used Package Manager

pnpm

### Validations

zizzfizzix:
### Overview

`unimport/unplugin` calls `MagicString#generateMap()` with no options, which defaults to `hires: false` — a line-level sourcemap. Because auto-imports are injected at the top of the file, every mapping points back at the first line of the original module, so tools that rely on columns (like V8 code coverage) see the whole module as a single statement.

The result: any module referencing an auto-imported variable (`browser`, `storage`, …) silently drops out of coverage reports, inflating the reported percentage and letting thresholds pass when they shouldn't.

This was originally fixed for the build path in #381, but #1412 replaced WXT's own `wxt:unimport` plugin with `UnimportPlugin.vite()` in **both** the build path and `WxtVitest()`, dropping the `hires` option along the way. So today both paths are affected, not just testing.

This PR restores the `wxt:unimport` plugin with the sourcemap options from #381 — `generateMap({ hires: 'boundary', source: id })`. `hires: 'boundary'` is enough to keep columns accurate and is cheaper than a full `hires: true` map.

To keep the maintenance surface minimal, the plugin is the pre-#1412 one plus the two things #1412 actually needed: `enforce: 'post'` and a regex `include`/`exclude` filter. The filter defaults are imported from `unimport/unplugin` (`defaultIncludes`/`defaultExcludes`) rather than copied, so they can't drift out of sync. Everything else — `dts` generation, the `autoImport` toggle — was unused by WXT and isn't reimplemented.

The plugin is shared by `packages/wxt/src/builtin-modules/unimport.ts` and `packages/wxt/src/testing/wxt-vitest-plugin.ts`, so build and test are fixed with one change.

The root cause is filed upstream as [unjs/unimport#562](https://github.com/unjs/unimport/issues/562) — the bare `s.generateMap()` is still on `main`, so every unplugin consumer is affected. If unimport changes its default, this plugin can go back to being `UnimportPlugin.vite(options)`; the JSDoc says so and links the ticket.

### Manual Testing

Added `packages/wxt/src/builtin-modules/__tests__/unimport.test.ts`, which asserts the generated sourcemap has more than one segment per line and that `sources` contains the module ID. Both assertions fail when `hires: 'boundary'`/`source` are removed.

End-to-end check with a probe module in `packages/wxt-demo` that uses `browser.runtime.id` across four branches, with a test covering only one of them:

```ts
// src/utils/coverage-probe.ts
export function probe(input: number) {
  const id = browser.runtime.id;
  if (input > 10) return `big:${id}`;
  if (input > 5) return `medium:${id}`;
  if (input > 0) return `small:${id}`;
  return `zero:${id}`;
}
```

```sh
bun run test run --coverage --coverage.include='src/utils/coverage-probe.ts' \
  --coverage.reporter=text src/utils/__tests__/coverage-probe.test.ts
```

Before — the module is invisible to the coverage reporter and reports a perfect score:

```
File               | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s
-------------------|---------|----------|---------|---------|------------------
-------------------|---------|----------|---------|---------|------------------
Statements   : 100% ( 0/0 )
Branches     : 100% ( 0/0 )
```

After — real numbers, with the uncovered branches correctly attributed:

```
File               | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s
-------------------|---------|----------|---------|---------|------------------
 coverage-probe.ts |    37.5 |    16.66 |     100 |    37.5 | 6-12
-------------------|---------|----------|---------|---------|------------------
Statements   : 37.5% ( 3/8 )
Branches     : 16.66% ( 1/6 )
```

(The probe files were only used for verification and are not part of this PR.)

Also ran the full `wxt` package suite (`bun run test run` → 53 files / 554 tests passing) and `bun run check` (ESLint, Oxlint, Publint, TypeScript) in `packages/wxt`.

### Related Issue

This PR closes #2604

### DIFF
--- a/packages/wxt/src/builtin-modules/__tests__/unimport.test.ts
+++ b/packages/wxt/src/builtin-modules/__tests__/unimport.test.ts
@@ -1,0 +1,97 @@
+import { describe, expect, it } from 'vitest';
+import { unimportPlugin } from '../unimport';
+import type { WxtResolvedUnimportOptions } from '../../types';
+
+const options = (): WxtResolvedUnimportOptions => ({
+  disabled: false,
+  eslintrc: { enabled: false, filePath: '', globalsPropValue: true },
+  presets: [{ from: 'wxt/browser', imports: ['browser'] }],
+});
+
+const transform = (
+  plugin: ReturnType<typeof unimportPlugin>,
+  code: string,
+  id: string,
+) => {
+  const transform = plugin.transform;
+  const handler =
+    typeof transform === 'function' ? transform : transform!.handler;
+  return handler.call({} as any, code, id) as Promise<
+    { code: string; map: { mappings: string; sources: string[] } } | undefined
+  >;
+};
+
+/**
+ * Sourcemap `mappings` are grouped by generated line (`;`), then by segment
+ * (`,`). A line-level sourcemap only has one segment per line, so counting
+ * segments tells us whether column information was included.
+ */
+const segmentsPerLine = (mappings: string) =>
+  mappings.split(';').map((line) => (line === '' ? 0 : line.split(',').length));
+
+describe('Unimport Module', () => {
+  describe('unimportPlugin', () => {
+    const id = '/src/utils/example.ts';
+    const code = [
+      'export function getExtensionId() {',
+      '  const id = browser.runtime.id;',
+      '  return id.toUpperCase();',
+      '}',
+      '',
+    ].join('\n');
+
+    it('should inject auto-imports', async () => {
+      const res = await transform(unimportPlugin(options()), code, id);
+
+      expect(res?.code).toContain("import { browser } from 'wxt/browser';");
+      expect(res?.code).toContain('browser.runtime.id');
+    });
+
+    it('should generate a sourcemap with column-level mappings', async () => {
+      const res = await transform(unimportPlugin(options()), code, id);
+
+      // Without `hires`, every line collapses to a single segment, which makes
+      // coverage tools treat the entire module as one statement.
+      // See https://github.com/wxt-dev/wxt/issues/2604
+      expect(Math.max(...segmentsPerLine(res!.map.mappings))).toBeGreaterThan(
+        1,
+      );
+    });
+
+    it('should include the module ID as the sourcemap source', async () => {
+      const res = await transform(unimportPlugin(options()), code, id);
+
+      expect(res!.map.sources).toEqual([id]);
+    });
+
+    it('should not transform files that do not use auto-imports', async () => {
+      const res = await transform(
+        unimportPlugin(options()),
+        'export const one = 1;\n',
+        id,
+      );
+
+      expect(res).toBeUndefined();
+    });
+
+    it('should not transform excluded files', async () => {
+      const res = await transform(
+        unimportPlugin(options()),
+        code,
+        '/node_modules/example/index.js',
+      );
+
+      expect(res).toBeUndefined();
+    });
+
+    it('should not transform non-JS files', async () => {
+      const res = await transform(
+        unimportPlugin(options()),
+        code,
+        '/src/a.css',
+      );
+
+      expect(res).toBeUndefined();
+    });
+  });
+});
--- a/packages/wxt/src/builtin-modules/unimport.ts
+++ b/packages/wxt/src/builtin-modules/unimport.ts
@@ -8,7 +8,8 @@ import type {
   EslintConfigVersion,
 } from '../types';
 import { type Unimport, createUnimport, toExports } from 'unimport';
-import UnimportPlugin from 'unimport/unplugin';
+import { defaultExcludes, defaultIncludes } from 'unimport/unplugin';
+import { createFilter, type FilterPattern, type Plugin } from 'vite';
 
 export default defineWxtModule({
   name: 'wxt:built-in:unimport',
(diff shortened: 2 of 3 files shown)


zizzfizzix:
This should be fixed in unplugin since 6.5.0 by https://github.com/unjs/unimport/pull/563

PR #2607, opened 2026-08-30
fix: use one-arg import.meta.resolve for Bun compatibility

bompus:
## Summary
- Bun's two-argument `import.meta.resolve(specifier, parent)` fails for bare package names (`Cannot find package 'wxt'` when resolving from inside the package, and the same for WXT modules with a synthetic `…/index.js` parent).
- Switch `resolveWxtModuleDir` and `resolveWxtUserModules` to one-arg `import.meta.resolve`, which works on both Bun and Node.
- Lets `bunx --bun wxt` / `bun --bun` run the WXT CLI (verified: HMR + production build).

## Related
- Bun: filing companion issue for two-arg `import.meta.resolve` package-name failures.

## Test plan
- [ ] `bunx --bun wxt --help` in a project that depends on `wxt`
- [ ] `bunx --bun wxt` (dev) starts without `Cannot find package 'wxt'`
- [ ] Project using `@wxt-dev/unocss` (or other `modules:`) still loads modules under Bun and Node
- [ ] Existing Node-based `npx wxt` / `node …/wxt.mjs` still works

### DIFF
--- a/packages/wxt/src/core/resolve-config.ts
+++ b/packages/wxt/src/core/resolve-config.ts
@@ -29,7 +29,7 @@ import { getEslintVersion } from './utils/eslint';
 import { safeStringToNumber } from './utils/number';
 import { loadEnv } from './utils/env';
 import { getPort } from 'get-port-please';
-import { fileURLToPath, pathToFileURL } from 'node:url';
+import { fileURLToPath } from 'node:url';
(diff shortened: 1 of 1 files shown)


bompus:
Companion Bun issue for the two-arg `import.meta.resolve` failure: https://github.com/oven-sh/bun/issues/40931

Issue #2301, opened 2026-04-24
Make impossible to use not supported options from manifest for specific browser.


PatrykKuniczak:
I mean instead of adding comments like https://github.com/wxt-dev/wxt/blob/09b5a1957708ad4fdf5d15bbc65ea2097949f568/packages/wxt/src/types.ts#L764-L769

We should create discriminator, which trims not available options for specific target and `inteliSense` won't even suggest it to user.

That's big issue, but it's valuable to do it someday.

PatrykKuniczak:
@aklinker1 Are you think it's good idea to add v1 milestone here?

aklinker1:
I don't think this is WXT's responsibility. It's way too many edge cases, I'm not willing to maintain that.

If you want to create a POC for a sort of manfiest linter, feel free to make something and propose it to me. But I don't want the core WXT package to be responsible for this.

PatrykKuniczak:
@aklinker1 Let's leave this one open for now.
If we'll have more maintainers someday(I hope), maybe we'll able able to handle that.

This really can be a game changer for newbies and not only, but that's massing update.

PR #2622, opened 2026-09-14
feat: resolve the actual browser for browser-specific code generation

### LINKED ISSUE #2588: Resolve the actual browser for any "--browser" strings
### Feature Request

Internally, there's lots of places where WXT makes assumptions about the strings people are using for the `--browser` flag. Common examples would be "chrome", "firefox", "safari", "opera", "brave", etc, but it can be anything.

The assumptions WXT makes are related to how it resolves specific things, like manifest options, that may only be supported by chrome or firefox or safari. But what if someone has both an "ios" target and a "safari" target? Any logic that checks `wxt.config.browser === "safari"` will miss ios, when ideally "ios" should result in the same manifest as "safari".

So WXT needs a way to resolve which browser to generate things for - but that is different from which label you're building under.

Proposing a few changes to the current `ResolvedConfig` and `InlineConfig` types:

1. Keep using `InlineConfig#browser` (`--browser` CLI flag) for the output directory, `import.meta.env.BROWSER` and related runtime constants
2. Introduce `InlineConfig#targetBrowserMap: Record<TargetBrowser, "` that has a good default, but that can be customized by projects to map different `--browser` strings to the actual browser the extension will run in
3. Introduce `ResolvedConfig.actualBrowser: "chromium" | "firefox" | "safari"` which is used when generating manifest or other browser-specific code. I'm not sure if opera (or future browsers/forks) but it could be added to the union in the future, and we'd have have to update the checks.

I thought this was already implemented, but I must have done it somewhere else, never ported into WXT.

Discussion here: https://github.com/wxt-dev/wxt/pull/2581#discussion_r3727922886

AFAIK there is no way to "know" if a `--browser` is chrome or not without a way to map custom values to their actual browser.

### Additional context

Related to #2301

devtechedge:
### Overview

Fixes #2588.

Right now every build-time browser check compares `wxt.config.browser` against a literal like `"firefox"` or `"safari"`. An `ios` build therefore misses every Safari check and generates the wrong manifest, even though it runs in Safari.

This PR implements the three changes proposed in the issue:

1. `InlineConfig#browser` keeps working exactly as before: it still drives the output directory template, `import.meta.env.BROWSER`, the `CHROME`/`FIREFOX`/`SAFARI`/`EDGE`/`OPERA` env constants, the zip artifact name, and browser launching. Nothing changes for existing labels.
2. `InlineConfig#targetBrowserMap` maps `--browser` strings to the actual browser. It defaults to:

   | `--browser` value | Actual browser |
   | --- | --- |
   | `chrome`, `edge`, `brave`, `opera`, `chromium` | `chromium` |
   | `firefox` | `firefox` |
   | `safari`, `ios` | `safari` |

   Unknown labels are not mapped to anything by default. They resolve to themselves until the project adds an entry for them, so their behavior is unchanged.
3. `ResolvedConfig#actualBrowser: "chromium" | "firefox" | "safari"` is resolved through that map and is now used for the manifest generation checks: `use_dynamic_url` (introduced in #2581), MV2/MV3 key stripping, background/background scripts, `version_name`, `browser_style`/`chrome_style`, `options_ui` in Safari, side panels, and the Firefox warnings. The default manifest version (MV2 for Firefox and Safari) is also resolved off the actual browser now, so an `ios` build gets the same manifest as a `safari` build.

The per-entrypoint `include`/`exclude` lists and per-browser entrypoint options still key off the raw `--browser` label, since those are user-authored values keyed by label.

### Manual Testing

- Added mapping tests to `resolve-config.test.ts`: every default mapping resolves correctly, a custom `targetBrowserMap` entry wins over the defaults, unknown labels resolve to themselves, `actualBrowser` appears on the resolved config, and the output directory still uses the raw label while the default manifest version keys off the actual browser.
- Added a manifest test proving an `ios` build with `actualBrowser: "safari"` skips the Chrome-only `use_dynamic_url` field.
- Verified fail-before/pass-after: with only the source changes stashed, the new tests fail (11 config resolution assertions plus the manifest assertion); with them restored, all pass.
- `bun run check` passes at the repo root, `bun run test` passes in `packages/wxt` (559 tests, one pre-existing flaky e2e that passes on re-run and is unrelated to this change).

### Related Issue

This PR closes #2588. Related discussion: https://github.com/wxt-dev/wxt/pull/2581#discussion_r3727922886 and #2301.


devtechedge:
Hi - gentle follow-up on the actual-browser resolution work whenever it fits your queue.

Builds keyed as `ios` (and other Chromium/Safari aliases) were missing browser-specific manifest checks because those compared `wxt.config.browser` to literals like `"safari"` (`Fixes #2588`).

This adds `InlineConfig#targetBrowserMap` and `ResolvedConfig#actualBrowser` (`chromium` | `firefox` | `safari`) in `packages/wxt/src/types.ts` / `resolve-config.ts` / `target-browser.ts`, and switches the manifest path in `packages/wxt/src/core/utils/manifest.ts` to `actualBrowser` while output dirs, `import.meta.env` constants, zip names, and launching stay on the raw label.

Tests in `resolve-config.test.ts` and `manifest.test.ts` cover the default map, custom overrides, unknown labels resolving to themselves, and the Safari/iOS manifest behavior (12 new assertions; local `bun run check` plus packages/wxt vitest were green here).

Validate / pkg.pr.new / Check PR Title are still waiting on first-contribution approval; Netlify preview and a few pages checks already completed successfully.

Happy to reshape the default map, tighten types, or add another fixture if that would make review easier.

No rush at all - still ready from my end whenever bandwidth allows.

Issue #1029, opened 2024-10-03
Built-in support for Content Scripts running on SPAs


aklinker1:
### Feature Request

Dealing with loading your content script correctly in SPAs is really annoying. WXT should be able to do it for you.

```ts
export default defineContentScript({
  matches: ['*://*.youtube.com/watch*'],
  spa: true,

  main(ctx) {
    console.log('YouTube content script loaded');
    mountUi(ctx);
  },
});
```

#### Is your feature request related to a bug?

N/A

### What are the alternatives?

As the docs state, manually listening to URL changes works. But managing `ctx` correctly is annoying, and not shown in the docs. This would also create ctx for each URL path, so the script can shut down properly when the path changes.

### Additional context

<img width="690" alt="Screenshot 2024-10-03 at 1 01 49 PM" src="https://github.com/user-attachments/assets/471b1ac1-9dd6-4fe6-a303-d0c32ede88ef">

> &ndash; https://discord.com/channels/1212416027611365476/1224500470278520904/1291359490121138207

aklinker1:
This won't be an easy first-contribution, but it should be relatively easy to implement:

1. Add new `spa` type for content script definition
2. Implement a new "SPA handler" utility that manages context and the URL change listener.
3. Add spa handler util to virtual entrypoints if `spa: true`
4. Remove path from content script match patterns when `spa: true`, somehow passing the real match pattern into the util

aklinker1:
I've implemented this in my own extension. There's a few TODOs, but it seems to be working well. It needs to be adapted into the form [mentioned above](https://github.com/wxt-dev/wxt/issues/1029#issuecomment-2392099502), and we should mark it as experimental.

https://github.com/anime-skip/player/blob/d62a241f92ee7c453d0d33e31939a5b741d0af5b/apps/extension/src/utils/define-spa-content-script.ts

PR #2623, opened 2026-09-15
feat: Add experimental spa option for content scripts

creeperkatze:
### Overview

Adds an experimental `spa: true` option for isolated world content scripts.

WXT strips the path from `matches` in the manifest so the script registers against the origin, then re-checks the real patterns at runtime on every URL change. `main` gets a fresh `ContentScriptContext` per matching page, and the previous one is aborted before the next call and when navigating away. `spa.key` controls when `main` re-runs, defaulting to everything but the hash.

Gated behind `experimental.spaContentScripts`. SPA scripts build through their own virtual entrypoint, so other content scripts don't bundle the handler.

Adapted from `define-spa-content-script.ts`, with two differences: it recreates the child context between matching pages rather than re-running `main` on the live one, and per-browser `matches` resolve via `import.meta.env.BROWSER`.

Also fixes `createLocationWatcher`. The Navigation API branch listened for `navigate`, which fires before the navigation commits, leaving `location.href` on the previous page. Switched to `navigatesuccess`.

### Manual Testing

- `bun run --filter wxt test run`
- Added `spa.content.ts` to `wxt-demo` matching `*://*.youtube.com/watch*`. home > video > other video > home calls `main` once per video with the correct `location.href` and aborts the previous context each time.

### Related Issue

Related to #1029


creeperkatze:
This also includes a small fix to `createLocationWatcher`.

The navigation API listened for navigate, which fires before the navigation commits, so `location.href` still pointed at the previous page inside `wxt:locationchange` listeners. Switched to `navigatesuccess`, matching the polling fallback.

I hit it because spa scripts ran main with a stale `location.href`, but it affects any `wxt:locationchange` listener on Chrome. Confined to location-watcher.ts and its test.

PR #2624, opened 2026-09-18
fix: Analytics plugin

### LINKED ISSUE #1847: Cannot use WXT analytics in contexts without access to the browser.runtime APIs
### Describe the bug

I am trying to integrate with ga via this package https://wxt.dev/analytics.html#wxt-analytics but no matter the zip or the dev, it got an error `Uncaught Error: Cannot use WXT analytics in contexts without access to the browser.runtime APIs`

I believe it is required by this module.
https://github.com/wxt-dev/wxt/blob/main/packages/analytics/modules/analytics/client.ts#L19

why does it have to require an id and if needed, how can I make sure the id are there?

### Reproduction

simple follow the step of installing wxt-dev/analytics follow the guideline in https://wxt.dev/analytics.html#wxt-analytics

### Steps to reproduce

_No response_

### System Info

```shell
Chrome Version 139.0.7258.67 (Official Build) (arm64)
```

### Used Package Manager

yarn

### Validations

creeperkatze:
### Overview

The analytics plugin ran in every entrypoint, including MAIN-world content scripts without `browser.runtime` access. It imported `#analytics` at the top of the file, so `createAnalytics()` threw during module load and crashed the entrypoint before its own code ran. Now the import is dynamic and guarded by a `browser.runtime.id` check, so it's skipped where there's no runtime access instead of crashing.

### Manual Testing
Added a MAIN-world content script next to the analytics module and confirmed it no longer crashes, while analytics still works in the background, popup, and isolated content scripts.

### Related Issue
This PR closes #1847


creeperkatze:
image

creeperkatze:
Once `addWxtPlugin` supports filtering by entrypoint type, the background-plugin could just be registered as background-only instead of guarding itself at runtime. That would fix it properly. This PR is a fix for the immediate crash.

